PR #130, opened 2015-02-02
FastLED3.1

peetervedic:


focalintent:
Why do you keep submitting pull requests of my own changed to my library?

focalintent:
(Ah, never mind - it seems today is the day that multiple people submitted pull requests for my own changes back into my tree)

jar10623:
now i find a solution for me with the timing problem and atmega 1284p in delay.h:

// #define NS(_NS) (_NS / (1000 / (F_CPU / 1000000L)))
#if F_CPU < 96000000
#if defined(__AVR_ATmega1284P__)
	#define NS(_NS) ( (_NS * ( (F_CPU*9L/10L) / 1000000L))) / 1000
	#define CLKS_TO_MICROS(_CLKS) ((long)(_CLKS)) / ((F_CPU*9L/10L) / 1000000L)
#else
	#define NS(_NS) ( (_NS * (F_CPU / 1000000L))) / 1000
	#define CLKS_TO_MICROS(_CLKS) ((long)(_CLKS)) / (F_CPU / 1000000L)
#endif


but now i see FastLED 3.0.3

with:

// #define NS(_NS) (_NS / (1000 / (F_CPU / 1000000L)))
#if 1 || (F_CPU < 96000000)

what is the  const 1 in the or? 
(F_CPU < 96000000) isn't nessesary an the #else never reached

focalintent:
Because I wanted to leave the rest of the code there as a reference because I may come back and re-work it.  The 1|| is an effective shortcut that cuts it out, but leaves it around for me in case I come back to it.

I'm glad you have a solution for the 1284p (because of the changes to the meanings of F_CPU, and because I can't be sure that your changes won't cause problems for the other 3-wire led chipset timings, I am not going to pull it back into the library).

PR #169, opened 2015-05-23
fix typos

flokli:


focalintent:
Most of these were fixed a while ago in other branches.

PR #184, opened 2015-07-05
Added initial support for UC3B in fastpin_avr.h

timseries:


focalintent:
Can you re-do this against the FastLED3.1 branch?  There's been some format/layout changes, otherwise this'll get lost when I pull 3.1 back to master.  (Also the pin cycle macro is much more universal on the 3.1 branch)

timseries:
Sorry, that was an unintentional PR (I had my remotes setup incorrectly). Will make sure I'm using 3.1, thanks!

PR #202, opened 2015-08-23
Add 8-bit palette interpolation function and comparative example.

generalelectrix:
The FastLED compact palettes are really nice creative tools for quickly creating color textures.  However, the effective 4-bit interpolation between palette entries (256-color palette) implies that animating a color fade from one palette entry to another will appear stepped/jagged if the fade time is longer than 16 frames.

I defined an "extended" version of the ColorFromPalette function which uses the 12 most significant bits of a 16-bit index; like the regular 8-bit version, the four most significant bits index the color palette, and the next 8 less significant bits are used as an offset to interpolate between the palette entries (the four LSB are ignored).  Essentially no major changes were required in the underlying function as the scale functions already accept 8-bit values.  This implies full 8-bit color fades between palette entries, so very slow animations using compact palettes are perfectly smooth at no extra expense (with 4-bit interpolation, they are jagged and aesthetically displeasing except when all color fades are faster than about 500 ms - very fast animation!).

I'm not entirely happy with the "ignore the 4 LSB" approach, but it does keep the interface basically identical to the regular 8-bit palette indexing, allows the use of overflow to reset index counters, etc.  Mostly it means that you have to use larger numbers to increment and index into palettes.  I'm happy to hear feedback on this design decision.

This function was the difference in getting a project working perfectly and beautifully with very little effort.  Load up 13 compact palettes, run the animations nice and slow, change them every minute or two, and boom.

Unfortunately the LED rig I wrote this to use on has left for the playa, so I don't have anything to test the example.  In a couple weeks I'll run it on that platform and check to make sure it exhibits the correct behavior.

(btw Hi @focalintent, long time no cross paths.  This is a very nice library you folks have put together!  -Chris Macklin)

### DIFF
--- a/colorutils.cpp
+++ b/colorutils.cpp
@@ -443,6 +443,69 @@ CRGB ColorFromPalette( const CRGBPalette16& pal, uint8_t index, uint8_t brightne
     return CRGB( red1, green1, blue1);  
 }
 
+CRGB ColorFromPaletteExtended( const CRGBPalette16& pal, uint16_t index, uint8_t brightness, TBlendType blendType)
+{
+    // This fucnction has the same intuitive behavior as the other ColorFromPalette
+    // functions, except it provides 8-bit interpolation between palette entries.
+
+    // Extract the four most significant bits of the index as a palette index.
+    uint8_t index_4bit = (index >> 12);
+
+    // Calculate the 8-bit offset from the palette index.
+    // Throws away the 4 least significant bits and uses the middle 8.
+    uint8_t offset = (uint8_t)(index >> 4);
+
+    // Get the palette entry from the 4-bit index
+    //  CRGB rgb1 = pal[ hi4];
+    const CRGB* entry = &(pal[0]) + index_4bit;
+    uint8_t red1   = entry->red;
+    uint8_t green1 = entry->green;
+    uint8_t blue1  = entry->blue;
+
+    uint8_t blend = offset && (blendType != NOBLEND);
+
+    if( blend ) {
+        // If palette blending is enabled, use the offset to interpolate between
+        // the selected palette entry and the next.
+
+        if( index_4bit == 15 ) {
+            entry = &(pal[0]);
+        } else {
+            entry++;
+        }
+
+        // Calculate the scaling factor and scaled values for the lower palette value.
+
+        uint8_t f1 = 256 - offset;
+
+        red1   = scale8_LEAVING_R1_DIRTY( red1,   f1);
+        green1 = scale8_LEAVING_R1_DIRTY( green1, f1);
+        blue1  = scale8_LEAVING_R1_DIRTY( blue1,  f1);
+
+        // Calculate the scaled values for the neighboring palette value.
+        uint8_t red2   = entry->red;
+        uint8_t green2 = entry->green;
+        uint8_t blue2  = entry->blue;
+        red2   = scale8_LEAVING_R1_DIRTY( red2,   offset);
+        green2 = scale8_LEAVING_R1_DIRTY( green2, offset);
+        blue2  = scale8_LEAVING_R1_DIRTY( blue2,  offset);
+
+        cleanup_R1();
+
+        // These sums can't overflow, so no qadd8 needed.
+        red1   += red2;
+        green1 += green2;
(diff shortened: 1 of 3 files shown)


kriegsman:
Thanks for the idea, suggestion, and code!
I'll take a look and see what and how we want to integrate it, but I like the direction here.

msurguy:
@generalelectrix this is exactly what I need to slowly shift colors in a palette, thanks for this PR!

rfeisel:
nice work.

devoh747:
Is this going to be merged?

RomRider:
Hi,
Anyway to have this merged?
It works well (with a slight modification). Please see here #1014  :)

colorutils.cpp
```diff
@@ -443,6 +443,69 @@ CRGB ColorFromPalette( const CRGBPalette16& pal, uint8_t index, uint8_t brightne
...
+
+        if( index_4bit == 15 ) {
+            entry = &(pal[0]);
+        } else {
+            entry++;
+        }
+
+        // Calculate the scaling factor and scaled values for the lower palette value.
+
+        uint8_t f1 = 256 - offset;
```
5chmidti, line 479: I think the problem described in #1014 has to do with this line. All other ColorFromPalette's calc f1 like this:
`uint8_t f1 = 255 - f2;`

5chmidti, line 479:
```suggestion
        uint8_t f1 = 255 - offset;
```

RomRider: Tested and this works. I didn't spot it 😊

sutaburosu:
This works beautifully.  I guess it hasn't been merged because it's incomplete; it doesn't support the 32-entry palette type.  That's a simple enough fix:
```c++
CRGB ColorFromPaletteExtended(const CRGBPalette32& pal, uint16_t index, uint8_t brightness, TBlendType blendType) {
  // Extract the five most significant bits of the index as a palette index.
  uint8_t index_5bit = (index >> 11);
  // Calculate the 8-bit offset from the palette index.
  uint8_t offset = (uint8_t)(index >> 3);
  // Get the palette entry from the 5-bit index
  const CRGB* entry = &(pal[0]) + index_5bit;
  uint8_t red1   = entry->red;
  uint8_t green1 = entry->green;
  uint8_t blue1  = entry->blue;

  uint8_t blend = offset && (blendType != NOBLEND);
  if (blend) {
    if (index_5bit == 31) {
      entry = &(pal[0]);
    } else {
      entry++;
    }

    // Calculate the scaling factor and scaled values for the lower palette value.
    uint8_t f1 = 255 - offset;
    red1   = scale8_LEAVING_R1_DIRTY(red1,   f1);
    green1 = scale8_LEAVING_R1_DIRTY(green1, f1);
    blue1  = scale8_LEAVING_R1_DIRTY(blue1,  f1);

    // Calculate the scaled values for the neighbouring palette value.
    uint8_t red2   = entry->red;
    uint8_t green2 = entry->green;
    uint8_t blue2  = entry->blue;
    red2   = scale8_LEAVING_R1_DIRTY(red2,   offset);
    green2 = scale8_LEAVING_R1_DIRTY(green2, offset);
    blue2  = scale8_LEAVING_R1_DIRTY(blue2,  offset);
    cleanup_R1();

    // These sums can't overflow, so no qadd8 needed.
    red1   += red2;
    green1 += green2;
    blue1  += blue2;
  }
  if (brightness != 255) {
    nscale8x3_video(red1, green1, blue1, brightness);
  }
  return CRGB(red1, green1, blue1);
}
```

netmindz:
Any updates?

Flavourdynamics:
Hi there, this is a great feature that I would love to see included into a future release. I think there many applications where this would really add to a project.

Sarah-C:
Just a reminder for removing janky color fades.... =)

netmindz:
All I want for Christmas is ... Smooth colour fades 😅

zackees:
It's been merged. Will go in on Monday for the next release.

generalelectrix:
I love that a variant of this made it in!

zackees:
Here's the question: is this fast enough for AVR to be the default?

PR #216, opened 2015-10-07
Fix JSON file for @PlatformIO

ivankravets:


focalintent:
Ah, I see the missing commas - out if curiosity, why the cutting down of keywords?

ivankravets:
I don't prefer to keep vendor's product names in keywords. All these data are searchable via `description` field. I like when relation between `keywords` and `libs` is 1:N, but not to 1:1.

What do you mean?

focalintent:
'k - I wasn't sure if keywords were indexed separately (I don't use platform.io, and the nature of my work and projects makes it unlikely that I ever will) - since my library isn't a vendor library, but rather an alternative for many of them, that was why I initially included the most commonly asked for/used chpisets in the keywords.

ivankravets:
Thanks, the library in registry has been updated http://platformio.org/#!/lib/show/126/FastLED

PR #218, opened 2015-10-13
Continuous Integration via @PlatformIO and @travis-ci

ivankravets:
Please enable Travis.CI for this project http://docs.travis-ci.com/user/getting-started/#To-get-started-with-Travis-CI%3A (steps 1, 2, 5)

See forked build: https://travis-ci.org/ivankravets/FastLED

@PlatformIO Documentation: http://docs.platformio.org/en/latest/ci/travis.html

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,37 @@
+language: python
+python:
+    - "2.7"
+
+env:
+    - PLATFORMIO_CI_SRC=examples/AnalogOutput/AnalogOutput.ino
+    - PLATFORMIO_CI_SRC=examples/Blink/Blink.ino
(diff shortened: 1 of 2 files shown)


micahcochran:
This seems like a good idea.  CI can make software a little more reliable because it atleast compiles.  What you could do is [`allow_failures` for tests that current fail](https://docs.travis-ci.com/user/customizing-the-build/#Rows-that-are-Allowed-to-Fail).  Then if the pull request were to be merged, it would allow a future PR to work on why those tests aren't compiling.

zackees:
We are doing CI now with github actions. This is an ancient PR so I doubt it applies anymore

Issue #212, opened 2015-09-24
Example code standardization


cneuzil:
This is probably a minor problem, but I noticed the examples have different names for variables.  Some have DATA_PIN, some have LED_PIN.  Likewise for LED_TYPE and CHIPSET, CLK_PIN and CLOCK_PIN.  NUM_LEDS seems to be consistent.

The first thing I do when I get a new board and LEDs is run blink, then I run other examples like cylon to verify everything works.  If I just copy/paste the variables, I get errors because the variable names don't match.

This is kinda minor, but maybe helpful to get fixed for beginners setting up their projects.

kriegsman:
We just did a bit of code cleanup on the examples; closing this old ticket, and thank you for the report (way back when!)

Issue #52, opened 2014-04-17
MeFirst example


focalintent:
Update/replace FirstLight.ino with a "MeFirst" example that includes enough for someone to get up and running and start playing with the leds.

zackees:
Closing as **superseded** — the onboarding gap this asked for has been filled several times over, and the specific remedy proposed now runs against project policy.

The ask was: *"Update/replace FirstLight.ino with a 'MeFirst' example that includes enough for someone to get up and running and start playing with the leds."*

**What now covers that:**

- **`README.md`** opens with "⚡ Get Blinking in 30 Seconds" — a complete, copy-pasteable 12-line sketch, above the fold, before any other content.
- **`examples/Blink/Blink.ino`** is the canonical minimal starting point (`NUM_LEDS 1`, serial banner, commented chipset lines to uncomment). It's also the default compile target — `bash compile uno` builds Blink.
- **`cookbook/README.md`** — the FastLED Cookbook, linked from README under "📖 Just Getting Started?", covers tutorials and walkthroughs from first blink onward. This is the "enough to start playing with the leds" material, and a prose guide serves that better than a sketch ever did.
- **`examples/FirstLight/FirstLight.ino`** still exists and still does what it always did (animate a white dot along the strip), so nothing was lost.

**Why a new "MeFirst" sketch specifically isn't the answer anymore:** the tree is at 87 examples, and every one is a row in the CI compile matrix. `CLAUDE.md` now states "Keep the `examples/` tree minimal — do NOT create new one-off `.ino` sketches", enforced by a `PreToolUse` hook (`ci/hooks/protect_example_ino.py`) that blocks new `.ino` files under `examples/`. Adding an 88th onboarding example that overlaps Blink and FirstLight would make the getting-started story worse, not better, by giving newcomers a third thing to choose between.

If a concrete gap in the onboarding path turns up, a focused issue against `cookbook/` or the README is the right shape for it — not a new example sketch.

PR #219, opened 2015-10-15
Standardize Examples

jcreedon:
Standardized all of the single stranded examples (i.e. not matrix or multiple) to use the same macros to define strip settings. addLeds() calls were consolidated into the three use cases. Full instructions were added to FirstLight for help on setting up the values for the macros.

This addresses #212 and is probably a step towards #52

### DIFF
--- a/examples/ColorTemperature/ColorTemperature.ino
+++ b/examples/ColorTemperature/ColorTemperature.ino
@@ -1,19 +1,21 @@
 #include <FastLED.h>
 
-#define LED_PIN     3
-
-// Information about the LED strip itself
-#define NUM_LEDS    60
-#define CHIPSET     WS2811
-#define COLOR_ORDER GRB
-CRGB leds[NUM_LEDS];
+#define NUM_LEDS 60
+#define LED_TYPE NEOPIXEL
+#define DATA_PIN 3
+#define CLOCK_PIN 13
+#define COLOR_ORDER RGB
 
 #define BRIGHTNESS  128
+
+// Define the array of leds
+CRGB leds[NUM_LEDS];
+
 
 
 // FastLED v2.1 provides two color-management controls:
 //   (1) color correction settings for each LED strip, and
-//   (2) master control of the overall output 'color temperature' 
+//   (2) master control of the overall output 'color temperature'
 //
 // THIS EXAMPLE demonstrates the second, "color temperature" control.
 // It shows a simple rainbow animation first with one temperature profile,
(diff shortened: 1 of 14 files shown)


examples/ColorTemperature/ColorTemperature.ino
```diff
@@ -70,16 +72,23 @@ void loop()
...
 void setup() {
   delay( 3000 ); // power-up safety delay
+
   // It's important to set the color correction for your LED strip here,
   // so that colors can be more accurately rendered through the 'temperature' profiles
-  FastLED.addLeds<CHIPSET, LED_PIN, COLOR_ORDER>(leds, NUM_LEDS).setCorrection( TypicalSMD5050 );
+  // Uncomment this line for a 3-Wire chipset
+  FastLED.addLeds<LED_TYPE, DATA_PIN, COLOR_ORDER>(leds, NUM_LEDS).setCorrection( TypicalSMD5050 );
+
+  // Uncomment this line for a SPI chipset
```
NicoHood: Im not that english pro, but dont you use "an SPI" here?

focalintent: It can go either way.  If someone "says" SPI by sounding out the letters, it'd be an, eg "an S P I interface", however if someone pronounces SPI as a single word then a is ok "a spy interface" - since I have heard people use SPI in both ways in spoken conversation I'm unlikely to nitpick one way or the other.

NicoHood: Sure. But I like the idea to standardize the examples.

jcreedon:
So after doing some deeper testing I realized this standardized example format will not work for NEOPIXEL as its template does not allow specifying a RGB_ORDER. We could do one of the following:
 - Drop COLOR_ORDER from the standard set and move it to the optional set (kind of like how I did for BRIGHTNESS and FRAMES_PER_SECOND in some of the examples) so it is settable in examples that need them, but it lets them stay at their default in all other examples.
 - Bring the NEOPIXEL template to match all the others in signature by adding RGB_ORDER back in, but override it to GRB anyways a la `template<uint8_t DATA_PIN, EOrder RGB_ORDER> class NEOPIXEL : public WS2812Controller800Khz<DATA_PIN, GRB> {};`. I assume it is set this way because the Adafruit NeoPixels all have that color order. If we update the template in this manner it will allow the interface to be consistent for all chipsets, but still enforce a color order for NEOPIXEL.

Thoughts? Any additional feedback?

focalintent:
I originally had NeoPixel defaulting to GRB and then allowing people to change the order later, and had multiple rounds of people getting confused by color ordering being off, so I hard removed it from NEOPIXEL.

(And you have stumbled into the reason why I, personally, am not a fan of using #define's for choosing led chipsets and the config options, and so I don't in any of my personal project :)

jcreedon:
@focalintent Totally agree about the #defines. The use case I was targeting was beginners who just want to copy and past a common set of setting across different examples. So would you rather remove the LED_TYPE and create a common uncomment-your-chipset set of lines to use in all the examples?

Issue #121, opened 2015-01-15
mega1284 issues


focalintent:
From @jar10623:

i have a Problem with fastLED since i changed from Arduino nanoV3 (all OK) to 
https://github.com/JChristensen/mini1284

i build two to test but this is only a Problem with fastLED!

with https://github.com/cpldcpu/light_ws2812 is no problem, Timing ist well

see the Data Timing with m1284p & fastLED
![m1284p_timing](https://cloud.githubusercontent.com/assets/10545958/5756687/77961f34-9cb2-11e4-8e83-df0d60b8cc82.jpg)

with 33µs (m1284p) für one LED the ws2812b dosn't work, the 30µs is from m328p

focalintent:
I don't have an mega1284p here to test with, but I just did some tweaks to better deal with the different i/o timings that m1284p and put them on the FastLED3.1 branch, so  you should grab that and try it out and let me know if that helps at all.

focalintent:
Re: the comments by @jar10623 on https://github.com/FastLED/FastLED/commit/b4bbd5ca60ffd6a8c35ee191a0d510f41e6c52b9#commitcomment-9311582 - 

There have been a large number of changes, are you using the FastLED3.1 branch or did you just apply the patch for that change?  As I mentioned in the ticket, this is on the FastLED3.1 branch which has a lot of other changes, you should be using that branch for this testing.

In the meantime, without having any m1284p based hardware to test with there isn't a whole lot else I can do at the moment, I'd need a dev environment to get into more accurate timings for individual instructions as well as what the compiler is doing with my assembler.  Also - please keep conversation about this on the page for issue #121 - https://github.com/FastLED/FastLED/issues/121 it's better to have all this tracked in one place rather than in comments spread across commits and issues.  

If you have a pointer for where I can buy a working dev board to work with I can look into getting one to do more direct debugging with this.

focalintent:
> but where is your problem to look at the bitbang code from tim cpldcpu 
> https://github.com/cpldcpu/light_ws2812 it works perfekt with m328p and m1284p

That code is not useful for what FastLED does.  FastLED does a _lot_ more when writing data out than just the bit banging, it is also performing color and brightness correction as well as doing dithering.  As a result, there's a lot more going on between the pin toggles than just nop's, and as such, things are a bit trickier when the CPU is changed from m328p and m1284p.

> a arduino working device based on m1284p?
> there are many in the internet and with use breadboard too

[...]

PR #227, opened 2015-11-19
Fixed pin definitions for the ATMega1284p

aphelps:
This came about from issue https://github.com/FastLED/FastLED/issues/121.  After the hint about hardware SPI I looked into the FastLED pin definitions for this chip and they were all off.  I've updated the definitions based on http://pighixxx.deviantart.com/art/Atmega1284p-352919461 (and a quick look at the spec sheet to verify these).

After these changes I now have FastLED driving my WS2801 strands with and without hardware SPI. I also did a quick test with a strip of WS2812B's and they appeared to work as well.


focalintent:
The problem with pin definitions is they're fairly board specific - the initial 1284p pin definitions that I had put in were based on some board out there that let you drop a 1284p into the slot where a 328p had been in the original uno layout.  What 1284 board are you using/working with - or are you just using a bare 1284?  (Want to see if there's a way to refine the define checking to have both sets available)

aphelps:
These are custom boards I've been developing, using the generic Mighty 1284p bootloader (from https://github.com/JChristensen/mighty-1284p).  I think that has the same pinout as the Mighty-Mini bootloader, as well as the MoteinoMega (http://lowpowerlab.com/blog/2014/08/09/moteinomega-available-now/) which I'd also tested this with when I was debugging why it wasn't working.

You're right that pin outs differ between boards, but this one seems pretty standard as its what is used in the "generic" hobbyist bootloader for this chip (and is a straight-forward mapping between the chip's pins and the software pin values).  Since there isn't an official Arduino using this chip there isn't a right way, but I do think someone randomly developing using a 1284 is more likely to expect this set since its the first match if you google for "1284p bootloader" or "1284p pin map".

focalintent:
Sounds good to me :)  Like i mentioned, my only exposure to the 1284 up until now had been the drop in chip someone else mentioned using.  I'll pull the request in!  Thanks again!

aphelps:
Glad to help.  I've been working on switching a number of projects over to FastLED, its much nicer working with a single library for multiple LED chipsets so I can eliminate all the wrapper functions and #defs I'd been dealing with previously.

PR #230, opened 2015-11-25
Add assignment function setRGB(CRGB&)

drewrisinger:
Enables setting color in array from single CRGB object instead of separate parameters.

### DIFF
--- a/pixeltypes.h
+++ b/pixeltypes.h
@@ -183,6 +183,16 @@ struct CRGB {
         b = nb;
         return *this;
     }
+	
+	// allow assignment from CRGB boject 
+	// (for setting color in CRGB* array from CRGB object, because "=" operator doesn't work)
+	inline CRGB& setRGB (const CRGB& rhs)	__attribute__((always_inline))
+	{
+		r = rhs.r;
+		g = rhs.g;
+		b = rhs.b;
+		return *this;
+	}
 
     // allow assignment from H, S, and V
 	inline CRGB& setHSV (uint8_t hue, uint8_t sat, uint8_t val) __attribute__((always_inline))


focalintent:
Where do you have an example of = not working?  I use = almost exclusively - and would rather actually fix operator= if you're having a problem with it than add an extra function.

kriegsman:
@drewrisinger: I'm also curious about what you mean about "=" not working.  The way it _does_ work (as far as I can tell!) is like this:

  CRGB source = {whatever};
  CRGB dest;
  dest = source; // just assign one CRGB to another.

Not working for you?

drewrisinger:
I tried recreating the problem, but can't seem to replicate it (I switched to setRGB(c.r, c.g, c.b) for the meantime). I was having compiler trouble that day with assignment to an array (Compiling under Arduino IDE for Teensy LC, so a 32-bit MCU). Some error similar to "Can't assign CRGB to uint32_t" with code similar to following:
```c++
void CRGBFunc(CRGB * leds, uint16_t num, CRGB c) {
  for(uint16_t i=0; i<num; i++) {
    leds[i] = c; //my pull request tried changing this to leds[i].setRGB(c);
  }
}
```
My best guess at the time was that the compiler was optimizing the CRGB array to an array of uint32_t, which then had a type mismatch with the CRGB color that I was assigning, so it threw a compiler error.

I just tried reverting to the old code ( leds[i] = c; ), and it seems to be working, so I guess I would ignore this pull request. Sorry about the trouble.

Drew

kriegsman:
Thank you for digging into it and letting us know what you found. You never know when the compiler is going to do something unusual in some special situation, so this kind of ticket is always appreciated!

Thanks again.

PR #257, opened 2016-02-05
Added WS2852 support

DavidMenting:
World Semi has released a 3.5 x 3.5mm package version of the WS2812B. It integrates the controller and RGB leds into a PLCC4 SMD package. 

Like the Neopixel, ordering is defaulted to GRB.


focalintent:
Having the chipset default to an rgb ordering w/neopixel has caused nearly as many problems as it solved, so I'm not doing that anymore with chipsets going forward.  (Especially since not all neopixels are GRB anymore!).

PR #265, opened 2016-02-25
Added Simblee to led_sysdefs.h (same as NRF51 and RFduino).

jasoncoon:
[Simblee](https://www.simblee.com) seems identical to NRF51 and RFduino.

I've verified that this compiles, and a friend with the hardware has verified that it actually works, while I wait on my hardware to arrive.  Please let me know if you'd prefer this handled differently (or not at all).  :)

Thanks!

### DIFF
--- a/led_sysdefs.h
+++ b/led_sysdefs.h
@@ -5,7 +5,7 @@
 
 #include "fastled_config.h"
 
-#if defined(NRF51) || defined(__RFduino__)
+#if defined(NRF51) || defined(__RFduino__) || defined (__Simblee__)
 #include "platforms/arm/nrf51/led_sysdefs_arm_nrf51.h"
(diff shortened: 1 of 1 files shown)


focalintent:
Seems like they just OEM'd the rfduino hardware :)

Out of curiosity, do you know why their build system isn't defining the NRF51 macro?

jasoncoon:
Actually, looks like Simblee is made by RFduino.  The footer on their [site](https://www.simblee.com) says: "Copyright 2015 RF Digital Corp.".  [RF Digital](http://www.rfdigital.com) has the same address as [RFduino](http://www.rfduino.com/).  And I just found [this RFduino forum post](http://forum.rfduino.com/index.php?topic=1066.0) announcing Simblee.  No idea why they didn't just define the NRF51 macro, but I can try to ask them.

PR #276, opened 2016-03-25
Added Arduino MKR1000 pin def

Daniel-dk:
Added Arduino MKR1000 (Arduino Zero -like  SAMD21G18 based controller )
pin definitions, SPI is on a different set of pins


focalintent:
Out of curiosity, why did you close this?  (I haven't had a chance to review it, just catching up on github emails)

PR #295, opened 2016-04-29
Fix qadd7()

NicoHood:
**No AVR implementation was added, defaulting to the c code!**

Test Sketch:
```cpp
#include "FastLED.h"
void setup() {}

void loop() {
  // put your main code here, to run repeatedly:
  Serial.println(qadd7(-100, -100));
  Serial.println(qadd7(100, 100));
  Serial.println(qadd7(-100, -10));
  Serial.println(qadd7(100, -10));
  Serial.println("*****");
  delay(1000);
}
```

Feel free to close the PR and copy the code to get a better commit if you wish.

### DIFF
--- a/lib8tion/math8.h
+++ b/lib8tion/math8.h
@@ -47,15 +47,16 @@ LIB8STATIC_ALWAYS_INLINE uint8_t qadd8( uint8_t i, uint8_t j)
 #endif
 }
 
-/// Add one byte to another, saturating at 0x7F
+/// Add one byte to another, saturating at 0x7F / 0x80
 /// @param i - first byte to add
 /// @param j - second byte to add
-/// @returns the sum of i & j, capped at 0xFF
+/// @returns the sum of i & j, capped at 0x7F / 0x80
 LIB8STATIC_ALWAYS_INLINE int8_t qadd7( int8_t i, int8_t j)
 {
-#if QADD7_C == 1
+#if QADD7_C == 1 || 1
     int16_t t = i + j;
-    if( t > 127) t = 127;
+    if(t > 127) t = 127;
+    if(t < -128) t = -128;
     return t;
 #elif QADD7_AVRASM == 1
     asm volatile(


NicoHood:
Something like this would also make a lot of sense to me:
```cpp
    uint8_t qmod8(uint8_t i, int8_t j)
    {
      int16_t t = i + j;
      if (t > 255) t = 255;
      if (t < 0) t = 0;
      return t;
    }
```
It is a mix of qadd8 and qsub8. I basically dont determine the plus or minus operation by the function name, but rather via a signed int8.

Maybe qadd7 can be renamed (or aliased) to qmod7 and qsub7, as it will all do the same. (qsub7 needs an inverted j though).

focalintent:
mod is the wrong name to use here - the modulus operation (usually abbreviated to mod) is the remainder of a division.  https://en.wikipedia.org/wiki/Modulo_operation.

And your qmod8 function is defeating the whole purpose of the q* functions.  qadd8 takes 2 unsigned  8-bit values and adds them together, clamping at 255.  qsub8 takes 2 unsigned 8-bit values and subtracts them from each other, clamping at 0.  By using an int8 for your qmod8 you cap the maximum amount of difference to 127 in either direction.  So, for example, your qmod8 can't be used to do qadd8(5, 192) - because 192 wraps around to a negative value in int8.  (Forget the fact that with math functions, most people new to programming _do_ distinguish between addition and subtraction by the function names, not by the sign on the variable/number involved).

NicoHood:
"mod" in the sense of "modify". I dont know any other name, suggest something.

The idea is to not make the inputs swapable. It will add/substract from the first value (i) depending on what the addition of value two (j) does. The result will be 0-255, but the 2nd parameter can be signed.

qadd8: uint8_t, uint8_t
qadd7: int8_t, int8_t
qxxx8: uint8_t, int8_t

So maybe just a more clever name has to be found.

tullo-x86:
The underlying operation is signed addition, so maybe `qadd87(uint8_t, int8_t)` makes sense?

I'm sure you've found a use for the function but whatever it is escapes me.

Issue #294, opened 2016-04-26
RFduino/NRF51/Simblee + APA102 timing issue?


jasoncoon:
There seems to be an issue on the Simblee (and possibly RFduino and other NRF51 boards) after commit https://github.com/FastLED/FastLED/commit/720f339e3ca5d60ccc1217c0aa8c70e4f3314852.  That commit works fine (with my changes to led_sysdefs.h from commit https://github.com/FastLED/FastLED/commit/023baa3b063d85fabe7ee05c97dfa67d997ffaa1), but the [next commit](https://github.com/FastLED/FastLED/commit/c635ab85fa013aba55cf9ea10313dc780c2bf870) doesn't.

Something like a simple color fade is glitchy and erratic.  Here's some video of the problem, if it helps: https://goo.gl/photos/2SB4m18DuKrECMAv6

* Arduino 1.6.5 (I've also tried 1.6.8)
* Windows 10
* APA102 (I've tried several different small strips and a 16x16 matrix)
* Followed install instructions in the [Simblee Quick Start](https://www.simblee.com/Simblee_Quickstart_Guide_v1.0.pdf)
* Works fine on Teensy, using the same sketch and LEDs
* Tried with and without a level shifter (SN74AHCT125N)
* Simple test code: https://gist.github.com/jasoncoon/975e225210eda41a66ea5462681a27b1

A guy I've been working with on a Simblee project had his working fine, but I could not get mine working.  After a few weeks of troubleshooting back and forth, we finally compared FastLED versions.  I diffed his copy to master, and tracked his version back to October 2015.

I'll keep digging and see if I can figure out how to fix it.  Any help you can offer would be greatly appreciated.  Thanks!

soamaven:
I see similar issues with current master branch. Reverting to [https://github.com/FastLED/FastLED/commit/720f339e3ca5d60ccc1217c0aa8c70e4f3314852](url) fixes these timing issues

jasoncoon:
Thanks for confirming!  Yeah, I'm afraid I'm still not skilled enough either, but I'm going to try to help at least.  :)

focalintent:
Thanks for the digging/tracking this down (both here and in the g+ post) - I just need to get some time to get one of my rfduino's set up and dig in to what's going on here (especially since it appears to only be the nrf* that's impacted by this)

soamaven:
Hey Guys,

[...]

PR #301, opened 2016-05-04
Properly clear EVENTS_READY register on nrf51 to use hardware SPI

soamaven:
The writeByte() method was not clearing the EVENTS_READY register after writing to the TX register, and so every byte was off by some delay at the beginning of each pixel (it would clear the EVENTS_READY register after each pixel, per the APA102 controller calling waitFully(), which clears the reg.). This led to issue [294](https://github.com/FastLED/FastLED/issues/294). Here is a blurb from the [NRF52 Product Specification] (http://infocenter.nordicsemi.com/index.jsp?topic=%2Fcom.nordic.infocenter.nrf52832.ps.v1.0%2Fspi.html&anchor=topic):

> The SPI master will move the incoming byte to the RXD register after a short delay following the SCK clock period of the last bit in the byte. This also means that the READY event will be delayed accordingly, see  [Figure 3.](http://infocenter.nordicsemi.com/topic/com.nordic.infocenter.nrf52832.ps.v1.0/spi.html#concept_ov2_f54_sr__fig.ready_event_timing_details) Therefore, it is important that you always clear the READY event, even if the RXD register and the data that is being received is not used.


focalintent:
Punting on the pull request - fixing this in a slightly different way - the pull wasn't properly using shouldWait/wait.  Slow clock speeds would mean it never waited, and then bytes would end up getting written on top of each other.

Issue #302, opened 2016-05-05
Assembler compile error


NicoHood:
I got a weird assembler error, with a special function. I reduced the sketch to this (only to get the error, the function itself does not make sense like this):

```cpp
#include "FastLED.h"
#define NUM_LEDS    25
CRGB leds[NUM_LEDS]; // Define the array of leds

void setup() {}

void loop() {
  heartbeat(leds, sizeof(leds) / 3, CRGB::Green);
}

void heartbeat(CRGB l[], uint16_t length, CRGB color) {
  for (uint16_t i = 0; i < 256; i++) {
    for (uint16_t led = 0; led < length; led++) {
      l[led] = color;
      l[led].nscale8_video(cos8(i + 128));
    }
  }
}
```

I guess some inline assembler misses that a special instruction requires an upper register (as the error already sais). This happens with 328 and 32u4 (didnt test the rest). Using latest IDE from git.

```
In file included from /tmp/arduino_modified_sketch_438581/sketch_may05a.ino:2:0:
/home/linuxuser/Documents/Arduino/sketchbook/libraries/FastLED/FastLED.h:12:2: warning: #warning FastLED version 3.001.001 (Not really a warning, just telling you here.) [-Wcpp]
 #warning FastLED version 3.001.001  (Not really a warning, just telling you here.)
  ^
/tmp/ccyvcv0T.s: Assembler messages:
/tmp/ccyvcv0T.s:195: Error: register number above 15 required
/tmp/ccyvcv0T.s:197: Error: register number above 15 required
Using library FastLED at version 3.1.1 in folder: /home/linuxuser/Documents/Arduino/sketchbook/libraries/FastLED 
exit status 1
Error compiling for board Arduino/Genuino Uno.
```

PR #303, opened 2016-05-05
Make gcc register allocation still more flexible

NicoHood:
Regarding  a5e92af and #302

Looking at the datasheet says:
mul: 0 <= d <= 31, 0 <= r <= 31
mov: 0 <= d <= 31, 0 <= r <= 31
eor: 0 <= d <= 31
swap: 0 <= d <= 31
andi: 16 <= d <= 31, 0 <= K <= 255
or: 0 <= d <= 31, 0 <= r <= 31

This means only mx and xr1 should be a "d", the other can be "r". This makes compilation still flexible but should fix the bug too.
I tested this and my assumption is correct.


focalintent:
Not taking this one - it's mostly moot because I'm in the process of completely re-writing this code to accommodate future library changes.

PR #316, opened 2016-06-10
Make lerp8 and lerp16 reference parameters const

mjbcopland:
Allows the use of `const CRGB` objects and rvalue expressions as the first argument to *CRGB::lerp8* and *CRGB::lerp16*.
Consider the following example:
```
CRGB pixel;
CRGB red = CRGB::Red;
const CRGB red_const = red;

pixel.lerp8(red,       0); // OK
pixel.lerp8(-red,      0); // error: no known conversion for argument 1 from 'CRGB' to 'CRGB&'
pixel.lerp8(red_const, 0); // error: no known conversion for argument 1 from 'const CRGB' to 'CRGB&'
pixel.lerp8(CRGB::Red, 0); // error: no known conversion for argument 1 from 'CRGB::HTMLColorCode' to 'CRGB&'
```
The proposed change allows the latter three lines to compile and function as expected.

### DIFF
--- a/pixeltypes.h
+++ b/pixeltypes.h
@@ -485,7 +485,7 @@ struct CRGB {
     }
 
     /// return a new CRGB object after performing a linear interpolation between this object and the passed in object
-    inline CRGB lerp8( const CRGB& other, fract8 frac)
+    inline CRGB lerp8( const CRGB& other, fract8 frac) const
     {
       CRGB ret;
 
@@ -497,7 +497,7 @@ struct CRGB {
     }
 
     /// return a new CRGB object after performing a linear interpolation between this object and the passed in object
-    inline CRGB lerp16( const CRGB& other, fract16 frac)
+    inline CRGB lerp16( const CRGB& other, fract16 frac) const
     {
       CRGB ret;
 
--- a/pixeltypes.h
+++ b/pixeltypes.h
@@ -485,7 +485,7 @@ struct CRGB {
     }
 
     /// return a new CRGB object after performing a linear interpolation between this object and the passed in object
-    inline CRGB lerp8( CRGB & other, fract8 frac)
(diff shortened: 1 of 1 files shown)


kavika13:
I didn't see your pull request, but filed this an issue independently. I wanted to define my colors as constants. I imagine this isn't an uncommon error.

https://github.com/FastLED/FastLED/issues/332

focalintent:
Honestly?  You two are the only two who have brought this up (I didn't merge this when it initially came in because as part of the RGBW/RGB16 work, I've completely re-written CHSV and CRGB (and am currently working on my way through everything else in the library that depends on them, and what time i've had available for working on this over the past few months has been dedicated to that) - and this has already all been taken care of there.

But i'll pull this back into the master branch for folks in the meantime.

mjbcopland:
I've updated the request to add the extra `const` qualifiers.

For what it's worth, since opening this pull request, I've discovered that [`blend(foo, bar, val)`](https://github.com/FastLED/FastLED/blob/master/colorutils.cpp#L262) achieves the same as `foo.lerp8(bar, val)` but already has the necessary `const` qualifiers. So perhaps this issue is moot and it would be better for these functions to be removed rather than fixed and end up duplicating another function, but that's a decision to be made when rewriting the library.

Having said that, I can understand wanting to fix these functions, at least until the library is rewritten, to help those still using `lerp8`/`lerp16`.

kavika13:
Yes, I agree it was a nit-pick. I had already filed what I assumed to be an extremely low priority issue, and assumed it would be de-prioritized for a while. Then found someone happened to make a pull request that was related, so I thought I'd link them.

Thanks for your great lib, and thanks for doing your best to make the community happy, however irrational we're being ;)

PR #328, opened 2016-07-22
Ws2813

alex-Arc:
support for the WS2813 chip both with the normal template, on a port and the newest version of OctoWS2811 has it.[https://github.com/PaulStoffregen/OctoWS2811](url)


focalintent:
Thanks for the pull!

I'm going to roll back a couple of the changes in here - namely the 300µs reset time - my experience with the WS28** chips is that the latching time is actually far faster than what's quoted in the data sheet (closer to 5µs in many cases - and I still need to finish going through and updating the defaults) - and that time is what's used to determine how long of a pause for handling interrupts is allowed.

alex-Arc:
could be that it is faster than 300µs 👍 , but using the default have given me problems if i push the frame rate up

focalintent:
Ws2811/ws2812 get twitchy when updated at over 400fps.  It's possible that the 2813 has a higher max frame rate (though I doubt it).  That time value is specifically about the maximum time allowed between writing led data (though it is also used to provide a backup to the minimum time in case the refresh rate has been overridden).  

This code is all about to get massively overhauled for the rgbw support.

alex-Arc:
looking forward to the rgbw support. i'm just about to make a pull request for RGBW in octo [https://github.com/alex-Arc/OctoWS2811](url)

unimatrix27:
Line 359 of FastLED.h breaks my project.  
`case OCTOWS2813: { static COctoWS2811Controller<RGB_ORDER,WS2813_800kHz> controller; return addLeds(&controller, data, nLedsOrOffset, nLedsIfOffset); }`

If I comment it out, it works. OCTOWS2813 seems not to be defined anywhere?

alex-Arc:
this is true
> OCTOWS2813 seems not to be defined anywhere?

and fixed here #346

PR #357, opened 2016-10-19
Add support for SODAQ Autonomo (SAMD21J18A)

keestux:
The SODAQ Autonomo is compatible with Arduino Zero, but with different
pin assignments, and slightly different SAMD.

These changes were tested with the SODAQ Autonomo in combination with a WS2812B.


keestux:
Thanks

When do you plan to update library.properties? (Together with a git tag)
I'm sure you know that you have to create a git tag (matching the version in library.properties)? The last known version in Arduino IDE (and on Github) is v3.1.3, but the version in library.properties has gone to v3.1.5

focalintent:
I keep hitting small things that I want to rev before I push it out to the auto update in the ide.

keestux:
OK, no problem. I was just checking.

PR #389, opened 2016-12-18
Fixes ESP8266 parallel output

benjamind:
Moved showRGBInternal into iram cache, increased interrupt wait time to 10. This appears to work but interrupt timing may still require tuning.

### DIFF
--- a/platforms/esp/8266/clockless_block_esp8266.h
+++ b/platforms/esp/8266/clockless_block_esp8266.h
@@ -16,7 +16,7 @@ extern uint32_t _frame_cnt;
 extern uint32_t _retry_cnt;
 #endif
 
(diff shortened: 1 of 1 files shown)


benjamind:
Further testing has shown that even with the wait time increased I am still getting flickering further down the strips. Checking the retry counter I can see that the retries stop at around the 10us mark, but I then get flickering at the start of my strips.

So while this PR fixes the first pixel flicker issue, it does not solve the general flickering issues with interrupts enabled.

benjamind:
I have found it much cleaner to just define INTERRUPT_THRESHOLD to adjust the WAIT_TIME. Best approach thus far for stable 1024 LED output has been to use INTERRUPT_THRESHOLD 1 and FASTLED_INTERRUPT_RETRY_COUNT 0.

While this occasionally will result in a dropped frame, its far preferable to the alternatives, and at 230+ FPS is barely noticeable :)

PR #401, opened 2017-01-15
Teensy 3.6 support

kcouck:
Add support for ARM MK66FX1M0 processor in Teensy 3.6.
Probably works for MK64FX512 in Teensy 3.5 (same pin layout and SPI capabilities)

Only tested with WS2812 led strip (+ 16-way parallel)


kcouck:
Pull request creation in VS added too many changes

PR #402, opened 2017-01-15
Teensy 3.6

kcouck:
Add support for ARM MK66FX1M0 processor in Teensy 3.6.
Probably works for MK64FX512 in Teensy 3.5 (same pin layout and SPI capabilities)

Only tested with WS2811 led strip (+ 16-way parallel)

### DIFF
--- a/led_sysdefs.h
+++ b/led_sysdefs.h
@@ -10,6 +10,9 @@
 #elif defined(__MK20DX128__) || defined(__MK20DX256__)
 // Include k20/T3 headers
 #include "platforms/arm/k20/led_sysdefs_arm_k20.h"
+#elif defined(__MK66FX1M0__)
+// Include k66/T3.6 headers
+#include "platforms/arm/k66/led_sysdefs_arm_k66.h"
 #elif defined(__MKL26Z64__)
 // Include kl26/T-LC headers
 #include "platforms/arm/kl26/led_sysdefs_arm_kl26.h"
--- a/platforms.h
+++ b/platforms.h
@@ -10,6 +10,9 @@
 #elif defined(__MK20DX128__) || defined(__MK20DX256__)
 // Include k20/T3 headers
 #include "platforms/arm/k20/fastled_arm_k20.h"
+#elif defined(__MK66FX1M0__)
+// Include k66/T3.6 headers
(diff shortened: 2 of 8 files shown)


kcouck:
The code was mostly copied from the k20.
The code for the k64 (Teensy 3.5) would almost be identical with the k66 code as well. I'd like to create some m4xxx files (like m0clockless) in the arm/common dir as well to group common code for the ARM Cortex-M4 line.
And maybe instead of creating 3 dirs (k20, k64, k66) for the remaining files, put the specific parts (mostly the pin mappings) into a "teensy" folder?

focalintent:
Thank you for this!  I hope to spend some time merging this back in this weekend.  I'll take care of figuring out what's common and splitting it all together.  (I want to do a little bit more datasheet digging to see how close the k20, k64, and k66's are).  I'm not sure yet on an m4 common - mostly because the exact registers used for clocking are different on some chipsets (the m0 common is possible because it's all hand cycle counted asm (with adjustments for m0+) so there's no io registers that'd be different between platforms)

kcouck:
Ok; I'll wait with that part then :)

I'll just try and check now if the 3rd byte of PORTB can be used as well for parallel processing :)

kcouck:
PORTB output added with some bit shifting to get the output byte into the third byte of the port.
Differentiation is needed between the MASK for the lanes (always on the first x bits of 'width'), and the MASK when actually pushing the data out.

I can't seem to get a decompile listing, but I hope that bithsift is getting optimized out incase we shift 0 bits  (for PORT C and D) :)

kcouck:
Any idea on how you'd like to have T3.5 support added?

adammhaile:
@kcouck You have instructions for using the Teensy 3.6? Used pins 11 and 13 which should be the main SPI, but no luck. Trying to figure out if I'm missing something.

kcouck:
Hello Adam, I haven't tested it with SPI (I only have WS2811's), but the code is almost identical to that for the Teensy 3.2, so I don't see any reason why it wouldn't work...

adammhaile:
Thanks... I did end up getting it working. Though now having trouble
getting it and the SD card to work for some reason. Weird since on the 3.6
the SD is a dedicated interface.

PR #420, opened 2017-03-05
Race condition fix for ARMHardwareSPIOutput::waitFully on Teensy 3.1/2

mikeburg:
SPI fifo byte push and TCF bit clear done using interrupts disable to prevent race condition between
writeByte* routines and waitFully. In additional, waitFully only disables interrupts briefly while waiting for the fifo to empty to allow interrupts to be serviced.


marmilicious:
Would this apply to the Teensy 3.0 as well?

mikeburg:
Definitely. Teensy 3.0, 3.1 and 3.2 use variants of the K20 MCU. In terms of FastLED, it would be the same code regardless which board you used.

focalintent:
Nice catch, thanks!

PR #422, opened 2017-03-08
Linux SPI port

mikeburg:
This adds supports for SPI based (4-wire) led strips under Linux.

### DIFF
--- a/FastLED.cpp
+++ b/FastLED.cpp
@@ -122,7 +122,22 @@ void CFastLED::clearData() {
 
 void CFastLED::delay(unsigned long ms) {
 	unsigned long start = millis();
-        do {
+#ifdef FASTLED_UNIX
+	unsigned long elapse;
+
+	show();
+	elapse = (millis() - start);
+
+	if (elapse >= ms)
+		return;
+
+	elapse = ms - elapse;
+    struct timespec ts;
+    ts.tv_sec = elapse / 1000;
+    ts.tv_nsec = (elapse % 1000) * 1000000;
+    nanosleep(&ts, NULL);
+#else
+    do {
 #ifndef FASTLED_ACCURATE_CLOCK
 		// make sure to allow at least one ms to pass to ensure the clock moves
 		// forward
@@ -134,6 +149,7 @@ void CFastLED::delay(unsigned long ms) {
 #endif
 	}
 	while((millis()-start) < ms);
+#endif
 }
 
 void CFastLED::setTemperature(const struct CRGB & temp) {
@@ -232,7 +248,9 @@ void CFastLED::setMaxRefreshRate(uint16_t refresh, bool constrain) {
   }
 }
 
+#if !defined(FASTLED_UNIX)
 extern "C" int atexit(void (* /*func*/ )()) { return 0; }
+#endif
 
 #ifdef NEED_CXX_BITS
 namespace __cxxabiv1
--- a/Makefile.linux
+++ b/Makefile.linux
@@ -1,0 +1,63 @@
+QUIET ?= 1
+ifeq ($(QUIET),1)
+  AD=@
+else
+  AD=
+endif
+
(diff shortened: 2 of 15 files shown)


focalintent:
Let me talk to Mark about this one - it might be better as a forked, separate project/repo.   I have to admit I'm not thrilled about the idea of including a platform as "supported" while only supporting a subset of the led types that FastLED supports - this has been the primary reason why I haven't pursued any kind of Linux/macOS support in the library.

kriegsman:
Thanks for this PR.  Dan and I will talk about the best way to share it with folks.  Really nice to see how you dug in here to work on the library!

mikeburg:
I understand about not having "complete" support. There could be some annoyance with having to deal with support requests wondering why their WS2812s do not work under Linux.  There has been some interest on the google+ group about this, so it might be useful to someone other than just me. :-)

It is possible to add support for the other LED types (aka 3-wire) by incorporating pieces from other libraries that have already done this. My biggest concern is having to run the program as root to do hardware programming voodoo. (The DMA registers are mapped into the program's address space, and then manipulated directly.) The program in theory could crash the board if some bit of code starts to misbehave and ends up stomping on the mapped registers. Also if I understand things correctly, the on-board audio  cannot be used at the same time due to a conflict over the PWM & hardware timers.

I don't mean to poo-poo it. Its just not as clean as supporting 4-wire LEDs, and would take a little more effort. Its something I could definitely tackle.

I've been enjoying using FastLED very much for a few projects. And I'm happy to give back in some way.

tals:
FWIW, this was helpful to getting FastLED to spit out OpenPixelControl packets (faster development times when coding against a simulator).

I hope this PR lands!

gdcolella:
I also used this to implement a simulator, which has proven to be very helpful in developing displays. Would you have any interest in a PR to add the simulator code to this repo for people to use, even if you don't want to support x86 for controlling strips? It seems like the simulator doesn't have as many of the problems you mentioned, and it seems like it could be helpful!

focalintent:
I think a simulator would also make for an excellent, _separate_ project (again, falls into the realm of code that I don't want to deal with directly supporting/ensuring I don't break with library rewrites/work)

tals:
@focalintent yeah that makes sense wrt the simulator.

I can understand not wanting to have these semi-supported platforms around, but the developer wins here are quite nice.
Using this, I can now easily write tests and use a standard desktop debugger. It's really awesome.

focalintent:
If someone takes this on as a project that they will maintain and keep up to date with the main FastLED repo - I will happily reference it/link to it/point people at it.

mikeburg:
Daniel (and Mark) --

This is one last push on this and after that I promise to keep quiet on the matter. I understand there's an issue of an incomplete port, Yet If this is really more a matter of not wanting to deal with questions and issues regarding the port, that's fine. I'll be happy to jump in deal with those myself. There's no reason for you need to respond to questions regarding someone else code contribution that you are not 100% familiar with or interested in. Also, I'm very happy to coordinate with you on any rewrites in the future to help ensure a smooth code release. Yes, this is your baby --  it doesn't have to be entirely on your head for support.

(The forked repo for the port under my account hasn't been updated for a while because I was hoping this would be picked up.  Yet, if the pull request is a final no-go I'll start keeping parity with the main repo.)

Please contact me directly if you want to talk about this more.

zackees:
We now have a linux x86 build
[linux_native](https://github.com/FastLED/FastLED/actions/workflows/build_linux.yml)

I'm open to a PR if you can make a clean CL.

zackees:
Re-opening because the community wanted this feature so bad that they forked the repo a bunch of times. On the testing side, we need this platform native stuff so that we can unlock unit tests to verify correctness by contributors.

These desktop platform tests also test that FastLED can continue to compile outside of an Arduino build environment.

PR #437, opened 2017-04-07
Adjust millis on Teensy LC

PaulStoffregen:
WS2812 on Teensy LC isn't adjusting the millis count when interrupts are disabled.

Here's a quick fix.  Conversation here:

https://plus.google.com/109760919253841662611/posts/LjcKxfxMKBb

### DIFF
--- a/platforms/arm/kl26/clockless_arm_kl26.h
+++ b/platforms/arm/kl26/clockless_arm_kl26.h
@@ -20,15 +20,18 @@ public:
     mPort = FastPinBB<DATA_PIN>::port();
   }
 
-	virtual uint16_t getMaxRefreshRate() const { return 400; }
+  virtual uint16_t getMaxRefreshRate() const { return 400; }
 
   virtual void showPixels(PixelController<RGB_ORDER> & pixels) {
     mWait.wait();
     cli();
-    if(!showRGBInternal(pixels)) {
+    uint32_t clocks = showRGBInternal(pixels);
+    if(!clocks) {
       sei(); delayMicroseconds(WAIT_TIME); cli();
-      showRGBInternal(pixels);
+      clocks = showRGBInternal(pixels);
     }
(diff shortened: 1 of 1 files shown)


focalintent:
Going to take this for now - eventually I want to kick things so that the teensy lc can run with interrupts enabled (in which case this code will be ifdef'd to cases where we aren't running with interrupts)

PaulStoffregen:
On my (very long) TODO list is writing a nonblocking PWM+DMA driver for WS2812 on Teensy LC.  I don't have a solid time frame in mind, but hopefully when it's working you might consider merging it into FastLED?

PaulStoffregen:
Running with interrupts will be really tough on Cortex-M0+.  If anyone can do this, it's you.  I tried some time ago and never got it to work well.  :-(

focalintent:
Either that or i'll do something similar that I did with the OctoWS2811 driver (I still need the clockless support code because there's a bunch of non-WS2812 3-wire led chipsets that I support) -- putting aside the 1001 paper cuts that I need to do for the library the next major piece is the RGBW/RGB16/RGBW16 work - which is a rewrite on many levels - but immediately after that is a model for DMA/interrupt/background LEDController objects so I can start supporting dma output of various sorts on a variety of platforms

focalintent:
It works well enough on the nrf51's m0 core to allow bluetooth and time interrupts to fire (that was where I initially developed the allow interrupts code for arm that I then moved to other platforms, if I remember the order right).  The one piece that I have defined for the nrf51 that I haven't done for the teensy-lc yet is the small block of code that detects whether or not an interrupt has run for too long to trigger either a retry of the frame or at least stopping writing out the frame (because if the interrupt runs too long, the leds latch/reset and when the library thinks it's writing out, say, leds[60] it'll be the first led on the strip that updates) - and I think the LC's clock rate is higher than the nrf51.

PaulStoffregen:
Wow, that's pretty awesome.  Let me know if you need any more LC's for testing?

PR #464, opened 2017-06-09
- added a few methods to accept an offset parameter

tweep:
this is useful if you keep all your LEDs in a single, large array.
  I do this to save dynamic memory -  because I have LED strands with
  different lengths; declaring a two-dimensional LED matrix will
  reserve too much memory; so i put them all in one single one-dimensional
  array
  New methods which accept an "offset parameter" are :
  - fill_solid
  - fill_rainbow
  - fadeToBlackBy
  - nscale8

### DIFF
--- a/colorutils.cpp
+++ b/colorutils.cpp
@@ -26,6 +26,15 @@ void fill_solid( struct CHSV * targetArray, int numToFill,
 }
 
 
+void fill_solid( struct CRGB * leds, int offset, int numToFill,
+                 const struct CRGB& color)
+{
+    for( int i = 0; i < numToFill; i++) {
+        leds[offset+i] = color;
+    }
+}
+
+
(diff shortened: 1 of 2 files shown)


mikeburg:
Hi,

Is there any reason why you could not simply pass the pointer to starting offset led  to these routines? 

Example: fill_solid(&leds[offset], count, CRGB::Green);

-- Michael

focalintent:
leds + offset should also work.  

This area of the library is due for a major overhaul to make use of the pixel sets, which will change up all of this - but make doing things like this far easier :)

MFornander:
Like @mikeburg said, this can literally by done by replacing "buffer" with "buffer+offset" in any method.  I don't think we need to add an argument here.

tweep:
Ooops, my bad; pointers work of course. 

Thanks !

PR #474, opened 2017-08-05
Support for ESP32

samguyer:
Credit to Rina Shkrabova for the first cut. Overall, this code is almost identical to the 8266 support, except for the pin definitions and the handling of interrupts. I put in calls to ets_intr_lock and ets_intr_unlock.

### DIFF
--- a/fastled_delay.h
+++ b/fastled_delay.h
@@ -33,6 +33,13 @@ public:
 ////////////////////////////////////////////////////////////////////////////////////////////
 
 // Default is now just 'nop', with special case for AVR
+
+// ESP32 core has it's own definition of NOP, so undef it first
+#ifdef ESP32
+#undef NOP
+#undef NOP2
+#endif
+
 #if defined(__AVR__)
 #  define NOP __asm__ __volatile__ ("cp r0,r0\n");
 #  define NOP2 __asm__ __volatile__ ("rjmp .+0");
--- a/led_sysdefs.h
+++ b/led_sysdefs.h
@@ -25,6 +25,8 @@
 #include "platforms/arm/d21/led_sysdefs_arm_d21.h"
 #elif defined(ESP8266)
 #include "platforms/esp/8266/led_sysdefs_esp8266.h"
+#elif defined(ESP32)
+#include "platforms/esp/32/led_sysdefs_esp32.h"
 #else
 // AVR platforms
 #include "platforms/avr/led_sysdefs_avr.h"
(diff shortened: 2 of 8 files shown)


focalintent:
Sam - real quickly - with this code checked in can you do a couple quick tests of compiling for esp8266, and arm based arduino, and one of the teensy 3.x platforms?  (I don't have a good test compile setup going) -- but if your changes aren't interfering with those (and it looks like it shouldn't I'm just being slightly paranoid here) -- I'll take the pull request) -- thank you for this!

samguyer:
I don't think I have a Teensy to test on, but I can certainly double-check with an ESP8266. Any particular kind of test?

focalintent:
Actually just need compile tests done (and i'm installing software here finally so I may be able to do the compile tests myself in a few -- assuming i can checkout your repo for the test -- hmmm, I wonder if I can re-target this pull request to a different branch) -- your changes shouldn't be impacting any running code on other platforms, I just know that sometimes when I add a new platform, I miss an endif or some other kind of guard somewhere that breaks other platforms from building :/

marmilicious:
@focalintent   I just tested DemoReel100 with this branch with a Teensy 3.0 and LPD8806 (32 pixels) and then also APA102 (38 pixels).  Using Teensyduino 1.30.  Looks to be running run fine.  Let me know if you'd like any other specific tests.

focalintent:
@marmilicious with sam's changes merged in?  (I haven't accepted the pull request yet) -- and thank you for checking!

marmilicious:
I grabbed from here thinking that would work, but maybe that's incorrect.
https://github.com/FastLED/FastLED/tree/b438131507adb3c3be068f9efe08543ee47c735b

focalintent:
Perfect, thank you!

samguyer:
Oh OK. It definitely compiles for both ESP32 and ESP8266. The only thing I had to change that was kind of sketchy was the way NOP and NOP2 are defined. The currently ESP32 core defines them like this:

#define NOP() asm volatile ("nop")

Whereas FastLED defines them like this:

#define NOP asm volatile ("nop")

which then doesn't compile because of the conflicting definitions. So for ESP32, I first undef NOP and then let FastLED redefine it.

samguyer:
So, I've run into some weird behavior with libraries that make heavy use of interrupts (for example, motion sensors that generate interrupts when the orientation changes). I'm not sure what's going on, but I am able to reproduce it.

focalintent:
It’s possible that the timing thresholds are wrong for deciding when/how to re-transmit frames — try either disabling interrupts during WS28xx output entirely or dropping the retry count to 1 — i’m can’t easily lookup code right now but if you look in fastled_config.h you’ll see the defines you need to use (put them before you include FastLED.h) and see if either of those help.

PR #487, opened 2017-08-25
Fix FastLED version

NicoHood:


focalintent:
Is anyone even using FASTLED_VERSION?  (Other than demoreel, that is)

NicoHood:
Possibly yes. It really makes sense to check for a minimum required version number if you depend on a special new fix.

BTW: Are you able to cherry pick my commits from [here](https://github.com/FastLED/FastLED/compare/master...NicoHood:makefile?expand=1), except the first or should I create a separate PR for that?

I got the whole code compiling fine now with my makefile. However I still got no hardware to test. I will publish the stuff soon, so maybe someone else can test it for me.

I also made a makefile for the FastPin header subset, so you can use fast pin access. For further versions you might want to outsource this library maybe, I find it extremely useful.

Issue #485, opened 2017-08-23
Makefile Compilation


NicoHood:
Hi,
I try to compile FastLED based on a [DMBS makefile](https://github.com/abcminiuser/dmbs) for an atmega32u4 (no Arduino environment).

I got the code compiling with a few tricks, but I have not tested it, and only included the FestLED.h file, nothing more. It would be nice if you could give me some guidiance what is required to use FastLED with a pure makefile. I am willing to write a DMBS makefile (and patches via PR, if required) so others can use that too.

My findings are based on FastLED 3.1.6

in pixelset.h abs() was undefined. I added used:
```cpp
#include "FastLED.h"
#ifndef ARDUINO
#include "stdlib.h"
#endif
```

In led_sysdefs_avr.h I had to uncomment:
```cpp
// Define the register types
//#if defined(ARDUINO) // && ARDUINO < 150
typedef volatile       uint8_t RoReg; /**< Read only 8-bit register (volatile const unsigned int) */
typedef volatile       uint8_t RwReg; /**< Read-Write 8-bit register (volatile unsigned int) */
//#endif
```

FastLED requires some Arduino functions for pinmapping (as a fallback). I disabled it, and then no pinMode functions (etc.) were not required anymore. Is this the correct way? The pins for an atmega 32u4 should be defined already, I am using the same layout.
```make
CC_FLAGS     =  -D FASTLED_NO_PINMAP
```

I also upgraded to c++11 (something I found on some website, more compiler flags might be required):
```make
CPP_STANDARD = gnu++11
CPP_FLAGS = -fno-threadsafe-statics
```

And finally I added some more code to suppress some more errors. I know that I need to reimplement micros() later.
```cpp
typedef uint8_t byte;
typedef bool boolean;

static uint32_t micros(){
    return 0;
}
```

---

Now I started to add other cpp files to my makefile. I made the following changes:

colorutils has no pow() function, so I added 
```cpp
#include <math.h>
```

noise.cpp missing memset:
```
#include <string.h>
```

power_mgt.cpp, dont use digitalwrite. I disabled the whole function `calculate_max_brightness_for_power_mW`

I end up getting the error:
```
/tmp/cc8iIq0B.ltrans0.ltrans.o:(.rodata+0x4): undefined reference to `__cxa_pure_virtual'
```

I fixed those by hints from this thread:
http://www.avrfreaks.net/forum/avr-c-micro-how

h file:
```cpp
void * operator new(size_t size);
[...]

PR #490, opened 2017-09-01
Makefile Compilation Fix

NicoHood:
Also see https://github.com/FastLED/FastLED/issues/485

I've made very clean commits and everyone should be simple and understandable on its own. The inclusion of the additional headers is safe, as they are included by Arduino.h.

**All changes were tested with an  Arduino environment. The changed LED code for the power management also!**

### Benefits
* Less Arduino overhead (240 bytes for blink) with makefiles
* Better control with makefile environments
* Still Arduino compatible as before

### DIFF
--- a/power_mgt.cpp
+++ b/power_mgt.cpp
@@ -125,7 +125,7 @@ uint8_t calculate_max_brightness_for_power_mW( uint8_t target_brightness, uint32
     if( requested_power_mW < max_power_mW) {
 #if POWER_LED > 0
         if( gMaxPowerIndicatorLEDPinNumber ) {
-            digitalWrite(gMaxPowerIndicatorLEDPinNumber, LOW);   // turn the LED off
+            Pin(gMaxPowerIndicatorLEDPinNumber).lo(); // turn the LED off
         }
 #endif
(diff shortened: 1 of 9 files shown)


power_mgt.cpp
```diff
@@ -125,7 +125,7 @@ uint8_t calculate_max_brightness_for_power_mW( uint8_t target_brightness, uint32
     if( requested_power_mW < max_power_mW) {
 #if POWER_LED > 0
         if( gMaxPowerIndicatorLEDPinNumber ) {
-            digitalWrite(gMaxPowerIndicatorLEDPinNumber, LOW);   // turn the LED off
+            Pin(gMaxPowerIndicatorLEDPinNumber).lo(); // turn the LED off
```
ladislas: where does that come from? is it part of the avr stdlib?

ladislas:
it looks great and promising!

power_mgt.cpp
NicoHood: @ladislas That is a FastLed class inside fastpin.h. Also checkout FastPin<pin>, which is even better, but not suitable in this flexible case.

PR #495, opened 2017-09-10
Fix compilation under ESP32

BlackEdder:
These additions allow me to compile FastLED for the esp32.

### DIFF
--- a/FastLED.cpp
+++ b/FastLED.cpp
@@ -230,7 +230,9 @@ void CFastLED::setMaxRefreshRate(uint16_t refresh, bool constrain) {
   }
 }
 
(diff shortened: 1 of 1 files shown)


tbnobody:
I can confirm that this code change is required to compile it successfull (using esp-idf)

focalintent:
What I want is someone to confirm that these changes _don't_ break things for folks building esp32 code in an arduino ide - then i'll merge these back in - otherwise, i'll need a PR that doesn't break building under the arduino ide first.

focalintent:
A related change has already been made to this area of code, no longer need the PR

PR #498, opened 2017-09-14
complet sheet with FastLED commands

alexerer1:
I just started my LED "adventure" on my arduino UNO, but i cant find a complet list over the commands available in the FastLED library, so i am slowly working my way through the examples but it is a very slow process.

does anybody have a worksheet with a list of commands avaiable in FastLED library for arduino?


focalintent:
There's documentation here - http://fastled.io/docs/3.1/ - as well as a lot of notes on the wiki - http://fastled.io/wiki -- also I have no idea what your'e trying to do with this pull request so i'm going to cancel it out.

PR #510, opened 2017-10-03
fixed gemma and trinket m0 support - tested all pins on both boards

ladyada:
Adding Trinket M0 and Gemma M0 support - these little boards <3 FastLED but needed their pins defined :)


focalintent:
Thank you!  I'll pull this and spin up a new release tonight!

PR #512, opened 2017-10-18
Add support for ATmega328 non-picopower hardware pins

tuna-f1sh:
Should be a fairly simple pull request as it's only one line. The ATmega328 (not p - picopower) is not currently included in the logic check with the others in the same series. I've added it so that the hardware pin map is correctly assigned and it does not revert to software mapping.

I had a project fail to compile when I had a ATmega328 part so I think this will help others.

John.


focalintent:
Thank you!

PR #517, opened 2017-11-01
Add WS2812Serial driver

PaulStoffregen:
This adds a 3rd party driver for the non-blocking WS2812Serial library.

Also uses the arm/k66 platform to support Teensy 3.5.


focalintent:
Thank you!  (I've mostly avoided putting things like this in because I haven't found a good way to map them to the variety of chipset timings out there -- but as a wrapper around another library is a good stop gap in the meantime)

PaulStoffregen:
Yeah, this seems like a pretty good way if you're not looking to put tons of chip-specific code into FastLED.

The only downside I see is the buffers need to be allocated with malloc().  Would be nice if they could show up in the Arduino memory usage summary somehow.

PR #530, opened 2017-11-27
Add support for Circuit Playground Express

ladyada:
https://github.com/FastLED/FastLED/issues/500#issuecomment-347079839


focalintent:
Thank you so much!

ladyada:
hiya can you do an update/ release for the Arduino library manager? that way this update will get distributed - thank you :)

PR #533, opened 2017-11-30
Fix typo in fastpin_esp32 header

tullo-x86:
Bottom of file should expand the `FASTLED_NAMESPACE_END` macro, not `#define` it.

This is the sort of bug that could go undetected for years until you reorder your `#include`s one day.


focalintent:
Namespaces are only used on the particle boards - so it’s “safe” here for now - but good catch!

tullo-x86:
Maybe one of these days I'll contribute more than a rogue `#warning` or an unused macro redefinition 😅

PR #535, opened 2017-12-03
Added Fastpin InputPullup() and get() functions

NicoHood:
Currently there are only operations to set a pin, but not to read it. It is very simple to add this functionality. I only implemented it for AVR, but it should not influence any of the other architectures if its missing there.

The source has tabs, it should be "untabified". Maybe other files as well. Is it okay when I do it inside the PR?

### DIFF
--- a/fastpin.h
+++ b/fastpin.h
@@ -206,14 +206,14 @@ template<uint8_t PIN> class FastPin {
 
 	static_assert(validpin(), "Invalid pin specified");
 
-	static void _init() {
-	}
+	static void _init() { }
 public:
 	typedef volatile RwReg * port_ptr_t;
 	typedef RwReg port_t;
 
 	inline static void setOutput() { }
 	inline static void setInput() { }
+    inline static void setInputPullup() { }
 
 	inline static void hi() __attribute__ ((always_inline)) { }
 	inline static void lo() __attribute__ ((always_inline)) { }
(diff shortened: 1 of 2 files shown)


NicoHood:
I've added another commit to also create In/OutputPin classes with the Fast template variant. Any chance to get this merged?

fastpin.h
```diff
@@ -236,6 +238,21 @@ template<uint8_t PIN> class FastPin {
 
 #endif
 
+template<uint8_t PIN> class FastOutputPin : public FastPin<PIN> {
```
focalintent: I don't think it makes sense to have these - especially since setInput can be called on a FastOutputPin and vice versa.  This doesn't really save a whole lot.

fastpin.h
```diff
@@ -206,14 +206,14 @@ template<uint8_t PIN> class FastPin {
...
-	static void _init() {
-	}
+	static void _init() { }
 public:
 	typedef volatile RwReg * port_ptr_t;
 	typedef RwReg port_t;
 
 	inline static void setOutput() { }
 	inline static void setInput() { }
+    inline static void setInputPullup() { }
```
focalintent: This functionality should probably get propagated to all the pin classes across platforms before merging this PR in.

fastpin.h
NicoHood: I know, but those also exist for the normal class(non template) variants. And I had a use case where I was missing those for templates. Still should remove them?

focalintent: Ah - not sure why those are still in there, they shouldn't be - not least of all because they are mutable, so they're a bit of a lie : ) -- i'll clean that up at some point, in the meantime, I don't want to encourage spreading their use by adding more variants.  

The new methods being made available across all the platforms is actually the more important one for me, though.

zackees:
Seems like a good idea but this PR is a little stale now and it's only implemented for the AVR so not cross platform compatible. Also, not clear why this is necessary in the first place.

PR #539, opened 2017-12-11
fixes build for ESP32s

ian7:
it seems that definition of __cxa_pure_virtual is already included in ESP32 build chain

PR #563, opened 2018-02-07
Don't archive library with @PlatformIO Builder

ivankravets:
https://github.com/FastLED/FastLED/pull/539


focalintent:
Is examples no longer valid in the json?

ivankravets:
PlatformIO Library Crawler detects them automatically.

PR #571, opened 2018-02-17
Use consistent include syntax/formatting

per1234:
- Angle brackets for external includes
- Space after `#include` directive


kriegsman:
Thanks for the cleanup PR!
The only item I have a question about in this PR is the change from "./dmx" to "dmx.h".  I wonder if the original included the "./" for a reason, e.g. some other dmx.h being accidentally found in a search path somewhere.  
Any thoughts?

per1234:
I puzzled over that one too. My understanding is that the #include "dmx.h" syntax should always search the current folder first before the rest of the include search paths so I can't see how that would be necessary.

https://gcc.gnu.org/onlinedocs/cpp/Include-Syntax.html
>#include "file"
    This variant is used for header files of your own program. **It searches for a file named file first in the directory containing the current file**

I'm happy to remove that part of the PR if you like.

PR #592, opened 2018-04-04
Use yield while delaying

wavesoft:
In general, it's not a good idea to block the chip without allowing background operations to be performed, but in my case I had TCP issues using this library with an ESP-12F chip and the `espressif8266` arduino stack.

This PR fixes this by allowing background operations while waiting.

### DIFF
--- a/FastLED.cpp
+++ b/FastLED.cpp
@@ -131,7 +131,7 @@ void CFastLED::delay(unsigned long ms) {
 		show();
 		yield();
 	}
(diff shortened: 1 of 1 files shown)


wavesoft:
Sorry, I was way too quick to jump in conclusions, don't even bother to look on what's changed 😄

focalintent:
Yeah - there’s something else that the esp does that lives outside of loop I think.

PR #596, opened 2018-04-16
WAIT_TIME=20 to allow interrupts without breaking matrix output.

marcmerlin:
WAIT_TIME set to 5 was too short and FastLED aborted frames too early
and restarted them incorrectly.
(as explained by Daniel Garcia in
https://plus.google.com/communities/109127054924227823508 )

Tested on ESP8266 with 768 LED matrix in both serialized output
and 3x 256 output.


focalintent:
Change it to 50 (to be inline with what the other clockless_* files are using) and i'll pull this in.  Thanks!

PR #604, opened 2018-05-01
Add new platform "PIN_DEFINITIONS_MANUALLY_CONFIGURED"

achew22:
PIN_DEFINITIONS_MANUALLY_CONFIGURED is used when none of the preexisting board definitions align with a custom board you've designed for a one-off. Generally these one-off projects are not worth contributing back to FastLED by virtue of their non-common use.


focalintent:
This will fall in under the null platform work that I'll be checking in shortly.

PR #625, opened 2018-07-09
Rewrite of fastpin_esp32.h -- take 2

samguyer:
I just fixed the code formatting, which was ugly.


FastLED.cpp
```diff
@@ -239,7 +239,7 @@ extern "C" void yield(void) { }
 #ifdef NEED_CXX_BITS
 namespace __cxxabiv1
 {
-	#ifndef ESP8266
+	#if !defined(ESP8266) && !defined(ESP32)
 	extern "C" void __cxa_pure_virtual (void) {}
```
focalintent: you know, it dawns on me that I could probably make this a weak symbol...

fastled_delay.h
```diff
@@ -34,6 +34,12 @@ template<int WAIT> class CMinWait {
 
 // Default is now just 'nop', with special case for AVR
 
+// ESP32 core has it's own definition of NOP, so undef it first
+#ifdef ESP32
+#undef NOP
```
focalintent: Gyah -- there's a reason why, with the CRGBW/CRGB16 re-write I'm re-doing all the macros to be prefaced with FASTLED_ or FL_ ...

platforms/esp/32/fastpin_esp32.h
```diff
@@ -24,68 +11,90 @@ template<uint8_t PIN, uint32_t MASK> class _ESPPIN {
   inline static void setOutput() { pinMode(PIN, OUTPUT); }
   inline static void setInput() { pinMode(PIN, INPUT); }
 
-  inline static void hi() __attribute__ ((always_inline)) { if(PIN < OUTPUT_PIN_LIMIT) { _GPB0._GPOS = MASK; } }
-  // inline static void hi() __attribute__ ((always_inline)) { gpio_set_level((gpio_num_t)PIN, HIGH); }
+  inline static void hi() __attribute__ ((always_inline)) { 
+      if (PIN < 32) GPIO.out_w1ts = MASK;
+      else if(PIN < 34) GPIO.out1_w1ts.val = MASK;
```
focalintent: Can you make this >= 32?  This way if an esp32 board comes out that exposes more pins on the secondary controller, this will still work.  Ditto for the other set below

platforms/esp/32/fastpin_esp32.h
```diff
@@ -24,68 +11,90 @@ template<uint8_t PIN, uint32_t MASK> class _ESPPIN {
...
-  // inline static void set(register port_t val) __attribute__ ((always_inline)) { gpio_set_level((gpio_num_t)PIN, val); }
+  inline static void set(register port_t val) __attribute__ ((always_inline)) {
+      if (val) hi();
+      else lo();
+  }
 
   inline static void strobe() __attribute__ ((always_inline)) { toggle(); toggle(); }
 
-  inline static void toggle() __attribute__ ((always_inline)) { if (PIN < OUTPUT_PIN_LIMIT){ _GPB0._GPO = MASK; } }
+  inline static void toggle() __attribute__ ((always_inline)) {  }
```
focalintent: Needs an implementation - e.g. something like:

```
  GPIO_OUT_REG ^= MASK;
```

platforms/esp/32/fastpin_esp32.h
```diff
@@ -24,68 +11,90 @@ template<uint8_t PIN, uint32_t MASK> class _ESPPIN {
...
   inline static void hi(register port_ptr_t port) __attribute__ ((always_inline)) { hi(); }
   inline static void lo(register port_ptr_t port) __attribute__ ((always_inline)) { lo(); }
   inline static void fastset(register port_ptr_t port, register port_t val) __attribute__ ((always_inline)) { *port = val; }
 
-  inline static port_t hival() __attribute__ ((always_inline)) { if (PIN<OUTPUT_PIN_LIMIT) { return GPIO_OUT_REG | MASK;    }}
-  inline static port_t loval() __attribute__ ((always_inline)) { if (PIN<OUTPUT_PIN_LIMIT) { return GPIO_OUT_REG & ~MASK;   }}
-  inline static port_ptr_t port() __attribute__ ((always_inline)) { if(PIN<OUTPUT_PIN_LIMIT) { return &_GPB0._GPO;   }}
-  inline static port_ptr_t sport() __attribute__ ((always_inline)) { if (PIN<OUTPUT_PIN_LIMIT) {return &_GPB0._GPOS; }}
-  inline static port_ptr_t cport() __attribute__ ((always_inline)) { if (PIN<OUTPUT_PIN_LIMIT) {return &_GPB0._GPOC; }}
+  inline static port_t hival() __attribute__ ((always_inline)) { return MASK; }
```
focalintent: hival/loval are used as an optimization in bitbang'd spi output - to avoid a few rounds of bitwise operators.  hival is the "write this value to the I/o register to make this pin high" while loval is the "write this value to the I/o register to make the pin low" -- so the value that's returned is the full state of the I/o register, adjusted for whether we want this pin to be high or low - for example:

```
	inline static port_t hival() __attribute__ ((always_inline)) { return ((PIN<32) ? GPIO.out : GPIO.out1) | _MASK; }
	inline static port_t loval() __attribute__ ((always_inline)) { return ((PIN<32) ? GPIO.out : GPIO.out1)  & ~_MASK; }
```

Meanwhile port returns the register itself - as in a pointer to the register:

```
    inline static port_ptr_t port() ((always_inline)) { return (PIN < 32) ? &GPIO.out : &GPIO.out1); }
```

platforms/esp/32/fastpin_esp32.h
```diff
@@ -24,68 +11,90 @@ template<uint8_t PIN, uint32_t MASK> class _ESPPIN {
...
-// _DEFPIN_ESP32(32,32); _DEFPIN_ESP32(33,33);
-
-#define PORTA_FIRST_PIN 32
-// The rest of the pins - these are generally not available
-// _DEFPIN_ESP32(11,6);
-// _DEFPIN_ESP32(12,7); _DEFPIN_ESP32(13,8); _DEFPIN_ESP32(14,9); _DEFPIN_ESP32(15,10);
-// _DEFPIN_ESP32(16,11);
-
-#endif
+_DEFPIN_32_33_ESP32(32);
```
focalintent: Note that according to https://github.com/espressif/esp-idf/blob/master/components/soc/esp32/include/soc/gpio_struct.h - this can go up to pin 39 - even if no boards are exposing anything above 32/33 yet.

focalintent:
Also - what's a bit strange is that this diff appears to be happening against an older version of master - which may be the reason for the conflicts that are identified by GitHub.

platforms/esp/32/fastpin_esp32.h
samguyer: No problem. I copied that log directly from the ESP-IDF core, though. I think the reason they write it like that is that there are pins 35-39, but they are input only.

platforms/esp/32/fastpin_esp32.h
samguyer: I need to think about this one.

platforms/esp/32/fastpin_esp32.h
samguyer: The way it seems to work on the ESP32 is that you use the same mask regardless of whether you want to set a pin high or low -- the difference is *which* register you write that mask to. That is, hival and lowval are the same, the difference is sport vs cport. The port() method will never work for this reason.

platforms/esp/32/fastpin_esp32.h
samguyer: Pins 34-39 are input only.

samguyer:
I resolved the conflicts, but I need to decide what to do about the issues you cite above.

platforms/esp/32/fastpin_esp32.h
focalintent: Yes - most higher end arm/32 bit cores have all _three_ types of register - a 32-bit register that gives the current hi/lo values for all 32 pins on that port, a 32-bit register that, when you write to it turns on any pin that has a corresponding 1 bit in the written value, and a 32-bit register that, when you write to it, turns _off_ any pin that has a corresponding 1 bit in the written value. - for the ESP32 there's GPIO.out - which is the port register - which, if I understand correctly you _can_ use to read the state of an output only pin (whether it is hi or low) - as well as bulk set the value (both hi and low) of multiple pins at the same time.  This is along side the out_w1tc which is the "write 1's to clear pins" and the out_w1ts "write 1's to set pins" (the c and s suffixes on those make it clear which one they are)- so, if you use GPIO.out then hival/lowval and port _will_ in fact work the way that the rest of the library is expecting it to.

From the gpio_struct header file:

```
    uint32_t out;                                   /*GPIO0~31 output value*/
    uint32_t out_w1ts;                              /*GPIO0~31 output value write 1 to set*/
    uint32_t out_w1tc;                              /*GPIO0~31 output value write 1 to clear*/
```

There are also a variety of other possible port options.  Some arm platforms have a toggle register, where any bit  in the 32-bit value that you write that has a 1 will flip the state of the associated pin (hi to low or low to high).  Then there's also yet another register writing technique called "bit-banding" - where a virtual address space is exploded out so that each pin has 4 32-bit memory locations - an out, a set, a clear, and a toggle - and writing any value will do the appropriate action with the pin - the nice thing about this is _any_ non-zero value triggers the action (setting the value, setting the pin hi, setting it low, toggling it) - which is a nice way to short circuit the whole "if (someValue != 0)  GPIO.set = MASK; else GPIO.clear = MASK;" - you can just do bitband[pin].val = someValue;

So - yes, port, hival, and lowval can work the way it does for other platforms.  The GPIO.out register is how, among other things, you would ask "Is pin 24 hi or low right now?" (the w1tc and w1ts registers are effectively write only, there's nothing meaningful to read from them).

platforms/esp/32/fastpin_esp32.h
focalintent: For now : )

platforms/esp/32/fastpin_esp32.h
focalintent: This is where you would use the GPIO.out register - honestly I'm disappointed that Espressif didn't also add a toggle gpio register - but their gpio architecture seems a little hacky at times.  Anyway - this would literally just be:

```
  if(PIN < 32) { GPIO.out ^= MASK; } else { GPIO.out1.val ^=MASK; }
```

platforms/esp/32/fastpin_esp32.h
focalintent: Espressif's internal headers seem to indicate that 35-39 could also be for out - from https://github.com/espressif/esp-idf/blob/master/components/soc/esp32/include/soc/gpio_struct.h -- 

```
    union {
        struct {
            uint32_t data:       8;                 /*GPIO32~39 output value*/
            uint32_t reserved8: 24;
        };
        uint32_t val;
    } out1;
    union {
        struct {
            uint32_t data:       8;                 /*GPIO32~39 output value write 1 to set*/
            uint32_t reserved8: 24;
        };
        uint32_t val;
    } out1_w1ts;
    union {
        struct {
            uint32_t data:       8;                 /*GPIO32~39 output value write 1 to clear*/
            uint32_t reserved8: 24;
        };
        uint32_t val;
    } out1_w1tc;
```

samguyer: Totally up to you. I'm using the Espressif implementations of the various pin functions as my guide (e.g., their version of digitalWrite) -- it seemed like the most likely to work. I can try something different, but I'll have to do some testing to make sure it works.

platforms/esp/32/fastpin_esp32.h
samguyer: Reading up on this question: it is definitely possible to implement all the functions by accessing the GPIO.out register. The only downside I see (which is very minor) is that it is slower because it requires three instructions: read, bit-op, write. Using the GPIO.out_wtx registers requires only a write.

samguyer:
OK, take a look now. The only minor issue I see is that there is an implicit constraint that the values returned by hival() and loval() should only be used with the register returned by port(), not the ones return by sport() and cport().

I'm also not sure if fastset() will work properly, since it doesn't read the register state and mask in only the bits that should be changed.

platforms/esp/32/fastpin_esp32.h
focalintent: Right - which is why hi should use the out_w1ts and lo should use the out_w1tc - but port/hival/lowval specifically work with the out register.  There's other code in FastLED (and other private code I have that uses FastPin) that makes use of this - and again, something low level like FastPin I want to work reliably the same way across all the platforms, which is why i'm being a stickler about this.

platforms/esp/32/fastpin_esp32.h
```diff
@@ -24,68 +11,92 @@ template<uint8_t PIN, uint32_t MASK> class _ESPPIN {
...
 };
 
-#define _DEFPIN_ESP32(PIN, REAL_PIN) template<> class FastPin<PIN> : public _ESPPIN<REAL_PIN, (1<<(REAL_PIN & 0xFF))> {};
-
+#define _DEFPIN_ESP32(PIN)  template<> class FastPin<PIN> : public _ESPPIN<PIN, ((uint32_t)1 << PIN)> {};
+#define _DEFPIN_32_33_ESP32(PIN) template<> class FastPin<PIN> : public _ESPPIN<PIN, ((uint32_t)1 << (PIN-32))> {};
 
-#ifdef FASTLED_ESP32_RAW_PIN_ORDER
+_DEFPIN_ESP32(0);
+// _DEFPIN_ESP32(1); Using TX causes flashiness when uploading
```
focalintent: don't block the use of these pins -- yes rx/tx are used if you're flashing via USB - but not if you're flashing wirelessly - also, a deployed device will spend far more of it's life running than being flashed.

platforms/esp/32/clockless_esp32.h.orig
```diff
@@ -0,0 +1,786 @@
...
+
+template <int DATA_PIN, int T1, int T2, int T3, EOrder RGB_ORDER = RGB, int XTRA0 = 0, bool FLIP = false, int WAIT_TIME = 5>
+class ClocklessController : public CPixelLEDController<RGB_ORDER>
+{
+    // -- RMT has 8 channels, numbered 0 to 7
+    rmt_channel_t  mRMT_channel;
+
+    // -- Store the GPIO pin
+    gpio_num_t     mPin;
+<<<<<<< HEAD
```
focalintent: not sure if you meant to leave this checked in as part of the PR - especially in a conflicted state?

focalintent:
The whole point of fastset is pre-computing the hi/lo register values to write to the out register before pushing out a set of led data so that you weren't doing the the read/bit-operation/write cycle of writing using the out register - plus it works on platforms where there is only one out register (and no dedicated set/clear registers).

The other point of fastset is allowing the setting of multiple pins simultaneously in one cycle (or two if the underlying hardware is two cycle writes to i/o registers) vs. one (or two) per pin that you want to set the state of - and using fastset you don't need to be setting the pins to the same value (e.g. you can set a few high while setting others low).

sport/cport are there for a combination of completness, but also because there can also be value in bulk setting/clearing pins, and wanted a standard way for that in FastPin.

platforms/esp/32/clockless_esp32.h.orig
samguyer: I don't see that in my branch. Maybe I checked in the orig file by accident.

samguyer:
I'm not sure I understand your logic about avoiding the read/bit-op/write cycle. Given that there are a lot of pins, you don't want to change the state of pins that are not doing any FastLED stuff (other sensors and actuators). So, to change a single pin using just the GPIO.out register you have to read its current value, combine that with the new value, then write it back. Seems like more operations than just writing to the set/clear registers. I totally get the consistency argument, but I don't see the performance argument.

It might be worth noting in the code that the values provided by hival and loval will not necessarily work with the registers returned by sport and cport.

focalintent:
> I'm not sure I understand your logic about avoiding the read/bit-op/write cycle. 

Again - this is an optimization that I use in some cases cases where I know for sure nothing else will be modifying the pins during the output.  I read the state of the port register - pre-compute the values that I want for the various pin states that will happen as I do the output - and then write those pre-computed values out to the port register while writing out the data - so outside of that pre-computing, it's just a series of writes.  And again, I want the FastPin class's functionality to be consistent across platforms even if you don't see an immediate use for it.

I can add a comment to the code that the hival/loval values are meant to be used with just the port register - but I know how port/sport/cport are supposed to work along with hival/loval (which is why I keep asking you to make them _work that way_) - i've used the sport/cport values in the past to determine whether or not two pins are on the same port, and then combine masks appropriately when setting values to do it in a single write operation.  

The other reason for having these methods is because I can then get that value and store it in a register - for example on arm systems, port operations end up being writes to known memory addresses - gcc often represents these as "constant load to register ; write mask to register2 ; write register2 to the address in register;" - and then re-uses register, so the next port write ends up getting an extra 2 cycles attached to it loading the constant.  When I can pre-compute/pre-cache values I can keep them in registers and replace a 2 cycle read, a 1 cycle constant load, and a 2 cycle write with just a single 2 cycle write.

samguyer:
OK, I think I finally got it. The definitions of hival() and loval() now do the proper read of the register first, then add in the new bits, so that set() will work properly. I also fixed a few other small things. See what you think now...

focalintent:
@samguyer, I can't selectively merge only some commits from this PR with Github's UI - is it ok to take in the experimental change as well?  As for why that change fixes things - it is likely because the pixel data structure is large enough that the constant allocating/deleting of it was fragmenting the hell out of the esp32's memory until eventually you got to the point where there wasn't any memory left.  Dynamic allocation is the devil on these MCUs - I tend to prefer not using it at all, and if I have to, finding ways to use it once.

samguyer:
@focalintent I think that's exactly it. Maybe you can give me some guidance
here. I need to save the pixel data so that it's available inside the
interrupt handler. My experimental version converts the whole pixel array
to raw bits, which are stored in a temp array. The advantages of this
approach are (1) one-time allocation of the buffer, (2) some work done up
front, (3) it could allow double buffering. The disadvantage is memory
usage.

What do you think? Any other ideas/strategies?

samguyer:
That's certainly an option. I'd have to do something different for parallel
output.

The ESP32 has a fair amount of RAM -- 500Kb I think. Even 10,000 pixels
would only require 30K of extra memory.

It's really your call...

samguyer:
@focalintent I just pushed a new version that saves the pixel data in a separate buffer. I tested it myself, and there is one other user who can vouch for it.

PR #650, opened 2018-09-07
add support for Hallowing! Its a samd21 but pins are rearranged quite…

ladyada:
Its a samd21 but pins are rearranged quite a bit


focalintent:
Thank you!  I’ll cycle the project versions tonight or over the weekend

ladyada:
w0w thanks!

focalintent:
Pushed/tagged FastLED 3.2.1 which should include this support as well!

PR #657, opened 2018-09-27
Update global brightness math to fix overflows

### LINKED ISSUE #656: APA102/SK9822 global brightness sometimes incorrect
I am very happy that FastLED is now making use of the global brightness aspect of the APA102/SK9822 LEDs.  Even with the brightness turned down very low (and using much less power), these LEDs are capable of providing a wide dynamic range of colors with this feature.

However, while trying out the global brightness setting for an SK9822 LED strip, I noticed some odd brightness behavior that I was able to track down the cause of.  Specifically, in the following _[chipsets.h](https://github.com/FastLED/FastLED/blob/38105c61de097066946f2df4d2bb2c36ddc78d7c/chipsets.h#L193-L201)_ code:
```
#if FASTLED_USE_GLOBAL_BRIGHTNESS == 1
                const uint16_t maxBrightness = 0x1F;
                uint16_t brightness = (max(max(s0, s1), s2) * maxBrightness >> 8) + 1;
                s0 = ((uint16_t)s0 + 1) * maxBrightness / brightness - 1;
                s1 = ((uint16_t)s1 + 1) * maxBrightness / brightness - 1;
                s2 = ((uint16_t)s2 + 1) * maxBrightness / brightness - 1;
#else
                const uint8_t brightness = 0x1F;
#endif
```
the right side of the `s0` calculation is occasionally greater than 255, in which case `uint8_t s0` wraps around (262 → 6), resulting in a grossly incorrect brightness.  This happens for several values of `s`.

Even if the overflow is caught and `s` is set to 255, the algorithm does not preserve scale\*brightness, leading to significant deviations in the resulting LED brightness at lower scale factors.  I suggest using the following, in which scale\*brightness deviates by at most 0.4% (and typically much less):
```
#if FASTLED_USE_GLOBAL_BRIGHTNESS == 1
                const uint16_t maxBrightness = 0x1F;
                uint16_t brightness = ((((uint16_t)max(max(s0, s1), s2) + 1) * maxBrightness - 1) >> 8) + 1;
                s0 = (maxBrightness * s0 + (brightness >> 1)) / brightness;
                s1 = (maxBrightness * s1 + (brightness >> 1)) / brightness;
                s2 = (maxBrightness * s2 + (brightness >> 1)) / brightness;
#else
                const uint8_t brightness = 0x1F;
#endif
```
The above algorithm does not experience the `uint8_t` overflow.

extesy:
Thanks to @cmsavage for a heads up. I was able to reproduce the issue and applied the suggested fix. With the fix I don't see overflows happening anymore. Fixes #656.

/cc @focalintent

### DIFF
--- a/chipsets.h
+++ b/chipsets.h
@@ -192,10 +192,10 @@ protected:
 		uint8_t s0 = pixels.getScale0(), s1 = pixels.getScale1(), s2 = pixels.getScale2();
 #if FASTLED_USE_GLOBAL_BRIGHTNESS == 1
 		const uint16_t maxBrightness = 0x1F;
-		uint16_t brightness = (max(max(s0, s1), s2) * maxBrightness >> 8) + 1;
-		s0 = ((uint16_t)s0 + 1) * maxBrightness / brightness - 1;
-		s1 = ((uint16_t)s1 + 1) * maxBrightness / brightness - 1;
-		s2 = ((uint16_t)s2 + 1) * maxBrightness / brightness - 1;
+		uint16_t brightness = ((((uint16_t)max(max(s0, s1), s2) + 1) * maxBrightness - 1) >> 8) + 1;
+		s0 = (maxBrightness * s0 + (brightness >> 1)) / brightness;
+		s1 = (maxBrightness * s1 + (brightness >> 1)) / brightness;
+		s2 = (maxBrightness * s2 + (brightness >> 1)) / brightness;
 #else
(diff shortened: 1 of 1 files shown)


extesy:
@focalintent Gentle reminder... Seems like you have accumulated a few pull requests in the merge pipeline :)

extesy:
@focalintent Gentle reminder... Seems like you have accumulated a few pull requests in the merge pipeline :)

Issue #640, opened 2018-08-17
Support for ItsyBitsy M4 (SAMD51)


vccabral:
I want to use FastLED with new ItsyBitsy. The board can only use pin 5 at 5v unless it is level shifted, not sure if that is relevant. 

This seems to be very similar to #639 

```
In file included from /var/folders/ns/936wknzn38x4vjh0l3z0sj740000gn/T/arduino_modified_sketch_375437/Blink.ino:1:0:
/Users/bracero/Documents/Arduino/libraries/FastLED/FastLED.h:17:21: note: #pragma message: FastLED version 3.001.007
 #    pragma message "FastLED version 3.001.007"
                     ^
In file included from /Users/bracero/Documents/Arduino/libraries/FastLED/led_sysdefs.h:32:0,
                 from /Users/bracero/Documents/Arduino/libraries/FastLED/FastLED.h:44,
                 from /var/folders/ns/936wknzn38x4vjh0l3z0sj740000gn/T/arduino_modified_sketch_375437/Blink.ino:1:
/Users/bracero/Documents/Arduino/libraries/FastLED/platforms/avr/led_sysdefs_avr.h:16:32: error: conflicting declaration 'typedef volatile uint8_t RoReg'
 typedef volatile       uint8_t RoReg; /**< Read only 8-bit register (volatile const unsigned int) */
                                ^
In file included from /Users/bracero/Library/Arduino15/packages/arduino/tools/CMSIS-Atmel/1.2.0/CMSIS/Device/ATMEL/samd51/include/samd51.h:40:0,
                 from /Users/bracero/Library/Arduino15/packages/arduino/tools/CMSIS-Atmel/1.2.0/CMSIS/Device/ATMEL/samd.h:141,
                 from /Users/bracero/Library/Arduino15/packages/arduino/tools/CMSIS-Atmel/1.2.0/CMSIS/Device/ATMEL/sam.h:540,
                 from /Users/bracero/Library/Arduino15/packages/adafruit/hardware/samd/1.2.1/cores/arduino/Arduino.h:48,
                 from sketch/Blink.ino.cpp:1:
/Users/bracero/Library/Arduino15/packages/arduino/tools/CMSIS-Atmel/1.2.0/CMSIS/Device/ATMEL/samd51/include/samd51g19a.h:54:33: error: 'RoReg' has a previous declaration as 'typedef volatile uint32_t RoReg'
[...]

Issue #639, opened 2018-08-15
Support for Feather M4 Express (SAMD51)


pickledgator:
Hi all, is there listed support for the newer Feather M4 Express boards from Adafruit? It seems that the M0 Express boards compile FastLED just fine, however, the SAMD51 boards throw a spew of conflicts.

```
Arduino: 1.8.5 (Mac OS X), Board: "Adafruit Feather M4 (SAMD51), Enabled"

Build options changed, rebuilding all
In file included from /Users/nicfischer/Documents/Arduino/blink/blink.ino:1:0:
/Users/nicfischer/Documents/Arduino/libraries/FastLED/FastLED.h:17:21: note: #pragma message: FastLED version 3.001.007
 #    pragma message "FastLED version 3.001.007"
                     ^
In file included from /Users/nicfischer/Documents/Arduino/libraries/FastLED/led_sysdefs.h:32:0,
                 from /Users/nicfischer/Documents/Arduino/libraries/FastLED/FastLED.h:44,
                 from /Users/nicfischer/Documents/Arduino/blink/blink.ino:1:
/Users/nicfischer/Documents/Arduino/libraries/FastLED/platforms/avr/led_sysdefs_avr.h:16:32: error: conflicting declaration 'typedef volatile uint8_t RoReg'
 typedef volatile       uint8_t RoReg; /**< Read only 8-bit register (volatile const unsigned int) */
                                ^
In file included from /Users/nicfischer/Library/Arduino15/packages/arduino/tools/CMSIS-Atmel/1.2.0/CMSIS/Device/ATMEL/samd51/include/samd51.h:44:0,
                 from /Users/nicfischer/Library/Arduino15/packages/arduino/tools/CMSIS-Atmel/1.2.0/CMSIS/Device/ATMEL/samd.h:143,
                 from /Users/nicfischer/Library/Arduino15/packages/arduino/tools/CMSIS-Atmel/1.2.0/CMSIS/Device/ATMEL/sam.h:540,
                 from /Users/nicfischer/Library/Arduino15/packages/adafruit/hardware/samd/1.2.1/cores/arduino/Arduino.h:48,
                 from sketch/blink.ino.cpp:1:
/Users/nicfischer/Library/Arduino15/packages/arduino/tools/CMSIS-Atmel/1.2.0/CMSIS/Device/ATMEL/samd51/include/samd51j19a.h:54:33: error: 'RoReg' has a previous declaration as 'typedef volatile uint32_t RoReg'
[...]

PR #662, opened 2018-10-06
Add basic support for SAMD51, used in M4 line from Adafruit

balbinus:
Support is based only on Adafruit ItsyBitsy M4 Express, since it’s the only M4 board I have. SAMD51 (“M4”) seems to be very similar to SAMD21 (“M0”/Zero), although I haven't dug deep enough to be 100 % sure. Feel free to correct me :)

Related to issue #640. Adding more boards (like the Feather M4 Express from #639) would require looking up their pin assignments, but since I don't have those boards, I can't test it out.


AzureEmpyrean:
Sorry for reviving this but I have a Feather M4 with the SAMD51. What do you need to make this work for the Feather M4? if its a certain file or something of that nature I can definitely find that for you

PR #673, opened 2018-10-26
included  time.h header file

DEVELOPER-IN-PROGRESS:


focalintent:
Why?  And is it time.h or Time.h - the difference is important on systems with case sensitive file systems (eg Linux)

DEVELOPER-IN-PROGRESS:
time.h

focalintent:
Ok - and why are you adding this include here?

focalintent:
What platforms do you need an explicit include of time.h on?

PR #678, opened 2018-11-13
Change SAMD21 fastpin to use PORT_IOBUS instead of PORT

kezorm:
SAM D21 processor has single-cycle I/O port access available at PORT_IOBUS registers.  The PORT registers go through bus matrix and thus take a few extra cycles even under ideal conditions.  See section 23.5.10 - CPU Local Bus of SAM D21 [datasheet](http://ww1.microchip.com/downloads/en/DeviceDoc/SAMD21-Family-DataSheet-DS40001882D.pdf).


focalintent:
Ah - excellent find, thank you!  I’ll merge this in shortly

PR #727, opened 2019-01-27
Enable SAMD51 (ARM Cortex M4) Support

leebyron:
Adds support for SAMD51 devices like the Adafruit M4 Express line.

I did see that support was added in 8f2e86434b2bfa4d6178622ca3745a68303f8b00 by assuming SAMD51 devices are the same as SAMD21, however in my experience this isn't the case. In fact I could not get FastLED to compile for the Adafruit Itsy-Bitsy M4 Express.

While this PR still mostly wholesale copies code from the D21 implementation (I noticed that D21 mostly wholesale copied from SAM, so hopefully this approach is ok) the critical difference is replacing `PORT_IOBUS` with `PORT` for fast pin access. `PORT_IOBUS` isn't defined by D51 devices (see another reference to this here: https://github.com/adafruit/Adafruit-GFX-Library/blob/master/Adafruit_SPITFT.cpp#L48)

With this patch I was able to compile and run the "Simple Example" from the readme on the Adafruit Itsy-Bitsy M4 and see it light up a DotStar

### DIFF
--- a/platforms.h
+++ b/platforms.h
@@ -21,8 +21,10 @@
 #include "platforms/arm/sam/fastled_arm_sam.h"
 #elif defined(STM32F10X_MD) || defined(__STM32F1__)
 #include "platforms/arm/stm32/fastled_arm_stm32.h"
-#elif defined(__SAMD21G18A__) || defined(__SAMD21J18A__) || defined(__SAMD21E17A__) || defined(__SAMD21E18A__) || defined(__SAMD51G19A__)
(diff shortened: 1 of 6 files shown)


focalintent:
Ah!  I know what happened - someone submitted a patch recently to use port iobus on the samd21’s for performance reasons.  I’ll merge this now and spin a release tomorrow.  Thank you!

leebyron:
Awesome! Thanks for the quick action.

PR #728, opened 2019-01-28
Fix ATtiny25 compilation issue (port definitions)

hugopilot:
The ATtiny25 does not have PORTA. It would give compilation errors.

Also, because the ATtiny25 has a very small storage capacity, a compilation message will be presented warning the user. (The library **will** (as of now) exceed the storage capacity on the ATtiny25)


kolyshkin:
@hugopilot can you please squash your commits into one?

focalintent:
No need - I squash on merge usually :)

PR #731, opened 2019-02-04
Remove unused variable

AymericPlanche:
Remove an used variable, warnings are shown at compilation time.


focalintent:
Taken by hand since there are other esp32 changes that don't have this merge cleanly.

PR #767, opened 2019-04-06
Fix nRF51 multi-statement macro

henrygab:
Multi-statement macros can cause **_evil, evil_** bugs that are really hard to track down.  This is because macros are presumed by most users to effectively equate to a single statement.

Help meet this expectation, by using the [time-honored
technique](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Multi-statement_Macro) of wrapping the multiple statements in a `do { ... } while(0)` loop, which makes the macro resolve into a single statement.


focalintent:
I'd want to make sure that the compilers involved are actually optimizing things like this out - the timing for clockless output is pretty tight and I've had compilers get weird on me when mixing in asm code where I'm using a bunch of registers, to the point where it completely blows up the asm statements (inserting random register window spills/fills and such).

That said, these macros are going away - I'm going to be completely redoing how the asm is done, and hopefully moving all of these checks into asm as well for the platforms that need it.

henrygab:
OK.  If the compiler validation is too much trouble, I won't complain if you close this one.

Issue #744, opened 2019-02-23
FastLED does not work on STM32/BluePill  : BOUNTY US$40 for making it work.


AndKe:
Arduino 1.8.8 , FastLED current master, the LED strip is working fine with this library on a Atmega328:

On a BluePill , STM32F103C8 , with STM32duino bootloader, I tried the example "ColorPalette" 
changed only "#define LED_PIN"

to 17  as "PA7" is not recognized
then tried PB5 , (recognized).
no LED activity.

focalintent:
FYI, the stm32 support is very loose - I don’t actually have any stm32 hardware to test with, outside of the photon, and so mostly have been relying on what other folks supply for the code - especially for things like pin/port mapping.

AndKe:
they cost next to nothing, https://www.ebay.com/itm/STM32F103C8T6-Minimum-System-Developmen-Board-Module-Core-Learning-For-Arduino/323633385054?hash=item4b5a0d8e5e:g:ilcAAOSw9m5cAggN:rk:10:pf:0
If you could look into it, you could order, and I'll paypal yoy some $ ?

AndKe:
I have tried both to address port as PB5  or  "41"  - no success.
I've also checked with oscilloscope, and there is no save output on any pins, 
This is a map of what works , and what not:
(existing port/pins)  OK(pins recognized)  "not"= pins not recognized
//PA 0...7,13,14    OK=2,3,4,13,14  (not 0,1,5...7)
//PB 0..15          OK=0...6  14,15 (not 7..12)
//PC 13,14,15      OK=13,14,15

Example: "PA2" works, "PA6" does not.
This tells me that it is likely, that the pin defines for "Blue Pill" does not apply as expected here: https://github.com/FastLED/FastLED/blob/master/platforms/arm/stm32/fastpin_arm_stm32.h#L107

stujamac:
So I have had the same issue, it seems to be a simple issue in that the pins are defined wrongly in "fastpin_arm_stm32.h".
_DEFPIN_ARM is being called with the physical pin number as option 1, whereas it should be the required Arduino pin that it will be aliased as.
Am testing on a Maple Mini and all seem to work now I have got the defines straightened out.  As mentioned above

Abhyudaya98:
> Arduino 1.8.8 , FastLED current master, the LED strip is working fine with this library on a Atmega328:
> 
[...]

PR #769, opened 2019-04-09
Fix for wrong pin defines on STM32F103 Maple Mini/Blue Pill etc.

stujamac:
Pins have been wrongly defined in "fastpin_arm_stm32.h", _DEFPIN_ARM is being called with the physical pin number as parameter 1, whereas it should be the required Arduino pin that it will be aliased as.

Schematic used is https://github.com/leaflabs/maplemini/blob/master/maplemini.pdf, it mentions V3, but there doesn't appear to be any poin out changes with any other versions I have seen.

My testing has purely been on a Maple Mini, not sure if the Blue Pills are any different pinout?  If they are we need another build variable to determine between.

Have changed MAX_PIN to 34, as there are 0-33 pins (34 total), not sure where this is used, but assume this is the correct value for this?

This is against Issue #744

### DIFF
--- a/platforms/arm/stm32/fastpin_arm_stm32.h
+++ b/platforms/arm/stm32/fastpin_arm_stm32.h
@@ -108,41 +108,48 @@ _DEFPIN_ARM(19, 2, A);
 
 _IO32(A); _IO32(B); _IO32(C);
 
-#define MAX_PIN 46
+#define MAX_PIN 34 // Max number of mapped pins.
 
-_DEFPIN_ARM(10, 0, A);	// PA0 - PA7
-_DEFPIN_ARM(11, 1, A);
-_DEFPIN_ARM(12, 2, A);
-_DEFPIN_ARM(13, 3, A);
-_DEFPIN_ARM(14, 4, A);
-_DEFPIN_ARM(15, 5, A);
-_DEFPIN_ARM(16, 6, A);
-_DEFPIN_ARM(17, 7, A);
-_DEFPIN_ARM(29, 8, A);	// PA8 - PA15
-_DEFPIN_ARM(30, 9, A);
-_DEFPIN_ARM(31, 10, A);
-_DEFPIN_ARM(32, 11, A);
-_DEFPIN_ARM(33, 12, A);
-_DEFPIN_ARM(34, 13, A);
-_DEFPIN_ARM(37, 14, A);
-_DEFPIN_ARM(38, 15, A);
+_DEFPIN_ARM(11, 0, A);	// PA0 - PA7
+_DEFPIN_ARM(10, 1, A);
+_DEFPIN_ARM(9, 2, A);
+_DEFPIN_ARM(8, 3, A);
+_DEFPIN_ARM(7, 4, A);
+_DEFPIN_ARM(6, 5, A);
+_DEFPIN_ARM(5, 6, A);
+_DEFPIN_ARM(4, 7, A);   // Note some schematics have this pin as A17 - typo should be A7
 
-_DEFPIN_ARM(18, 0, B);	// PB0 - PB11
-_DEFPIN_ARM(19, 1, B);
-_DEFPIN_ARM(20, 2, B);
-_DEFPIN_ARM(39, 3, B);
-_DEFPIN_ARM(40, 4, B);
-_DEFPIN_ARM(41, 5, B);
-_DEFPIN_ARM(42, 6, B);
-_DEFPIN_ARM(43, 7, B);
-_DEFPIN_ARM(45, 8, B);
-_DEFPIN_ARM(46, 9, B);
-_DEFPIN_ARM(21, 10, B);
-_DEFPIN_ARM(22, 11, B);
+_DEFPIN_ARM(27, 8, A);	// PA8 - PA15
+_DEFPIN_ARM(26, 9, A);
+_DEFPIN_ARM(25, 10, A);
+_DEFPIN_ARM(24, 11, A);
+_DEFPIN_ARM(23, 12, A);
+_DEFPIN_ARM(22, 13, A);
+_DEFPIN_ARM(21, 14, A);
+_DEFPIN_ARM(20, 15, A);
 
-_DEFPIN_ARM(2, 13, C);	// PC13 - PC15
-_DEFPIN_ARM(3, 14, C);
-_DEFPIN_ARM(4, 15, C);
+_DEFPIN_ARM(3, 0, B);	// PB0 - PB11
+_DEFPIN_ARM(33, 1, B);  // LED on Maple Mini
+_DEFPIN_ARM(2, 2, B);
+_DEFPIN_ARM(19, 3, B);
+_DEFPIN_ARM(18, 4, B);
+_DEFPIN_ARM(17, 5, B);
+_DEFPIN_ARM(16, 6, B);
+_DEFPIN_ARM(15, 7, B);
+_DEFPIN_ARM(32, 8, B);  // Button on Maple Mini
+//_DEFPIN_ARM(34, 9, B);  // PB9 is not connected on Maple Mini
+_DEFPIN_ARM(1, 10, B);
+_DEFPIN_ARM(0, 11, B);
+_DEFPIN_ARM(31, 12, B);
+_DEFPIN_ARM(30, 13, B);
+_DEFPIN_ARM(29, 14, B);
+_DEFPIN_ARM(28, 15, B);
+
+
+_DEFPIN_ARM(14, 13, C);	// PC13 - PC15
+_DEFPIN_ARM(13, 14, C);
+_DEFPIN_ARM(12, 15, C);
+
 
 #define SPI_DATA BOARD_SPI1_MOSI_PIN
 #define SPI_CLOCK BOARD_SPI1_SCK_PIN


stujamac:
I think the cause of confusion is this diagram: http://reblag.dk/wordpress/wp-content/uploads/2016/07/The-Generic-STM32F103-Pinout-Diagram.pdf
The numbers in blue boxes directly next to the board are the MCU pin numbers.  These do not correspond to the aliased pins.  I am not sure what the pins should be called on the Blue Pill, looking at the above diagram there are no pin 'logical' pin numbers which map to a simple 0-33...

focalintent:
The pin definitions that were in there were submitted for the blue pill I believe - so if there's going to be another set of pin/port mappings - if you're going to change up the pin mappings, you shouldn't remove the old mapping, but you should add the define checks to chose between the existing mappings and the one that you're adding (given that a bunch of folks are using the stm32 code with the existing mappings - if I took your pull request as it is, it would suddenly break all their projects, which is something that I'd prefer to avoid doing).

stujamac:
I have to agree with you not wanting to break things, and my mappings are different.

However, I don't see how the Blue Pill defines can work.  My rationale being that the Blue Pill and the Maple Mini are the same processor, same pin out, peripherals etc.  The layout of the Blue Pill has Port and Pin references on it, which cannot be changed on the MCU.  These are the same as the 
Maple Mini (or any STM32F103 clone), so the pin references that are set up must be the same.

The defines that are being created in "fastpin_arm_stm32.h" are basing off against the physical pin layout, which are essentially arbitary numbers with FastLED as they do not match up with the internal logical layout of the pins. 

To verify this on the Maple Mini, I can load the master FastLED library and if I set for example LED_PIN as 14, then according to the Generic pin out diagram for the Blue Pill, this should give output on Pin 14, PA4.  As PA4 is the same on both Maple Mini and Blue Pill, then this should equate to output 7 on my Maple Mini (which is physical pin 14).  However, nothing is output and I get exactly the same issue as #744.

So unfortunately this means that the definitions for the Blue Pill that are currently in the master branch are incorrect.  This is borne out by the issues seen in #744 and I would be surprised if any one has a working project with the Blue Pill with these mappings.

stujamac:
So having a bit more of a thought about it and I think it runs deeper than just the defines...

The Blue Pills pin outs are mapped directly to the physical pin names in the current mappings, which is fine as they are just names.  However, somewhere in the code, the pin numbers are being used in the Arduino numbering system i.e. 0 - 31 (or whatever it goes up to).   The mappings I have put in work to make this correct with the Maple Mini's Arduino compatible pin out.  The Blue Pill does not have this compatible pin out naming, hence it cannot access the correct peripherals etc.

Therefore there must be something at a lower layer (in FastLED?) that is using the Arduino compatible pins to reference the mappings, which then don't translate to anything...

I will have more of a dig, and yes we will need to create a new mapping for Maple etc. but we also need to fix this lower layer bug where it is referencing the wrong pins?  Will also get a Blue Pill on order...

stujamac:
With regards to Red Pill which is meant to be pin-compatible with Maple Mini and Blue Pill, its schematic is here: http://wiki.stm32duino.com/index.php?title=File:Red_Pill_pinout.png and has the same layout as the Maple Mini, i.e. IO pins are named in the Arduino convention.  

So first issue appears to be Blue Pill numbering doesn't align with its 'pin compatible' brethren.

Second issue appears to be that any different pin-numbering than the Arduino standard is probably being broken by the pin mapper (.../arduino/hardware/Arduino_STM32-master/STM32F1/variants/generic_stm32f103c/board.cpp) in the STMDuino project, as it maps the pins to be Arduino specific internally, then FastLED maps back using its own pin definition, which in the Blue Pills case is not Arduino, so doesn't match up, so doesn't work.  

So maybe if you are not using Arduino and use FastLED, it is OK with the pin numbering?  Then once STMDuino is also in use we have a mis-match of pin redirection?

focalintent:
I think you are conflating a handful of things here - so I'm going to take a step back for a moment./

At a low level - as in the asm/hardware level, to access a GPIO pin, you need two pieces of information - it's PORT (as in, the collection of GPIO registers that this pin is a part of for setting, reading, clearing, toggling, etc...) and the bit on that port that associates to the pin.  On the AVR platform, which is an 8-bit platform, there are 8 pins per port and so one pin might be on port A, bit 0, the next pin on port A, bit 1, etc... etc..   On most 32-bit arm platforms, a port has 32 pins on it, and for reasons only known to the gods and some engineers who I suspect were drunk or high when they did it, on the STM32 a port has 16 pins attached to it.

The pin mapping that FastLED does is for a number to a port/bit pairing.  Then, when you say ```FastLED.addLeds<WS2812, 13>(...);``` - that number 13 is used to instantiate FastPin<13> which in turn is a class with compile time hard coded references to the appropriate port and offset.  

I do this because I have need to know the actual port and bit of a pin at _compile time_ - not runtime.  

On the Arduino side, there's a table maintained at runtime that does this mapping - so when you say digitalWrite(13, HIGH); the first thing it does is look up in one table for a reference to the GPIO register for writing pin 13 and then a lookup in another table for what bit on that GPIO register corresponds to pin 13.  FastLED completely bypasses this entirely (because, again, I need this information to be static at compile time).

There's no intersection between the two.  

This is also why FastLED needs to have build target specific pin mappings set up - because just because two stm32 Arduino boards have the same PIN numbers in the same location on their boards, those PIN numbers may reference different port and bits on the the physical underlying chip - and what's important to FastLED at a low level is the port/bit index, _not_ the number that is written on the board.  So - what needs to be built is a map of the PIN number on the board to the actual port/bit for that specific platform. 

The one possible exception to this is if people use the "D12" style of naming pins that some Arduino platforms use - and there's a reason why no FastLED example uses that, and why whenever I come across folks using it in questions I tell them not to do it.  Because D12 does not mean "12" - it might be a #define or a const pointing to a different number altogether, that gets replaced at compile time.  For example if a platform says:

const int D12 = 17;

Then if you say ```FastLED.addLeds<WS2812, D12>(...);``` it would be identical to having written ```FastLED.addLeds<WS2812, 17>(...);``` - which is far less likely to map up to the pin/port mapping you might think you're mapping to.


And in fact, if you look at:

https://wiki.stm32duino.com/index.php?title=Blue_Pill

you will see the port/bit number mapping that corresponds to what's in FastLED, e.g. pin 40 is Port B, bit 4, while pin 17 is Port A bit 7.  

Now, what's interesting is that the red pill pinout that you linked to above at http://wiki.stm32duino.com/index.php?title=File:Red_Pill_pinout.png, the physical positions on the boards (once you account for the fact that they are rotated 180 degrees from each other) represent the same Port and bit (e.g. PA7, PB4, etc...) - however the _pin number_ that's presented on that diagram is wildly different.

So, this is one of those weird cases where, if you are thinking about things at a low level gpio register level, the blue pill and the red pill have the same port/bit layout as each other.  However, at a higher level, as far as what's printed on the pcb/in reference documentation for the number given to each of those ports, the numbering is wildly different.

But that said, it does look like there is a very different map between PIN number as given on the blue pill than on the red pill - and so, again, there needs to be two separate mappings provided in the header file here, one for the red pill's numbering and one for the blue pill's numbering.

And it needs to be made sure that the correct mapping is selected for the specific platform that is being built for.

stujamac:
Your method of setting up the pins is good and efficient, I like it, wasn't saying that this is at fault.  I don't know the Arduino or FastLED well enough to know if there is anything else in the way, hence my string of thinking from previous post.  Your explanation has helped clarify that there isn't.

My main issue for thinking was that the Blue Pill / Red Pill / Maple Mini were all Pin compatible.  This doesn't seem to be the case and I was thinking that the Blue Pill pinout is wrong and should adhere to the same as the Maple Mini etc.  There appears to be some confusion on the internet with regards to this too.  Anyhow, I am wrong on this part, the pinout in the diagram is what is out there so is correct, we are not out here to fix the issue of other boards etc...

Soooo... I propose two steps:
1) I do as you said originally ;-) I will define out the changes i made to Maple Mini (and maybe Red Pill).
2) Will review what is up the with the Blue Pill, I have one on order so will have a look over the weekend.

AndKe:
@stujamac great ! I tested your changes, but did not got it working, (BluePill) it compiles just fine using names like PB6  - but it simply does not work.  (may be my "dirty" library)  - did you try, know a particular pin that should be "safe" to try ? (one you are pretty confident is mapped correctly ?)

stujamac:
@AndKe thanks for the trying and the feedback!
You should be able to get output on a pin using an integer pin name.  If you have my changes, you *should* be able to set (for example) led pin as pin 3, and hopefully see output on the Blue Pill pin PB0 ( pin 18 on the pin out diagram).
It would be interesting to know if this works as expected on the Blue Pill, not that is a fix for it as such, but is a step in the right direction...

AndKe:
Using your fastpin file: https://github.com/FastLED/FastLED/blob/4fd33bbd0f325425d6faa59ca75425b649f6b04b/platforms/arm/stm32/fastpin_arm_stm32.h + FastLED master.
I compiled the "ColorPalette" , assigned LED_PIN 3 , connected to PB0 , and nothing. (checked with scope, no output)

Another strange thing is that PA 11 and PA12 emit some strange signals every 1ms, and PA13 is always high.

stujamac:
Strange, that still seems similar to what I got with the Maple Mini before sorting out its pin mapping...
I also noted the strange outputs, PA11 and PA12 are specified as USB pins and I think that when they don't get initialised, they must revert to being USB inputs, hence the noise.  As for PA13, no-idea what it is trying to be in its un-initialised state.

Thanks for trying that though, it proves that there is something else we are missing.

My Blue Pill gets here tomorrow, I will have a play over the weekend and hopefully have a better idea.  Watch this space...

AndKe:
@stujamac - Did you get your BluePill yet / had chance to try it?

stujamac:
@AndKe Sorry slow reply, have been struggled to get time to look at this.

However, yes I have got my Blue Pill and there is some strange things going on...
I too cannot get my Blue Pill working, with my changes above or any number of changes.  Using my change above there does appear to be some updates on the pin output, but it is almost more like noise of the signal and not the signal itself.
No matter what I change the pin map to it does not seem to work correctly.  I have checked physical connection of chip to output pins and also that the oscillator is running correctly and this seems to be OK.

I do not quite understand it, after accounting for any different pin output it does not work.  The MCU is the same (maybe not exactly as one looks like the CB version and the other C8, but this should only be a difference in flash/ram amounts).  So if the mapping of ports on the Maple Mini works, it should work on the Blue Pill, the mechanism is the same... 

I will try and have another proper look at it one evening and hook it up to another dev environment to check the hardware out fully.

AndKe:
It sounds like some register may be left in a different state, so some io or timing is different than what is the case with Maple Mini.  - or even fuses?
I am using the boot20_pc13 bootloader: https://github.com/rogerclarkmelbourne/Arduino_STM32/wiki/Flashing-Bootloader-for-BluePill-Boards
Keep up the good work.

DX12Diodes:
I am also running into pin out issues with the blue pill but not sure its just FASTLED that's having issues here. 
What ive found so far... 
Im using a 
https://www.amazon.co.uk/gp/product/B06XHFVHPW/ref=ppx_yo_dt_b_asin_title_o00_s00?ie=UTF8&psc=1 build quality seems good.

Depending on what STM32 Library you use will produce different results. Ive been testing..
STM32duino from http://dan.drown.org/stm32duino/package_STM32duino_index.json and 
STM32(STMicro Release) from https://github.com/stm32duino/BoardManagerFiles/raw/master/STM32/package_stm_index.json

Forgetting FASTLED for a second just trying to send a simple blink to an led on a pin out I get different results
On STM32duino using PB5 I can get an LED to blink find. If I set the exact same code on PB4 it doesn't work and just outputs a dimly lit "always on" LED.
On STM32(StMicro) blink works fine on PB4 and PB5

Now when looking at FASTLED
Im just using a simple code to turn all the LEDs white.
`void loop() { 
  // Turn the LED on,
  fill_solid( &(leds[0]), NUM_LEDS, CRGB::White);
  FastLED.show();
}`
**On STM32duino** I can get the code to compile (if i switch to 128K, which my blue pill isnt) but its massive... Sketch uses 75992 bytes (57%) of program storage space. Maximum is 131072 bytes.
I can even write it without errors to my pill but nothing outputs, probably because of the size issues.
Ive seen a few posts on removing exceptions to reduce the size but cant see specifically how to do this @evq and others have done some work on this.

**On STM32(StMicro)** i just get an error about a file its trying to reference that it cant find.
'libraries\FastLED-master/platforms/avr/led_sysdefs_avr.h:12:20: fatal error: avr/io.h: No such file or directory

 #include <avr/io.h>
                    ^
compilation terminated.

exit status 1
Error compiling for board Generic STM32F1 series.'
Again theres some reference to this in the fastLED library but as far as I can tell all of this has now been included so there shouldn't be an issue with it any more.

Following with interst…...

AndKe:
@DX12Diodes .. to fix the ridiculous flash consumption, see https://github.com/rogerclarkmelbourne/Arduino_STM32/issues/296

jihao:
I've got a few "Black pill", according to the [Black_Pill_Schematic.pdf](https://wiki.stm32duino.com/images/5/52/Black_Pill_Schematic.pdf) it should be the same pin out like defined in the "fastpin_arm_stm32.h". Should be the same as "Blue pill". 

I've tested the FastLED [Blink](https://github.com/FastLED/FastLED/tree/master/examples/Blink) example, tried several pins, can't get it to work with my "black pill" board. 

Code like this, 42 = PB6 which has PWM.

```
#define DATA_PIN 42  
FastLED.addLeds<WS2812B, DATA_PIN, RGB>(leds, NUM_LEDS);
```

I also tried @stujamac fix (4fd33bbd0f325425d6faa59ca75425b649f6b04b), no luck there. 

Thanks your investigation. 
So I am also following, if any patch, would like to do the test.

Fxlr8:
@jihao, same thing for me, have you solved this problem?

jihao:
@Fxlr8 I gave up, switch back to esp8266 board which works fine.

berylleigh:
I have it working.  

I have used fastpin_arm_stm32.h from here: https://github.com/13rac1/FastLED-STM32.  I simply replaced the existing file in  FastLED\platforms\arm\stm32 with the one from the stm32 fork git.     The pin mapping that works looks like this:

#if defined(__STM32F1__)
#define _RD32(T) struct __gen_struct_ ## T { static __attribute__((always_inline)) inline gpio_reg_map* r() { return T->regs; } };
#define _IO32(L) _RD32(GPIO ## L)

_IO32(A); _IO32(B); _IO32(C); _IO32(D);

#define MAX_PIN PB1

_DEFPIN_ARM(PB11, 11, B);
_DEFPIN_ARM(PB10, 10, B);
_DEFPIN_ARM(PB2, 2, B);
_DEFPIN_ARM(PB0, 0, B);
_DEFPIN_ARM(PA7, 7, A);
_DEFPIN_ARM(PA6, 6, A);
_DEFPIN_ARM(PA5, 5, A);
_DEFPIN_ARM(PA4, 4, A);
_DEFPIN_ARM(PA3, 3, A);
_DEFPIN_ARM(PA2, 2, A);
_DEFPIN_ARM(PA1, 1, A);
_DEFPIN_ARM(PA0, 0, A);
_DEFPIN_ARM(PC15, 15, C);
_DEFPIN_ARM(PC14, 14, C);
_DEFPIN_ARM(PC13, 13, C);
_DEFPIN_ARM(PB7, 7, B);
_DEFPIN_ARM(PB6, 6, B);
_DEFPIN_ARM(PB5, 5, B);
_DEFPIN_ARM(PB4, 4, B);
_DEFPIN_ARM(PB3, 3, B);
_DEFPIN_ARM(PA15, 15, A);
_DEFPIN_ARM(PA14, 14, A);
_DEFPIN_ARM(PA13, 13, A);
_DEFPIN_ARM(PA12, 12, A);
_DEFPIN_ARM(PA11, 11, A);
_DEFPIN_ARM(PA10, 10, A);
_DEFPIN_ARM(PA9, 9, A);
_DEFPIN_ARM(PA8, 8, A);
_DEFPIN_ARM(PB15, 15, B);
_DEFPIN_ARM(PB14, 14, B);
_DEFPIN_ARM(PB13, 13, B);
_DEFPIN_ARM(PB12, 12, B);
_DEFPIN_ARM(PB8, 8, B);
_DEFPIN_ARM(PB1, 1, B);

#define SPI_DATA BOARD_SPI1_MOSI_PIN
#define SPI_CLOCK BOARD_SPI1_SCK_PIN 

#define HAS_HARDWARE_PIN_SUPPORT

#endif

AndKe:
@berylleigh Fix confirmed, works fine on BluePill
are you about to make a PR out of it?  - would love to see it in master/stable

zackees:
Maple mini and bluepill boards are now compiling.

I see what Dan was saying about not wrecking legacy boards. I also see that the situation was so bad that there was a fork that was made in order to support these boards.

What I've done is ingested the forked stm32 implementation from https://github.com/13rac1/FastLED-STM32 and made sure it works with the newer compiler chains without additional warnings. I do test for a define for blue pill or maple mini and if they match then they use the new pin definition file.

Since both boards now compile I'm marking this as closed. Feel free to open this issue if the problem remains.

PR #789, opened 2019-05-14
Updated ESP32 support

samguyer:
This pull request contains improvements to the default RMT-based implementation as well as the new I2S implementation that supports up to 24-way parallel output.


platforms/esp/32/clockless_i2s_esp32.h
```diff
@@ -0,0 +1,767 @@
...
+    {
+        // Precompute the bit patterns based on the I2S sample rate
+        Serial.println("Setting up fastled using I2S");
+
+        // -- First, convert back to ns from CPU clocks
+        uint32_t T1ns = ESPCLKS_TO_NS(T1);
+        uint32_t T2ns = ESPCLKS_TO_NS(T2);
+        uint32_t T3ns = ESPCLKS_TO_NS(T3);
+        
+        Serial.print("T1 = "); Serial.print(T1); Serial.print(" ns "); Serial.println(T1ns);
```
focalintent: Serial.print's should either be commented out or gated by #ifdef or a variable (I try to not have the library generally writing to the Serial port - for cases where folks are using the Serial port as, say, a control channel).

samguyer:
OK, I commented out all the serial output.

PR #794, opened 2019-05-18
Update fastpin_arm_d21 add ItsyBitsy M0

starkatt:
Add support for Adafruit ItsyBitsy M0 including onboard APA102 ("Dotstar") LED


focalintent:
Thank you!

PR #802, opened 2019-05-22
Enable support for nRF52 chipset.

henrygab:
This has FastPIN, FastSPI, and clockless.

LED strings for clockless are temporarily limited to 144 LEDs,
adjustable via led_sysdefs.h #define to any length needed,
limited only by RAM..

### DIFF
--- a/chipsets.h
+++ b/chipsets.h
@@ -428,6 +428,16 @@ protected:
 //
 // Clockless template instantiations - see clockless.h for how the timing values are used
 //
+// Base template for clockless controllers.  These controllers have 3 control points in their cycle for each bit.
+// At T=0        : the line is raised hi to start a bit
+// At T=T1       : the line is dropped low to transmit a zero bit
+// At T=T1+T2    : the line is dropped low to transmit a one bit
+// At T=T1+T2+T3 : the cycle is concluded (next bit can be sent)
+//
+// The units used for T1, T2, and T3 is nanoseconds.
+// For 8MHz/16MHz/24MHz frequencies, these values are also guaranteed
+// to be integral multiples of an 8MHz clock (125ns increments).
+//
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
 #ifdef FASTLED_HAS_CLOCKLESS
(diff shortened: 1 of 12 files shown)


chipsets.h
```diff
@@ -493,7 +509,7 @@ template <uint8_t DATA_PIN, EOrder RGB_ORDER = RGB>
 class TM1803Controller400Khz : public ClocklessController<DATA_PIN, 6 * FMUL, 9 * FMUL, 6 * FMUL, RGB_ORDER> {};
 
 template <uint8_t DATA_PIN, EOrder RGB_ORDER = RGB>
-class TM1829Controller800Khz : public ClocklessController<DATA_PIN, 2 * FMUL, 5 * FMUL, 3 * FMUL, RGB_ORDER> {};
+class TM1829Controller800Khz : public ClocklessController<DATA_PIN, 2 * FMUL, 5 * FMUL, 3 * FMUL, RGB_ORDER, 0, true, 500> {};
```
henrygab: TM1829 is an inverted-signal chip, as seen further below....

chipsets.h
```diff
@@ -505,82 +521,87 @@ template <uint8_t DATA_PIN, EOrder RGB_ORDER = RGB>
 class PL9823Controller : public ClocklessController<DATA_PIN, 3 * FMUL, 8 * FMUL, 3 * FMUL, RGB_ORDER> {};
 
 #else
+    
+// Similar to NS() macro, this calculates the number of cycles for
+// the clockless chipset (which may differ from CPU cycles)
+#define C_NS(_NS) (((_NS * ((CLOCKLESS_FREQUENCY / 1000000L) + 999) / 1000)
```
focalintent: Next time - please make sure your stuff compiles - and compiles for other platforms before you submit a PR -- you broke compilation for pretty much every platform because of unbalanced parens here.

henrygab: You're right.  This was a mistake on my part, in that I didn't have a setup to compile multiple platforms.  I hope it's obvious that it wasn't my intent to break the build.

Wondering if there's travis support for Arduino...  Yes!  (See https://github.com/adafruit/travis-ci-arduino and https://learn.adafruit.com/the-well-automated-arduino-library/repository).

Would there be interest in Travis CI builds for FastLED?  Not a cure-all, but a step in the right direction....

henrygab: I'll just see if I can get it going, and if so, I'll offer it up.  Should be easier than adding support for a new chip.

PR #846, opened 2019-07-20
Enable TravisCI builds

henrygab:
Logging is designed to make it quick to find failures, and focus on only the relevant portions of the build log:
- Color-coding is used to help visually review build log status.
- Summary of pass/fail is output at the end of each log.
- Each platform is "folded" in Travis-CI output
- Each build failure is "folded" in Travis-CI output

To enable the simplest builds, examples were modified to use pins 2, 3, and 4, which are defined as available for use on all the supported platforms.

All samples are built by default.  This is easily configurable both per-platform and per-sample by creating files in the sample's directory:
1. Skip for all platforms by creating file '.test.skip'
2. Skip specific platform by creating file '.test.skip.[platform]'
3. Enable only for specific platform(s) by creating one or more files '.test.only.[platform]'

Initial platforms:
- stm32
- m4
- due
- zero
- trinket
- gemma
- esp8266
- esp32
- nrf52840

This builds successfully.  See example output:
See: https://travis-ci.org/henrygab/FastLED/builds/561486345

For a sense of failed builds:
https://travis-ci.org/henrygab/FastLED/builds/561469510

_NOTE: I will be slow to respond until early august._


henrygab:
Please strongly consider a **_squash-based merge_**.  I rebase'd to pickup latest commits, and don't know of a simple way to clean the history otherwise.....

henrygab:
_Note: When pulled, also revisit PR #218 (which was opened October 2015)._

focalintent:
Yeah - at that point I was holding off on that until I set up a platformio environment (which is still technically not an officially supported/tested build environment :) - and then Chris died and I stopped working on the library for a while.  On the FastLED 4 work I’m going to likely pull in to platform io soon so that when released it’ll use that as well.  (Outside of the teensy 4 port - most of my “I have time to work on FastLED” is focused on the rewrite)

henrygab:
I'm sorry about Chris … that must have been hard then, and still hard.  I'm excited to see the eventual rewrite (no pressure, just interest).  PlatformIO, if it can work, would be great.  I tried that first, but had trouble, so shifted to TravisCI because I could look at a working example.

"FastLED 4" .... is that the official name?  😁

henrygab:
Rebased against current master.

While the eventual enabling of PlatformIO is an awesome goal, this provides immediate value, today, by providing a sanity-check for all commits (both in the official repo, and for users who enable TravisCI on their forks).

henrygab:
Updating title to reflect WIP -- some new boards are failing builds, but the errors suggest that the problems are pre-existing, and thus not caused by this PR.  Still, want to generate a clean PR, so marking a work-in-progress.

PR #857, opened 2019-08-06
Fix nRF52 race condition / hard lock

### LINKED ISSUE #840: SparkFun Pro nRF52840 Mini + Neopixels locking up when calling FastLED.show() too quickly
I have a new SparkFun Pro nRF52840 MIni and I'm trying to get it up and running with Neopixels + FastLED.  I've got it most of the way there but I'm running into a small issue with things locking up when I call FastLED.show() too quickly if I have more than 32 LEDs set up.  

To start, I'm using the Adafruit nRF52 board definition (v0.10.1 because FastLED won't compile against the newest, v0.11.1, probably a separate issue).   I'm using the board definition and variant available here:  https://github.com/sparkfun/nRF52840_Breakout_MDBT50Q/tree/master/Firmware/Arduino

I made one change to the board definition, they define it as an NRF52840_FEATHER but I don't think that's correct because the pinouts are different, so I changed the board definition to NRF52840_PCA10056 as that matches the pinouts on this board.  I am running the data line off a pin which support high speed GPIO.

Everything seems to work ok if I'm trying to drive up to about 32 pixels.  But when I get beyond that, the board locks up relatively quickly if I call FastLED.show() without a rate limit.   When it locks up, the LEDs stop refreshing and it stops writing to the Serial console (although Bluetooth DOES continue to work, I can connect and send commands).  If I wrap FastLED.show() in an EVERY_N_MILLISECONDS(3) call then it seems to run fine forever. 

I've tried adjusting the delay how quickly show() is called.  It doesn't seem to be related to the number of pixels - if it locks up it will lock up at 1 or 2ms delay and it works with a 3ms delay, no matter how many pixels there are.

I am able to reproduce the issue without having any Bluetooth stuff turned on.  The below code runs fine forever at 32 pixels.  At 33 pixels it locked up after about 10 minutes.  At 34 pixels it locks up after about 10 seconds.  At 64 pixels it locks up immediately.  Wrapping the FastLED.show() call in EVERY_N_MILLISECONDS(3) will let it run fine forever with 64 pixels.

I'm powering the LEDs of the 3.3V of the board but it seems unlikely to be a power issue.  Turning the brightness from 10 to 100 does not cause any issues with 64 pixels if there is a 3ms delay around the show() call.

Edit: it did eventually lock up on 32 pixels

```
#include <FastLED.h>

#define NUM_LEDS 48
#define LED_PIN 5


CRGB leds[NUM_LEDS];

void setup() {
  delay(2000);

  pinMode(LED_PIN, OUTPUT);
  
  FastLED.addLeds<NEOPIXEL, LED_PIN>(leds, NUM_LEDS);
  FastLED.setBrightness(10);
  fill_solid(leds, NUM_LEDS, CRGB::Blue);
  FastLED.show();
}

void loop() {
  Serial.println("Showing");
  static int i = 0;
  if(i) {
    fill_solid(leds, NUM_LEDS, CRGB::Blue);
    i = 0;
  } else {
    fill_solid(leds, NUM_LEDS, CRGB::Red);
    i = 1;
  }
  FastLED.show();
}
```

henrygab:
Fixes #840.
<details>
    <summary>A textbook case of where `volatile` is necessary</summary>

<p>

The variable `s_SequenceBufferInUse` is modified in an ISR.
However, the variable was **_not_** marked `volatile`:
```C++
virtual void showPixels(PixelController<_RGB_ORDER> & pixels) {
    while (s_SequenceBufferInUse != 0);
```

Because it wasn't marked `volatile`, the compiler was allowed to use the first value it read.  Using [compiler explorer](https://gcc.godbolt.org/) set to `ARM gcc 7.2.1 (none)` with optimizations (e.g., `-O5`), you would get something similar to:

```asm
.L10
        .word .LANCHOR0 // address of .LC0
        .word .LC0             // s_SequenceBufferInUse
showPixels():
        ldr     r3, .L10
        ldr     r3, [r3]
        cmp     r3, #0
        beq     .L6
.L7:
        b       .L7 // Note this doesn't re-read the value.....
.L6:
        // remainder of function
```     

</details>

### DIFF
--- a/platforms/arm/nrf52/clockless_arm_nrf52.h
+++ b/platforms/arm/nrf52/clockless_arm_nrf52.h
@@ -123,6 +123,18 @@ private:
     FASTLED_NRF52_INLINE_ATTRIBUTE static void startPwmPlayback_StartTask(NRF_PWM_Type * pwm) {
         nrf_pwm_task_trigger(pwm, NRF_PWM_TASK_SEQSTART0);
     }
+    FASTLED_NRF52_INLINE_ATTRIBUTE static void spinAcquireSequenceBuffer() {
+        while (!tryAcquireSequenceBuffer());
+    }
+    FASTLED_NRF52_INLINE_ATTRIBUTE static bool tryAcquireSequenceBuffer() {
+        return __sync_bool_compare_and_swap(&s_SequenceBufferInUse, 0, 1);
+    }
+    FASTLED_NRF52_INLINE_ATTRIBUTE static void releaseSequenceBuffer() {
+        uint32_t tmp = __sync_val_compare_and_swap(&s_SequenceBufferInUse, 1, 0);
+        if (tmp != 1) {
+            // TODO: Error / Assert / log ?
+        }
+    }
 
 public:
     static void isr_handler() {
@@ -135,7 +147,7 @@ public:
             nrf_pwm_event_clear(pwm,NRF_PWM_EVENT_STOPPED);
 
             // mark the sequence as no longer in use -- pointer, comparator, exchange value
-            __sync_fetch_and_and(&s_SequenceBufferInUse, 0);
+            releaseSequenceBuffer();
             // prevent further interrupts from PWM events
             nrf_pwm_int_set(pwm, 0);
             // disable PWM interrupts - None of the PWM IRQs are shared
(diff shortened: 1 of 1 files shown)


henrygab:
FYI -- TravisCI build succeeded.
See https://travis-ci.org/henrygab/FastLED/builds/568439071

Issue #877, opened 2019-08-30
(Suggestion) Improvement for RGBCalibrate Example


jblanteigne:
I like how unambiguous this example is but I expected to find commands to change the order of the LEDs to fix my issue right away. They should be in there instead of having me search the Wiki for it. 
Thanks

zackees:
Closing this issue — it was opened over 4 years ago with no activity. If this is still relevant with a current FastLED version, please reopen with a minimal reproduction sketch and version details. Thank you!

PR #879, opened 2019-09-03
Make examples consistent and document the likely non-RGB types

### LINKED ISSUE #878: Examples are inconsistent in terms of color order for WS2812(B) and NeoPixel
The color signalling order for NeoPixels (the WS2812 variety), WS2812, WS2812B, WS2852, and SK6812 LEDs are all `GRB`, not `RGB`. However, the examples are unclear on this matter.

For instance, [Blink.ino](https://github.com/FastLED/FastLED/blob/master/examples/Blink/Blink.ino) shows many pixel types and their configurations, but of the types above only the default NeoPixel one will work right "out of the box" - the WS2812, WS2812B, WS2852, and SK6812 (where present) will not, as they specify the `RGB` color order, not their normal `GRB` order.

It would be good to have a small comment by the NeoPixel line noting its implicit default to `GRB` signalling, and the WS2812/WS2812B/WS2852/SK6812 ones (where present) should actually specify `GRB` rather than `RGB`.

I've seen notes suggesting inconsistencies within these groups, but I suspect this is down to users mixing up APA102 / SK6822 with WS2812 variants above (APA102 seems to have several color orders - but I've yet to see a properly marked WS2812 vary from GRB - if one does, it's not the norm).

Making the examples more explicit about NeoPixel / WS2812 / WS2812B / WS2852 / SK6812 being typically `GRB` will help reduce user confusion.

MartyMacGyver:
Resolves #878

### DIFF
--- a/examples/RGBCalibrate/RGBCalibrate.ino
+++ b/examples/RGBCalibrate/RGBCalibrate.ino
@@ -23,12 +23,14 @@
 //
 //////////////////////////////////////////////////
 
-#define NUM_LEDS 6
+#define NUM_LEDS 7
 
-// Data pin that led data will be written out over
-#define DATA_PIN 6
+// For led chips like WS2812, which have a data line, ground, and power, you just
+// need to define DATA_PIN.  For led chipsets that are SPI based (four wires - data, clock,
+// ground, and power), like the LPD8806 define both DATA_PIN and CLOCK_PIN
 // Clock pin only needed for SPI based chipsets when not using hardware SPI
-//#define CLOCK_PIN 8
+#define DATA_PIN 3
+#define CLOCK_PIN 13
 
 CRGB leds[NUM_LEDS];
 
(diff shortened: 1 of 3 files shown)


examples/RGBCalibrate/RGBCalibrate.ino
```diff
@@ -65,8 +88,7 @@ void loop() {
     leds[2] = CRGB(0,255,0);
     leds[3] = CRGB(0,0,255);
     leds[4] = CRGB(0,0,255);
-    leds[5] = CRGB(0,0,255);
-    leds[6] = CRGB(0,0,0);
```
MartyMacGyver: Noticed and fixes a latent defect here - the array is only 6 entries long, so this is a buffer overflow as-is.

skypanther:
Looks like this also resolves #877

examples/RGBCalibrate/RGBCalibrate.ino
marmilicious: leds[5] still needs to be set to blue (0,0,255) though.

MartyMacGyver: I thought the last led should still be black as apparently was intended in the original code, but I can easily change this.

MartyMacGyver: I've made the change

MartyMacGyver:
This covers #928 as well.

kriegsman:
Thank you for the contributions and the help!

MartyMacGyver:
Thanks for accepting my changes! Glad to help out.

PR #891, opened 2019-09-18
Fix truncated SPI_DATA_RATE template parameter to addLeds with Teensy 4

woundedlion:
The SPI_DATA_RATE template parameter to addLeds() is declared as a uint8_t. On Teensy 4, DATA_RATE_MHZ and DATA_RATE_KHZ are defined to pass the raw clock rate in Hz to the Controller class, yielding values larger than 255. The value is truncated resulting in an incorrect/unusable SPI clock rate. I fixed this by redefining as uint32_t.


kriegsman:

Issue #881, opened 2019-09-03
Division by zero in countFPS()


limpens:
This issue maybe caused by calling FastLED.show() too frequently, but running FastLED on an ESP8266, with the CPU at 160Mhz, I occasionally have a 'millis() == 8191' and lastframe also 8191.

With the logic at https://github.com/FastLED/FastLED/blob/master/FastLED.cpp#L212 this will end up in a 'now' becoming 0 and an exception calculating the m_nFPS.

Sincerely, Eric.

andjoeg:
Trying out changing from millis() to micros() to give higher resolution

PR #906, opened 2019-10-21
Added error handling to prevent divide-by-zero

andjoeg:
Added a max() function to the logic at https://github.com/FastLED/FastLED/blob/master/FastLED.cpp#L212 to ensure the time elapsed from last frame is never used as 0 in a division operation, to resolve [#881.](https://github.com/FastLED/FastLED/issues/881)

### DIFF
--- a/FastLED.cpp
+++ b/FastLED.cpp
@@ -210,7 +210,8 @@ void CFastLED::countFPS(int nFrames) {
   if(br++ >= nFrames) {
 		uint32_t now = millis();
 		now -= lastframe;
-		m_nFPS = (br * 1000) / now;
+	  	//the max function ensures there will never be a value of 0 for now, as pointed out in issue #881 - it should only cause inaccuracies past 1000FPS, which is huge
+		m_nFPS = (br * 1000) / max(now,1); 
     br = 0;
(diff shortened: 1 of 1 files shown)


kriegsman:
Hi @andjoeg-  Thank you for the PR and for the assistance!  I just wanted to let you know that I'm taking a look at this now.

kriegsman:
The only part of your PR that I have any hesitation about is the call to the "max" function.  I think we could get the same effect, without the overhead of a function call, and without having to include the "max" function in the compiled binary, if we did something like this
```
if( now == 0) {
    now = 1;
}
```
I'm not set up right now to get metrics on code size and number of instructions, but I'm going to try it both ways, and whichever is smaller and faster is the one I'll go with, and I'll let you know.

Thanks again for the assistance, much appreciated.

kriegsman:
Thanks for the help, @andjoeg.  The generated code turns out to be smaller with the if/then than using max(), so I went that way in https://github.com/FastLED/FastLED/commit/5608ecf7771274a9724d72bc848c33e0026dad7b Thank you again for the PR, and for the help with the FastLED library.

PR #907, opened 2019-10-23
Add Atmega16 support to fastpin_avr.h (Mightycore Arduino)

cmooney3:
The `atmega16` has the same pinout at the `atmega32` when using the MightyCore board
in arduino.

The pinout according to the Mightcore documentation can be seen here:
image

This commit just adds `__AVR_ATmega16__` to the macro in `fastpin_avr.h` so it builds with the atmega16 target.

It seems like this should be a safe change, but I'm not 100% sure this is how it should be done.  However, I tested this on my Atmega16 board, and appears to work like a charm so I figured it was worth tossing a PR out incase it can help other people :)

Signed-off-by: Charlie Mooney <cmooney3@gmail.com>


kriegsman:
Thank you for the PR, and for the help!

PR #916, opened 2019-11-02
Convert rand multiplication constants to shifts and adds.

grandinquisitor:
Since many popular low-cost chipsets do not have a native multiply operation,
this calculation can be executed much more efficiently using only add and left shift.

### DIFF
--- a/lib8tion/random8.h
+++ b/lib8tion/random8.h
@@ -12,8 +12,11 @@
 #define FASTLED_RAND16_2053  ((uint16_t)(2053))
 #define FASTLED_RAND16_13849 ((uint16_t)(13849))
 
-// equivalent to x * 2053
+#if defined(LIB8_ATTINY)
 #define APPLY_FASTLED_RAND16_2053(x) (x << 11) + (x << 2) + x
+#else
+#define APPLY_FASTLED_RAND16_2053(x) (x * FASTLED_RAND16_2053)
+#endif
 
 /// random number seed
 extern uint16_t rand16seed;// = RAND16_SEED;
--- a/lib8tion/random8.h
+++ b/lib8tion/random8.h
@@ -12,13 +12,16 @@
 #define FASTLED_RAND16_2053  ((uint16_t)(2053))
 #define FASTLED_RAND16_13849 ((uint16_t)(13849))
 
+// equivalent to x * 2053
+#define APPLY_FASTLED_RAND16_2053(x) (x << 11) + (x << 2) + x
+
 /// random number seed
 extern uint16_t rand16seed;// = RAND16_SEED;
 
 /// Generate an 8-bit random number
 LIB8STATIC uint8_t random8()
 {
-    rand16seed = (rand16seed * FASTLED_RAND16_2053) + FASTLED_RAND16_13849;
+    rand16seed = APPLY_FASTLED_RAND16_2053(rand16seed) + FASTLED_RAND16_13849;
     // return the sum of the high and low bytes, for better
     //  mixing and non-sequential correlation
     return (uint8_t)(((uint8_t)(rand16seed & 0xFF)) +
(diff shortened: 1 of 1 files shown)


kriegsman:
Hi @grandinquisitor -  Thanks for the PR, and for the help.  And on a personal note, I really appreciate the focus on speed of execution!

I know that the ATtiny microcontroller line has no on-chip hardware multiplier, but other than that, I think all of the other chips that FastLED supports have a hardware multiplier.  Am I forgetting some other of our supported chips that don't have a multiply unit?  

Also, I find myself wondering if the compiler itself will turn multiplication-by-a-compile-time-constant into a combination of shifts and adds in this case.  I know that it does when you do things like x *= 2, and other 'no-brainer' cases.  And of course a lot of that depends on how 'new' the compiler version is, too.  

Anyway, I'm interested in this PR with two main questions: first, which 'popular low-cost chipsets' do not have a hardware multiplier other than the ATtiny.  And second, do you have benchmark data yet that you could share about the speed change resulting from this code change?  Run it a million times in a loop, that sort of thing.  Any actual benchmark data would be appreciated; I always like making our library faster.

Thanks-
-Mark

kriegsman:
Writing some notes here:  
On ATmega (which has MUL instruction), the plain multiply is 16 instructions, but the shift-add implementation is 23 instructions.
https://godbolt.org/z/DkpTLm

Haven't checked ATtiny yet.

 @grandinquisitor Are there other FastLED-supported platforms that don't have a hardware multiplier, other than ATtiny?

grandinquisitor:
Hi Mark,

Thanks for verifying the number of instructions. That's a cool tool.

I was hoping that it would still come out ahead on the atmega because while it has a MUL instruction, it's only 1 or 2 instructions if it's a single byte, but looks like it would expand it to ~19 instructions if it's a 2 byte int, per doc AVR201. And in this case it's a 2 byte int.

http://ww1.microchip.com/downloads/en/AppNotes/Atmel-1631-Using-the-AVR-Hardware-Multiplier_ApplicationNote_AVR201.pdf

It looks like it still comes out behind according to that tool.

Per code size, I noticed that it's 10 fewer bytes of compiled hex for attiny85 but is 8 more bytes for the arduino uno. (though, flash space is an order of magnitude more precious for attiny's).

I was able to get Serial output working on an attiny. Here are my results.

arduino uno (16mhz, external crystal):

multiply = 1320ms / 1 million executions

shift&add = 1383ms / 1 million executions


attiny24a (8mhz internal oscillator):

multiply = 21338ms / 1 million executions

shift&add = 3321ms / 1 million executions


So, in my tests it was 4.78% slower on the arduino uno, but was 84.44% faster for the attiny. There was very little variance in the results for any of the conditions.

So this is obviously a huge improvement for the attiny but does have a small negative cost for the atmega, which is presumably much more popular.

Is this the list of supported platforms? https://github.com/FastLED/FastLED#supported-platforms . AFAICT, only the attiny-based boards would not have a native multiply.

grandinquisitor:
FWIW a slight alternative occurred to me:

`#define APPLY_FASTLED_RAND16_2053(x) (x << 11) + x + x + x`

This is now only comes at a cost of 4 bytes of hex (vs 8), 5 extra instructions in that compiler tool (vs. 7 extra), and is only 4.2% slower (vs. 4.8% slower above) in my tests for the arduino uno.

kriegsman:
All of this talk about how to do specific efficient multiplies on an 8-bit microcontroller without a hardware multiplier, and without a barrel shifter, reminds me of programming the 6502, back in the day.

So: I am tempted to make this new implementation the default on ATtiny (or other MUL-less), but leave it with the multiply method on other platforms.  I checked ARM (and PowerPC, just for fun), and the multiply method is never slower or larger than the shift method, and it is often a little smaller and faster, as it is on ATmega.  I didn't check ESP/xtensa but I suspect it's similar.

There are other ATtiny-specific optimizations that we could also make (the 'scale8' family of functions comes to mind), but I'm mostly tempted not to pursue that right now, given that all of our other supported MCUs have single-cycle hardware multipliers.

We already have LIB8_ATTINY as a manifest constant that we can #ifdef.  Actually, now that I look at it, it looks like we may have introduced a bug into detection of ATtiny at compile time!   Hrm.

kriegsman:
Thanks for the contribution, and for the help with FastLED!

Issue #928, opened 2019-11-29
RGBCalibrate example - index out of bounds


toybuilder:

toybuilder:
Small change made to fix this in the example. Pull request made.

zackees:
Closing — fixed. @toybuilder's PR landed.

The example now declares `NUM_LEDS 7` (`examples/RGBCalibrate/RGBCalibrate.ino:30`), and the writes are `leds[0]` through `leds[6]` — indices 0–6, exactly 7 elements. In bounds.

Good catch, and the crash-on-ESP32 detail was the useful part: on AVR an out-of-bounds LED write silently corrupts adjacent memory (see #887, where it clobbered a neighbouring variable and looked like a variable spontaneously resetting), while the ESP32 faults outright. Same bug, and the platform that fails loudly is the one that makes it findable.

Worth noting the example is *meant* to have 7 entries — one per calibration colour plus a trailing off pixel — so raising `NUM_LEDS` was the right fix rather than dropping the last write.

PR #929, opened 2019-11-29
Index out-of-bounds

toybuilder:
Accessing leds[6] when NUM_LEDS==6 causes out-of-bounds access.  On ESP32, this causes an out-of-bounds exception and the program will not work.

Issue #928

### DIFF
--- a/examples/RGBCalibrate/RGBCalibrate.ino
+++ b/examples/RGBCalibrate/RGBCalibrate.ino
@@ -23,7 +23,7 @@
 //
 //////////////////////////////////////////////////
 
-#define NUM_LEDS 6
+#define NUM_LEDS 7
 
 // Data pin that led data will be written out over
 #define DATA_PIN 6
--- a/examples/RGBCalibrate/RGBCalibrate.ino
+++ b/examples/RGBCalibrate/RGBCalibrate.ino
@@ -66,7 +66,7 @@ void loop() {
     leds[3] = CRGB(0,0,255);
     leds[4] = CRGB(0,0,255);
     leds[5] = CRGB(0,0,255);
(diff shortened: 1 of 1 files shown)


jspolsky:
That is expected. When a C array has 6 elements they are numbered 0 through 5.

MartyMacGyver:
Note that this and a lot of other things are pending merge from https://github.com/FastLED/FastLED/pull/879 ... once they start merging changes again.

kriegsman:
Hi @toybuilder- and thank you very much for the PR, and the help.

Could I ask for a slight modification?  Instead of removing the write of black to  leds[6] which is out of bounds with the array declared as-is, could you instead change the definition of the leds array to be 7, so that the last pixel can be safely set to black?

We're trying to get things moving again here, and I really appreciate all the help with it.

toybuilder:
> Hi @toybuilder- and thank you very much for the PR, and the help.

You are welcome!  

> Could I ask for a slight modification? Instead of removing the write of black to leds[6] which is out of bounds with the array declared as-is, could you instead change the definition of the leds array to be 7, so that the last pixel can be safely set to black?

Done.  I'm still getting my feet wet with working with git under Github, so I hope the way I updated the branch is suitable for you.

> We're trying to get things moving again here, and I really appreciate all the help with it.

My pleasure.  I plan to use FastLED more in my future, so thank *you* and all other contributors!

MartyMacGyver:
@kriegsman Can you give a look to my PR #879 , which covers this and a lot of other related cleanups? Thanks!

kriegsman:
This problem was taken care of in #879; closing this PR.  Thanks for the help.

PR #941, opened 2019-12-17
Fix typos causing blue LEDs to not clear

sharpchris:
Correction so that the Blue LEDs are properly blacked out


kriegsman:
Thank you very much for the PR and the help!

Issue #942, opened 2019-12-20
Laggy LEDs when doing multiple instructions


MisterGoodDeal:
Hello, 
I'm using _FastLED.h_ on my Arduino Leonardo Pro Micro and I've got an issue. When I'm using the default rainbow effect with no other instructions (just "Make a rainbow pls"), every thing is working smoothly, but when I'm writting other instructions in my code like controlling LEDs through serial port or coding"switch case"... ; the LEDs effect seems to be a bit laggy/LEDs goes red I think.
Here is the [video](https://youtu.be/Fcb78X6y3u0) showing my problem and here is the [LEDs I've bought on Amazon](https://www.amazon.fr/gp/product/B01CP5PQU6/ref=ppx_yo_dt_b_asin_title_o01_s00?ie=UTF8&psc=1)
And here is my code if it can help : 

```
String data;

// Gestion des LEDs
#include <FastLED.h>
#define DATA_PIN    9
//#define CLK_PIN   4
#define LED_TYPE    WS2811
#define COLOR_ORDER GRB
#define NUM_LEDS    50
CRGB leds[NUM_LEDS];

#define BRIGHTNESS          50
#define FRAMES_PER_SECOND  300

void setup() {
    delay(3000); // Delai 3 secondes
    Serial.begin(9600);
    Serial.setTimeout(50);

    // Initialisatation des LEDs
    FastLED.addLeds<LED_TYPE,DATA_PIN,COLOR_ORDER>(leds, NUM_LEDS).setCorrection(TypicalLEDStrip);
    FastLED.setBrightness(BRIGHTNESS);
}

uint8_t gHue = 0; // Variable de modification de la couleur pour les LEDs en mode rainbow
bool eclairage = true;
int type = 0;
int rapiditeRainbow = 50;

void loop() {
  gestionLEDs();
  gestionMain();
}

void gestionLEDs() {
  if(eclairage == true) {
    switch(type) {
      case 0:
        rainbow();
        break;
    }
  }
}

void gestionMain() {
    while(Serial.available()) {
      data = Serial.readString();// read the incoming data as string
      Serial.println(data);
    }
    if (data == "light=on")
      eclairage = true;
    else if (data == "light=off") {
      eclairage = false;
      eteindreLEDs();
    }
}

void rainbow() {
  fill_rainbow( leds, NUM_LEDS, gHue, 7);
  FastLED.show(); // Afficher les mises à jour des LEDs
[...]

PR #943, opened 2019-12-20
Random red pixel fix

MisterGoodDeal:
This change fix the issue with the random red pixel while using the rainbow effect based on issues #668 and #942

### DIFF
--- a/colorutils.cpp
+++ b/colorutils.cpp
@@ -40,7 +40,7 @@ void fill_rainbow( struct CRGB * pFirstLED, int numToFill,
     CHSV hsv;
     hsv.hue = initialhue;
     hsv.val = 255;
-    hsv.sat = 240;
+    hsv.sat = 255;
     for( int i = 0; i < numToFill; i++) {
         pFirstLED[i] = hsv;
         hsv.hue += deltahue;
@@ -54,7 +54,7 @@ void fill_rainbow( struct CHSV * targetArray, int numToFill,
     CHSV hsv;
     hsv.hue = initialhue;
     hsv.val = 255;
-    hsv.sat = 240;
+    hsv.sat = 255;
     for( int i = 0; i < numToFill; i++) {
         targetArray[i] = hsv;
         hsv.hue += deltahue;


MisterGoodDeal:
This change fix the issue with the random red pixel while using the rainbow effect based on issues #668 and #942

kriegsman:
Hi and thanks for the PR. We're just starting to work on the backlog of contributions, and I just wanted to let you know that I was taking a look at this now.  I really appreciate the help.

nenadicpetar:
Thank you a lot, man.

Jostikas:
Fixed my jumping red led with:
```
            if( r ) r = scale8_LEAVING_R1_DIRTY( r, val) + 1;
            asm volatile("");
            if( g ) g = scale8_LEAVING_R1_DIRTY( g, val) + 1;
            asm volatile("");
            if( b ) b = scale8_LEAVING_R1_DIRTY( b, val) + 1;
            asm volatile("");
```
in all places in hsv2rgb_rainbow() where the scale8_LEAVING_R1_DIRTY appeared...

Haven't tested it, don't exactly understand the issue. But I do understand the lower saturation in `fill_rainbow()` was there for a purpose and I do like the uniform brightness it enables.

zackees:
I'm not in favor of changing the saturation of the rainbow at this time. Too risky.

The asm noop fix is very interesting. But the AVR chipsets are becoming less relevant each year. If someone has a new PR with the ASM changes (the comment above is not definitive) then I'll accept it.

zackees:
Hi there, I'm looking at this again and would like to get this old PR fix in. Overall it looks good and I have absolutely noticed bad color with the hue algorithm in many tests.

What I'm noticing here is that you missed a few places. Wanted to give you a chance to review before I proceed.

In src/colorutils.cpp I notice that there are four rainbow functions that each set the set `hsv.sat = 240`


Here is what I currently see with ripgrep:

image


**1**

```C++
void fill_rainbow( struct CRGB * targetArray, int numToFill,
                  uint8_t initialhue,
                  uint8_t deltahue )
{
    CHSV hsv;
    hsv.hue = initialhue;
    hsv.val = 255;
    hsv.sat = 240;
    for( int i = 0; i < numToFill; ++i) {
        targetArray[i] = hsv;
        hsv.hue += deltahue;
    }
}
```

**2**

```C++
void fill_rainbow( struct CHSV * targetArray, int numToFill,
                  uint8_t initialhue,
                  uint8_t deltahue )
{
    CHSV hsv;
    hsv.hue = initialhue;
    hsv.val = 255;
    hsv.sat = 240;
    for( int i = 0; i < numToFill; ++i) {
        targetArray[i] = hsv;
        hsv.hue += deltahue;
    }
}
```
**3**

```

void fill_rainbow_circular(struct CRGB* targetArray, int numToFill, uint8_t initialhue, bool reversed)
{
    if (numToFill == 0) return;  // avoiding div/0

    CHSV hsv;
    hsv.hue = initialhue;
    hsv.val = 255;
    hsv.sat = 240;

    const uint16_t hueChange = 65535 / (uint16_t)numToFill;  // hue change for each LED, * 256 for precision (256 * 256 - 1)
    uint16_t hueOffset = 0;  // offset for hue value, with precision (*256)

    for (int i = 0; i < numToFill; ++i) {
        targetArray[i] = hsv;
        if (reversed) hueOffset -= hueChange;
        else hueOffset += hueChange;
        hsv.hue = initialhue + (uint8_t)(hueOffset >> 8);  // assign new hue with precise offset (as 8-bit)
    }
}
```

**4**

```

void fill_rainbow_circular(struct CHSV* targetArray, int numToFill, uint8_t initialhue, bool reversed)
{
    if (numToFill == 0) return;  // avoiding div/0

    CHSV hsv;
    hsv.hue = initialhue;
    hsv.val = 255;
    hsv.sat = 240;

    const uint16_t hueChange = 65535 / (uint16_t) numToFill;  // hue change for each LED, * 256 for precision (256 * 256 - 1)
    uint16_t hueOffset = 0;  // offset for hue value, with precision (*256)

    for (int i = 0; i < numToFill; ++i) {
        targetArray[i] = hsv;
        if (reversed) hueOffset -= hueChange;
        else hueOffset += hueChange;
        hsv.hue = initialhue + (uint8_t)(hueOffset >> 8);  // assign new hue with precise offset (as 8-bit)
    }
}
```

sutaburosu:
@zackees can we please get the [suggested changes for AVR](https://github.com/FastLED/FastLED/pull/943#issuecomment-845751972) at the same time. Iirc, this value of 240 for saturation was reduced from 255 to work around a very old, and previously misunderstood issue on AVR. Since then, I've seen it manifest in my own uses of scale8_LEAVING_R1_DIRTY. The proposed `volatile asm("");` additions should fix the original problem and allow saturations up to 255 to work correctly on AVR too. They serve as sequence points, so the optimiser cannot reorder the asm()s.

zackees:
> Fixed my jumping red led with:
> 
> ```
>             if( r ) r = scale8_LEAVING_R1_DIRTY( r, val) + 1;
>             asm volatile("");
>             if( g ) g = scale8_LEAVING_R1_DIRTY( g, val) + 1;
>             asm volatile("");
>             if( b ) b = scale8_LEAVING_R1_DIRTY( b, val) + 1;
>             asm volatile("");
> ```
> 
> in all places in hsv2rgb_rainbow() where the scale8_LEAVING_R1_DIRTY appeared...
> 
> Haven't tested it in other cases but mine, don't exactly understand the issue and don't know if it changes the execution time (on AVRs, it shouldn't, right?). But I do understand the lower saturation in `fill_rainbow()` was there for a purpose and I do like the uniform brightness it enables.

Thanks @sutaburosu 

Here is the new pr that fixes this: https://github.com/FastLED/FastLED/pull/1907

zackees:
@sutaburosu okay the `asm("")` fix is in. Am I correct that this means this issue is now fixed and can be closed?

sutaburosu:
I'm sorry, I haven't had time in the last couple of days to put together the visual test case that I wanted to provide. I tried briefly, but this problem can be difficult to reproduce when you actually want to see it.

After skimming the linked issues, I notice [this commentor reached the same conclusion](https://github.com/FastLED/FastLED/issues/668#issuecomment-562924729): prevent the compiler re-ordering anything in this stanza.

My vote is to close this and the others.

zackees:
ok it's fixed. Will reopen if not fixed

PR #950, opened 2020-01-08
Improvements to ESP32 support in the RMT driver

samguyer:
This pull request has a few substantial changes. The main goal is to eliminate interrupt-related problems that cause flashy weirdness when FastLED is used in conjunction with the WiFi and web support libraries.

Change 1: All RMT channels are started at the same time. In the previous version, we would fill each channel buffer and then start it immediately. But in cases where there are many parallel LED strips, it looks like the earlier channels were already getting refill interrupts while the later channels were still initializing their buffers, causing timing errors. The symptom is that the higher the channel number, the worse the timing errors.

Change 2: Each interrupt encodes exactly 1 pixel's worth of data, simplifying the fill routine and speeding it up.

Change 3: The refill interrupt is created at level 3. I'm not sure if this makes a huge difference, but we really don't want LED output to be disrupted unless there is a damn good reason.


samguyer:
Thanks @BlueAndi 
I will take a look at the SPIFFS issue.

Issue #905, opened 2019-10-21
Any one tested the ''new'' Arduino Nano 33 ioT?


alfaaleksi:
Since it's based on  SAMD21G18A chip and has a different  pin map to that of the regular Nano I'm wondering if one could use the FastLED library on projects utilizing the Nano 33.

nittanygeek:
Anything become of this?  I can't seem to get it working on mine.

PR #951, opened 2020-01-08
Fix Nano 33 IOT pindef

Ampt:
The pins here were based on an early draft of the Nano 33 IOT package where the pin definitions were mislabeled in the comments - https://github.com/arduino/ArduinoCore-samd/blob/08629f90b1f803017cc526c4cbbe3eaaccc8f062/variants/nano_33_iot/variant.cpp

This has recently been fixed - 
https://github.com/arduino/ArduinoCore-samd/blob/master/variants/nano_33_iot/variant.cpp

This change uses the correct pin definitions.

I chose to use the last SPI, but we might want to change that to something else, as it's used by the WiFi module and I'm not exactly sure what FastLED plans to use it for, but we may run into conflicts.


Ampt:
This should fix https://github.com/FastLED/FastLED/issues/905

kriegsman:

PR #956, opened 2020-01-21
Add platform pin mappings for Arduino MKR WiFi 1010

anoadragon453:
Allows this library to compile for the [Arduino MKR WiFi 1010](https://store.arduino.cc/arduino-mkr-wifi-1010). The MKR 1000 and MKR WiFi 1010 share the same mappings.

Tested and works with my own 1010.

Source: https://github.com/FastLED/FastLED/issues/414#issuecomment-444734818


kriegsman:

PR #958, opened 2020-01-26
wrong comment location

8633brown:
```set_max_power_indicator_LED``` has not been depracated and is still use in ```power_mgt.cpp```


kriegsman:
Thank you for the contribution, and the help with the FastLED library!

PR #961, opened 2020-02-03
fix for older gcc versions

rocketstrong600:
fixes clockless_trinket.h:326:24: error: expected primary-expression before '__attribute__' on older gcc versions

### DIFF
--- a/platforms/avr/clockless_trinket.h
+++ b/platforms/avr/clockless_trinket.h
@@ -323,14 +323,10 @@ protected:
 #define DUSE (0xFF - (DADVANCE-1))
 
 // Silence compiler warnings about switch/case that is explicitly intended to fall through.
-#if defined(__GNUC__) && __GNUC__ >= 7
- #define FL_FALLTHROUGH __attribute__ ((fallthrough));
-#else
- #define FL_FALLTHROUGH ((void)0);
-#endif /* __GNUC__ >= 7 */
+//#define FL_FALLTHROUGH __attribute__ ((fallthrough));
 
-	// This method is made static to force making register Y available to use for data on AVR - if the method is non-static, then
-	// gcc will use register Y for the this pointer.
+// This method is made static to force making register Y available to use for data on AVR - if the method is non-static, then
+// gcc will use register Y for the this pointer.
 	static void /*__attribute__((optimize("O0")))*/  /*__attribute__ ((always_inline))*/  showRGBInternal(PixelController<RGB_ORDER> & pixels)  {
 		uint8_t *data = (uint8_t*)pixels.mData;
 		data_ptr_t port = FastPin<DATA_PIN>::port();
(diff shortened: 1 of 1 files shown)


kriegsman:
Thanks for the PR!  I know we're working with a wide range of compilers here -- from old gcc, to newer gcc, to also (believe it or not) clang.  I definitely want to find a solution here that works for all of these.  Have you tried this one with (current) clang?

rocketstrong600:
Forgot about clang yeah changes will have to be made for clang I'm pretty sure

kriegsman:
I believe that there’s an Official way to do this now in C++11 or 14. Maybe we can ifdef testing for new C++, and if not defined, fall back to the void trick? That’ll cover contemporary compilers correctly, and older ones compatibly?

kriegsman:
Oops, apparently I meant C++17:
“[[fallthrough]];”

rocketstrong600:
As far as I am aware the void trick works on almost everything else but will have do some research and make sure.

platforms/avr/clockless_trinket.h
```diff
@@ -323,7 +323,11 @@ class ClocklessController : public CPixelLEDController<RGB_ORDER> {
 #define DUSE (0xFF - (DADVANCE-1))
 
 // Silence compiler warnings about switch/case that is explicitly intended to fall through.
-#define FL_FALLTHROUGH __attribute__ ((fallthrough));
+#if defined(__GNUC__) && __GNUC__ >= 7
+ #define FL_FALLTHROUGH __attribute__ ((fallthrough));
+#else
+ #define FL_FALLTHROUGH ((void)0);
```
JulienVdG: you could just
#define FL_FALLTHROUGH
here, the ((void)0) trick is to avoid adding no instruction before the semicolon (in case it's not in the macro) so here it can be defined to 'empty' and works

rocketstrong600: I like the comment trick I will try it.

rocketstrong600:
Tried The /* fall through */ comment compiles fine on both gcc versions i was using. i don't have a setup to test clang

kriegsman:
Thanks for looking into the options here. I'm going to see if I can test it on clang, where I think the comment method works, but I'd like to be sure.

Ideally, I'd use the newer C++17 standard way to do this on C++17 and later compilers, but I'm not sure it's worth the fuss, and the comment method is backward compatible -- even if it makes things more of a PITA for future people, some of whom may be ... us.

kriegsman:
Well, let's go with this for now since it works on old and new gcc.  If we have issues with clang, we'll fix them when we get there.

kriegsman:
Thank you for the help and the contributions to the FastLED library!

samguyer:
Hello all, I am going to reapply this PR manually. I messed something up during a big commit, which ended up rolling back a bunch of good changes (including this one).

PR #963, opened 2020-02-07
Overload array subscript operator for CHSV structs

jackw01:
RGB values in CRGB structs can be accessed through the array subscript operator but the values in CHSV structs can not.


kriegsman:
Thanks for the contribution and thanks for helping with FastLED!

PR #965, opened 2020-02-10
Better timing stability for ESP32

samguyer:
(1) A forced minimum delay of 50 microseconds between calls to show(). The way the driver is written it does not explicitly send a latching signal. I noticed that on the ESP32 it is possible to call show() so quickly that the strip doesn't latch, and the new signal appears to be an extension of the previous one.

(2) An option to lock out SPI flash access (e.g., using SPIFFS) during a call to show(). SPI flash access on the ESP32 requires a lockdown of the memory bus. To enforce this requirement, the ESP-IDF system software disables all tasks on *both* cores until flash operations complete. If this lockout happens during a call to show(), the timing gets off, and the LEDs look flashy. I added a switch that causes the driver to acquire a lock on the flash system, effectively delaying the flash operations until show() is complete. You can enable it by adding the following line *before* you include FastLED.h:

#define FASTLED_ESP32_FLASH_LOCK

### DIFF
--- a/platforms/esp/32/clockless_rmt_esp32.h
+++ b/platforms/esp/32/clockless_rmt_esp32.h
@@ -64,7 +64,7 @@
  *      this from happening, define this flag. It will force flash
  *      operations to wait until the show() is done.
  *
- * #define FASTLED_ESP32_FLASH_LOCK
+ * #define FASTLED_ESP32_FLASH_LOCK 1
  *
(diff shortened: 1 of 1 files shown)


kriegsman:
Awesome- thanks, @samguyer!  Glad for the great things you're doing with ESP32 and FastLED!

Issue #971, opened 2020-02-16
esp32 IDF v4.0


PhilColbert:
Anyone got fastled running on the latest esp32 code ?

Thanks

littleblackfish:
What would be the most straightforward way? I am trying to follow this with little luck so far. I could build arduino-esp32 as a component, but not FastLED so far. 

https://github.com/espressif/arduino-esp32/blob/idf-release/v4.0/docs/esp-idf_component.md

littleblackfish:
After adding a `add_definitions(-DESP32)` on the top CmakeLists.txt file and no other changes, here is where I'm at : 

```
In file included from ../components/arduino/libraries/FastLED/FastLED.h:58,
                 from ../components/arduino/libraries/FastLED/FastLED.cpp:2:
../components/arduino/libraries/FastLED/colorutils.h: In copy constructor 'CHSVPalette16::CHSVPalette16(const CHSVPalette16&)':
../components/arduino/libraries/FastLED/colorutils.h:455:69: error: 'void* memmove(void*, const void*, size_t)' writing to an object of type 'struct CHSV' with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Werror=class-memaccess]
         memmove8( &(entries[0]), &(rhs.entries[0]), sizeof( entries));
                                                                     ^
In file included from ../components/arduino/libraries/FastLED/controller.h:9,
                 from ../components/arduino/libraries/FastLED/FastLED.h:47,
                 from ../components/arduino/libraries/FastLED/FastLED.cpp:2:
../components/arduino/libraries/FastLED/pixeltypes.h:23:8: note: 'struct CHSV' declared here
 struct CHSV {
        ^~~~
In file included from ../components/arduino/libraries/FastLED/FastLED.h:58,
                 from ../components/arduino/libraries/FastLED/FastLED.cpp:2:
../components/arduino/libraries/FastLED/colorutils.h: In member function 'CHSVPalette16& CHSVPalette16::operator=(const CHSVPalette16&)':
../components/arduino/libraries/FastLED/colorutils.h:459:69: error: 'void* memmove(void*, const void*, size_t)' writing to an object of type 'struct CHSV' with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Werror=class-memaccess]
         memmove8( &(entries[0]), &(rhs.entries[0]), sizeof( entries));
                                                                     ^
In file included from ../components/arduino/libraries/FastLED/controller.h:9,
[...]

PR #977, opened 2020-02-27
cast memmove parameters to (void *) to avoid class-memaccess warnings

littleblackfish:
Solves #971 by matching stdlib's memmove signature, should be pretty harmless otherwise.

### DIFF
--- a/colorutils.h
+++ b/colorutils.h
@@ -452,11 +452,11 @@ public:
 
     CHSVPalette16( const CHSVPalette16& rhs)
     {
-        memmove8( &(entries[0]), &(rhs.entries[0]), sizeof( entries));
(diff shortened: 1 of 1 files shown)


kriegsman:
I think that these conflicts are ok to resolve now that other code has been merged, but could you take a look and verify, @littleblackfish?  Thanks for the help, @littleblackfish and @5chmidti -- much appreciated.

kriegsman:
I think we're getting ready to merge this PR once the (new) conflicts are resolved. We just made some changes to the FastLED directory structure, so I suspect that it'll be easy for you to resolve the conflicts and get this ready for merging. Thank you for the help and the contribution!

5chmidti:
@kriegsman due to #990 this shouldn't be needed.

PR #986, opened 2020-03-24
Squelch newer GCC warning on memmove8

marcmerlin:
Discussion in  https://github.com/FastLED/FastLED/issues/985

Issue #985, opened 2020-03-19
FastLED compiler warnings on gcc/linux


marcmerlin:
I'm not sure if they are worth fixing/easy to fix. Full list attached:
[out.txt](https://github.com/FastLED/FastLED/files/4356359/out.txt)

```
libraries/FastLED/colorutils.h: In copy constructor ‘CHSVPalette16::CHSVPalette16(const CHSVPalette16&)’:
libraries/FastLED/colorutils.h:455:69: warning: ‘void* memmove(void*, const void*, size_t)’ writing to an object of type ‘struct CHSV’ with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
         memmove8( &(entries[0]), &(rhs.entries[0]), sizeof( entries));
                                                                     ^
In file included from libraries/FastLED/controller.h:9,
                 from libraries/FastLED/FastLED.h:47,
                 from examples/GFX_Table_Mark_Estes/neomatrix_config.h:112,
                 from examples/GFX_Table_Mark_Estes/Table_Mark_Estes.ino:5,
                 from src/main.cpp:13:
libraries/FastLED/pixeltypes.h:23:8: note: ‘struct CHSV’ declared here
 struct CHSV {
        ^~~~
In file included from libraries/FastLED/FastLED.h:58,
                 from examples/GFX_Table_Mark_Estes/neomatrix_config.h:112,
                 from examples/GFX_Table_Mark_Estes/Table_Mark_Estes.ino:5,
                 from src/main.cpp:13:
libraries/FastLED/colorutils.h: In member function ‘CHSVPalette16& CHSVPalette16::operator=(const CHSVPalette16&)’:
libraries/FastLED/colorutils.h:459:69: warning: ‘void* memmove(void*, const void*, size_t)’ writing to an object of type ‘struct CHSV’ with no trivial copy-assignment; use copy-assignment or copy-initialization instead [-Wclass-memaccess]
         memmove8( &(entries[0]), &(rhs.entries[0]), sizeof( entries));
                                                                     ^
In file included from libraries/FastLED/controller.h:9,
                 from libraries/FastLED/FastLED.h:47,
                 from examples/GFX_Table_Mark_Estes/neomatrix_config.h:112,
                 from examples/GFX_Table_Mark_Estes/Table_Mark_Estes.ino:5,
                 from src/main.cpp:13:
libraries/FastLED/pixeltypes.h:23:8: note: ‘struct CHSV’ declared here
[...]

PR #990, opened 2020-03-28
use default copy constructor and copy assignments for CRGB and CHSV

5chmidti:
fix for class-memaccess warnings for the `CRGB` and `CHSV` structs.
fixes the same warnings as #986 (discussion #985) and #977 (discussion #971)(which are duplicates), except that this removes the warning from all uses that require `CRGB` and `CHSV` to be trivially copyable.

As littlebalckfish [said](https://github.com/FastLED/FastLED/issues/971#issuecomment-591734979)
<details><summary>Quote</summary>

> memmove signature from stdlib is
> 
> `void * memmove ( void * destination, const void * source, size_t num );`
> 
> So, casting all the pointers to (void*), I believe, is an appropriate, if not optimal solution to the problem, i.e.
> 
> `memmove8( &(entries[0]), &(rhs.entries[0]), sizeof( entries));`
> 
> to
> 
> `memmove8( (void *) &(entries[0]), (void *) &(rhs.entries[0]), sizeof( entries));`
> 
> etc.
> 
> This builds without complaints. I'd love to do a pull request but I have no way to test this with all the platforms, is there a recommended way to proceed?

</details>

the `memmove` signature takes `void *` as arguments and the warnings could be avoided by casting to `void *`, but the warnings could be completely removed for `CRGB` and `CHSV` by making both have a trivial copy assignment and a trivial copy constructor. Thus removing warnings when using for example `memmove` in user code and in the future.

Performance wise this shouldn't make a difference, but please test it. I've done a bit of testing myself and haven't found any real performance difference although I'm not sure how good my tests are.

### DIFF
--- a/pixeltypes.h
+++ b/pixeltypes.h
@@ -62,20 +62,9 @@ struct CHSV {
     }
 
     /// allow copy construction
-    inline CHSV(const CHSV& rhs) __attribute__((always_inline))
-    {
-        h = rhs.h;
-        s = rhs.s;
-        v = rhs.v;
-    }
+    inline CHSV(const CHSV& rhs) __attribute__((always_inline)) = default;
 
-    inline CHSV& operator= (const CHSV& rhs) __attribute__((always_inline))
-    {
-        h = rhs.h;
-        s = rhs.s;
-        v = rhs.v;
-        return *this;
-    }
+    inline CHSV& operator= (const CHSV& rhs) __attribute__((always_inline)) = default;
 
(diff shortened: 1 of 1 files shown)


5chmidti:
I've noticed that I could've made the default constructor default constructable.
Also [this](https://godbolt.org/z/bKTYv5) as a representation of avr assembly. This might have performance improvements after all (looking at avr asm, esp32 asm reordered three instructions but with no difference in used instructions)

kriegsman:
Thanks for the help with this, and the code fixes.

PR #991, opened 2020-04-02
add platform support for the Seeeduino XIAO board

5kft:
This PR adds support to FastLED for the Seeeduino XIAO Cortex-M0+ board [(http://wiki.seeedstudio.com/Seeeduino-XIAO/).](http://wiki.seeedstudio.com/Seeeduino-XIAO/).  

The mapping was derived from [https://github.com/Seeed-Studio/ArduinoCore-samd/blob/master/variants/XIAO_m0/variant.cpp](https://github.com/Seeed-Studio/ArduinoCore-samd/blob/master/variants/XIAO_m0/variant.cpp), and I verified these changes on one of my XIAO boards.

Thanks!

### DIFF
--- a/platforms/arm/d21/fastpin_arm_d21.h
+++ b/platforms/arm/d21/fastpin_arm_d21.h
@@ -97,6 +97,17 @@ _FL_DEFPIN(17,  4, 0);   _FL_DEFPIN(18,  5, 0);   _FL_DEFPIN(19, 6, 0);
 
 #define HAS_HARDWARE_PIN_SUPPORT 1
 
+#elif defined(SEEED_XIAO_M0)
+
+#define MAX_PIN 10
+_FL_DEFPIN( 0, 2,0); _FL_DEFPIN( 1, 4,0); _FL_DEFPIN( 2,10,0); _FL_DEFPIN( 3,11,0);
+_FL_DEFPIN( 4, 8,0); _FL_DEFPIN( 5, 9,0); _FL_DEFPIN( 6, 8,1); _FL_DEFPIN( 7, 9,1);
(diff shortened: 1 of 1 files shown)


Danielsan-labs:
Hi, it doesn't work for me. Perhaps I miss something. 

I get Fastled working with an Arduino Nano, I added your code to the fastpin_arm_d21.h file and in my code I start with:
#define SEEED_XIAO_M0
#include <FastLED.h>
I'm also using 74AHCT126 to bring the voltage up to 5V. 
Do I need to do anything else?
Thanks!

5kft:
OK, I have several boards running with these changes and they work well...  A few questions:

Are you using the latest official "Seeed SAMD Boards" package version 1.7.1 in Arduino?

In Tools + Board are you selecting "Seeeduino XIAO"?

In making your local changes to FastLED, are you applying the patch as-is?  It's important that in the FastLED code that the patch changes are _before_ the #elif check for ARDUINO_SAMD_ZERO as the Seeed board configuration for the XIAO also #defines that ID.

The data output line for the LEDs needs to be connected to one of the XIAO IOs and the ID should be used appropriately (A0, A1, A2, etc.)

Danielsan-labs:
Hi thanks for your reply, 
I found the issue, it was a silly one as usual.. the LED power cable had a bad connection. I checked all cables twice, but connections on the breadboards are sometimes bad. 

Many thanks! It works fantastic!

5kft:
Great - glad to hear this!

> Hi thanks for your reply, I found the issue, it was a silly one as usual.. the LED power cable had a bad connection. I checked all cables twice, but connections on the breadboards are sometimes bad. Many thanks! It works fantastic!
> […](#)

PR #1001, opened 2020-04-19
Reduce noisy compilation for nrf52 boards

henrygab:
* Mark Adafruit Feather nRF52840 Express as tested board
* Mark Nordic PCA10056 as tested board
* Allow suppression of warning about untested board
* Disable (unused) high-speed port check due to unreliable nature of result

PR #1018, opened 2020-05-19
add support for Adafruit ItsyBitsy nRF52840 Express

ardnew:
simply added pin map definitions based on [Adafruit's variant definition](https://github.com/adafruit/Adafruit_nRF52_Arduino/tree/master/variants/itsybitsy_nrf52840_express).

derived from the unmerged changes in #1001, but also works without them.

verified using only data pin 5 (the 5V output level pin) and an Adafruit Neopixel strip (30px)


AB624:
using dotstar and adafruit itsybitsy nrf52840 express M4.
for the constructor what should I use for the pin number ? (DataPin and clock_pin)
FastLED.addLeds<DOTSTAR, DATA_PIN, CLOCK_PIN, RGB>(leds, NUM_LEDS); 
I want to use the hardware spi. 
Regards

ardnew:
@AB624 from Adafruit's site:
> The DotStar is connected to pin #6 (clock) and #8 (data) in Arduino, so just use our DotStar library and set it up as a single-LED strand on pins 6 & 8.

https://learn.adafruit.com/adafruit-itsybitsy-nrf52840-express/pinouts#qspi-flash-and-dotstar-3052026-15

AB624:
Thanks for your answer,

First of all, the pins 6 and 8 are not identified on the itsy-bitsy nrf52840.
With the adafruit_dotstar library my best refresh rate is 760 us for 50 leds.  I tought that I should get better performance with fastled.
That’s when I use hardware spi  (mosi and sck pins)

Regards,
AB

AB624:
Good morning ardnew,

Actually I have a refresh time of 976 us at best for 50 leds, same number for only 1 led.
Is it normal, can I do better ?

Thanks again,
AB

AB624:
Hi again,

Just to mention that the line : SPI.setClockDivider(SPI_CLOCK_DIV2);
Seems to have no effect...

Sorry but I’m a bit lost here and I need some help to figure it out.

Thank you many times,

AB

ardnew:
> First of all, the pins 6 and 8 are not identified on the itsy-bitsy nrf52840.

Sorry, the Itsybitsy nrf52840 has 1 built-in DotStar onboard connected *internally* to pins 6 and 8 (indeed, these pins are not exposed via board headers), so when you said:

> using dotstar and adafruit itsybitsy nrf52840 express M4. 

I assumed that's what you were referring to. If you're using an external strip of DotStars, then yeah use the following:

The hardware SPI pins are identified in comments in a file in this pull request as `MOSI`/data `24` and `SCLK`/clock `25`, labeled `MO` and `SCK` on the headers.

As for the expected refresh rates I really have no idea, someone else will have to answer

AB624:
Thank you ,

So my constructor should look like that ?

FastLED.addLeds<DOTSTAR, 15, 13, RGB>(leds, NUM_LEDS);  // BGR ordering is typical

but I got an error on compiling :

C:\Users\DELL\Documents\Arduino\libraries\FastLED/platforms/arm/nrf52/fastspi_arm_nrf52.h:139:45:

error: 'nrf_pin' is not a member of 'FastPin<13>'

                 FastPin<_CLOCK_PIN>::nrf_pin(),

Thanks again,
AB

Issue #1017, opened 2020-05-15
FastLED.clear() and FastLED.clearData() clear only the first LED strip


NiekLambert:
Minor issue, easily worked around and probably easily fixed in the library: I am controlling four WS2812 LED strips from a Teensy4. The Arduino 1.8.12 environment with Teensyduino 1.51 reports "FastLED version 3.003.001". Everything works like a charm but when I issue FastLED.clear() or FastLED.clear(true) or FastLED.clearData() only the first LED strip gets cleared which is probably not intended.

MartyMacGyver:
@NiekLambert What is your code doing immediately before and after the `clear()` call? I can reproduce this by calling:

```
  LEDS.clear(true);
  FastLED.delay(1000);
```

This is notable as `clear()` is currently:

```
void CFastLED::clear(bool writeData) {
	if(writeData) {
		showColor(CRGB(0,0,0), 0);
	}
    clearData();
}
```

So I see a brief flicker as the strip sets to all black, then immediately switches back to whatever color it was previously. If I reduce the refresh rate down below 10, it's even more obvious.

What catches my eye is the call to `show()` that's integral to `delay()`'s busy loop... I don't understand the purpose of that, but it certainly has an impact (however, on the Teensy, removing it (while still ensuring a normal call to `FastLED.show();` in the example code) causes some oddness when the Serial port is connected - again).

Under all this, though, are calls to `millis()` and `micros()` - I have some suspicion that there's something about Teensyduino's implementation of these (and Serial as well) that's leading to side-effects. It's just a hunch for now, but I'm seeing things I can't explain purely with FastLED code. (For instance, using regular `delay()` instead of `FastLED.delay()` can cause odd behavior if the Serial port is connected). I think the common denominator may be interrupts.

Will update as I learn more.

MartyMacGyver:
Well, I can see what the root cause is now in the library - not yet sure how to resolve it though.

Given numLanes and numLeds, you need to create an `leds` array thus:

`CRGB leds[numLanes * numLeds];`

When you use the `FastLED.addLeds<numLanes, WS2812B, DATA_PIN, RGB>(leds, numLeds);` it creates a controller with `numLanes` and `numLeds`. 

Each controller knows its own size, BUT `clear()` calls `clearData()` which calls `clearLedData()` which finally calls

`memset8((void*)m_Data, 0, sizeof(struct CRGB) * m_nLeds);`

[...]

PR #1021, opened 2020-05-25
Ensure getMaxRefreshRate() returns a non-zero default for Teensy 4.x

### LINKED ISSUE #912: teensy 4.0 only parallel output works, garbage output when FastLED.delay() is used
With a teensy 4.0 only parallel output works. When compiling with without the parallel output it compiles fine, but the output is garbage.

FastLED.delay() also creates garbage output. The same code changed to delay() works.

MartyMacGyver:
Fixes #912

### DIFF
--- a/platforms/arm/mxrt1062/block_clockless_arm_mxrt1062.h
+++ b/platforms/arm/mxrt1062/block_clockless_arm_mxrt1062.h
@@ -92,6 +92,7 @@ public:
 
   }
 
+  virtual uint16_t getMaxRefreshRate() const { return 400; }
 
   virtual void showPixels(PixelController<RGB_ORDER, LANES, __FL_T4_MASK> & pixels) {
 		mWait.wait();
--- a/platforms/arm/mxrt1062/clockless_arm_mxrt1062.h
+++ b/platforms/arm/mxrt1062/clockless_arm_mxrt1062.h
@@ -38,6 +38,8 @@ public:
     FastPin<DATA_PIN>::lo();
 	}
 
+  virtual uint16_t getMaxRefreshRate() const { return 400; }
+
 protected:
 
 	virtual void showPixels(PixelController<RGB_ORDER> & pixels) {


kriegsman:
Shall I merge this as-is with the "400" plugged in there, or would you like me to wait a bit yet?  I ask because it's marked "WIP".  Thanks for the help!

MartyMacGyver:
I put it on there because I noticed last night that beyond about 64 LEDs the `FastLED.delay()` still causes artifacts on the Teensy 4.x. Dialing down the refresh rate compensates (per the wiki, it seems like setting it to 60 Hz works fine for 512 leds per lane, and 30 Hz for 1024, but I didn't exhaustively test higher values to failure).

**My question for you is, does `FastLED.delay()` cause flickering or other artifacts on other platforms if the refresh rate is too high for a give strip length and the default max refresh rate is too high for that configuration?** (Is `setMaxRefreshRate()` generally needed for long strips?) This is _excluding_ ESP32 using RMT with interrupts (which doesn't have this issue because of how it leverages that combination which is itself unusual versus other devices).

If the user is expected to self-throttle by limiting the max refresh rate for long LED strips, then this is ready to be merged and I'll take the WIP tag off.

MartyMacGyver:
I want to add that just using regular `delay()` on the Teensy 4.x will also cause odd problems if the refresh rate isn't set right (particularly if the Serial connection is open). With my fix though, FastLED.delay() does the job (by default for shorter lengths, and for longer lengths if `setMaxRefreshRate()` is lowered accordingly).

So, I think my fix is helpful regardless (that default ought to be set), but the other oddities (which seem to be some kind of odd interaction between timers, interrupts, and Serial on the Teensy) are worthy of investigation but  out of scope here. Issue #1017 in particular grabs my attention, as I've seen that work some times and not others and it *seems* to be timer related as well. (This branch doesn't address that but I'm clearly seeing that problem as reported, and will see if I can help at all in narrowing it down.)

MartyMacGyver:
WIP tag removed - ready to merge!

kriegsman:
Thank you very much for the help, and the contributions!  Much appreciated!

PR #1030, opened 2020-05-30
Update API reference Doxygen file to 1.8.18; update .gitignore

MartyMacGyver:
I was looking at the documentation on the website at http://fastled.io/docs/ (as linked via https://github.com/FastLED/FastLED/wiki/API-Reference) and saw that it is very out of date (3.1 is the latest).

I installed and used Doxygen 1.8.18 to generate the documentation successfully. This PR allows one to build Doxygen easily, and the .gitignore ensures the output isn't accidentally checked into the repo. The locally built version (which works in and of itself locally) is then suitable for publication to the fastled.io site.


MartyMacGyver:
@kriegsman Likewise, it would be good to get this merged so we can generate current documentation.

MartyMacGyver:
Thank you! Hopefully things will start getting merged more frequently here.

kriegsman:
Thanks for the help keeping us a bit more current with the docs.

PR #936, opened 2019-12-11
ESP8266 SPI

BenoitAnastay:
Implementation of SPI for ESP8266

PR #1047, opened 2020-07-01
Add hardware SPI support on ESP32

ntwallace:
Hardware SPI driver for ESP32, derived from code written by @BenoitAnastay: #936 and the Teensy 4 hardware SPI implementation. Tested on a [Lolin/WeMos D32 Pro](https://docs.wemos.cc/en/latest/d32/d32_pro.html) and two custom WROVER boards.

To use, add `#define FASTLED_ALL_PINS_HARDWARE_SPI` before including FastLED.h. 

The driver uses the VSPI bus by default (pins 18, 19, 23, & 5), but can be set to use the HSPI bus (pins 14, 12, 13, & 15) by adding `#define FASTLED_ESP32_SPI_BUS HSPI` before FastLED.h


samguyer:
Hi Nick! Thanks for the contribution. I've got a couple of questions for you before we merge it in.
What kinds of LED strips have you tested this code on? And do you have a sense of how much faster it is to use SPI?

ntwallace:
Hi Sam, glad to give back to this project. I've used it for years. To answer your questions:

- I have tested this with APA102 and SK9822s. I don't have any other clocked LEDs on hand to test with.
- I haven't done extensive testing, but if getFPS() is anything to go by it's roughly twice as fast. For example, running a fill_rainbow() animation on a strip of 144 APA102s at 24MHz averages ~80FPS when bit banging, ~160FPS with this hardware SPI implementation

BenoitAnastay:
Personally with the ESP8266, I do not have a logic analyzer but the SPI change was the only way to be able to drive my LEDs, it wans't only laggy, the behavior of the bit bangging was random (interupted animation wrong colors, flickering etc, there is reports of such issues on YouTube, SPI implementation resolved them all in my case)

Maybe the bit banging tend to have clocking speed issue and "dephasing" on some hardwares when there is too much LEDs or a too high refresh frequency (just presuming)

endotronic:
What compiler are you using? I was very excited about trying this out, but I can't get it to compile.

The first error I see is
```
chipsets.h:417:3: error: 'mSPI' was not declared in this scope
   mSPI.init();
```

Looking at this, I can see that mSPI is an instance variable. Changing the code to `this.mSPI.init();` fixes it, but there are tons of other instance variables accessed without `this` causing tons of other problems. I'm guessing your compiler is permissive of this and mine is not.

I'm using the Arduino IDE from the Windows Store, 1.8.13 (Windows Store 1.8.42.0). I'm not sure what C compiler is used; it does not tell me.

ntwallace:
Hi @endotronic, that's strange. I have no issues compiling with defaults on Arduino IDE 1.8.13 for MacOS, 1.8.13 for Ubuntu, or PlatformIO 4.3.4. Both Arduino and PIO are using the standard Xtensa GCC compiler for ESP32 setup. I don't have a windows machine to test with unfortunately.

If you enable verbose logging during compilation the Arduino IDE will show you the path to the compiler. Mine looks like this:

endotronic:
Thanks for pointing that out. I forgot to mention that I'm using an ESP8266 and was hopeful that these changes would work with that processor as well, given  the comment by BenoitAnastay. With verbose output, I see this path for gcc (for ESP8266):
`"C:\\Users\\Kevin\\Documents\\ArduinoData\\packages\\esp8266\\tools\\xtensa-lx106-elf-gcc\\2.5.0-4-b40a506/bin/xtensa-lx106-elf-g++"`

It makes sense it wouldn't be the same as the one for ESP32. I will explore this more in a bit.

ntwallace:
Ah, yeah this won't compile for the ESP8266 without some changes, unfortunately. In the FastLED SPI template it's looking for the board to be defined as ESP32. 

The ESP32 has 2 SPI buses, and this is passing a pointer in the ESP32 specific driver to initialize the SPI setup on one or the other.

endotronic:
Ah gotcha. Thanks. I will probably get an ESP32 then. :)

BenoitAnastay:
@endotronic The PR for the ESP8266 is here https://github.com/FastLED/FastLED/pull/936

ChisholmKyle:
I really appreciate you adding this support. It's great! However, I'm using PlatformIO and there are issues compiling.

Part of the problem is that `#include <SPI.h>` is not in any header when using ESP32. I don't think the FastLED avr implementation uses the Arduino.h SPI library (accesses registers directly instead). 

The Arduino IDE compiles and links all the source files at the same time, but PlatformIO compiles source objects individually, so each file is compiled in separate commands, then in the link step combines everything. This allows for fast incremental builds and is a much more common practice for C/C++ development. It might be why other compilers are having issues as per comments above.

So first build has many errors like:
```
FastLED\platforms/esp/32/fastspi_esp32.h:52:1: error: 'SPIClass' does not name a type
 SPIClass ledSPI(FASTLED_ESP32_SPI_BUS);
```

If I add `#include <SPI.h>` above `FASTLED_NAMESPACE_BEGIN` in fastspi_esp32.h then the above errors go away but the link step has issues:

```
.pio\build\esp32thing\libfb0\FastLED\bitswap.cpp.o:(.bss.ledSPI+0x0): multiple definition of `ledSPI'
.pio\build\esp32thing\libfb0\FastLED\FastLED.cpp.o:(.bss.ledSPI+0x0): first defined here
.pio\build\esp32thing\libfb0\FastLED\colorpalettes.cpp.o:(.bss.ledSPI+0x0): multiple definition of `ledSPI'
.pio\build\esp32thing\libfb0\FastLED\FastLED.cpp.o:(.bss.ledSPI+0x0): first defined here
.pio\build\esp32thing\libfb0\FastLED\colorutils.cpp.o:(.bss.ledSPI+0x0): multiple definition of `ledSPI'
.pio\build\esp32thing\libfb0\FastLED\FastLED.cpp.o:(.bss.ledSPI+0x0): first defined here
.pio\build\esp32thing\libfb0\FastLED\hsv2rgb.cpp.o:(.bss.ledSPI+0x0): multiple definition of `ledSPI'
.pio\build\esp32thing\libfb0\FastLED\FastLED.cpp.o:(.bss.ledSPI+0x0): first defined here
.pio\build\esp32thing\libfb0\FastLED\lib8tion.cpp.o:(.bss.ledSPI+0x0): multiple definition of `ledSPI'
.pio\build\esp32thing\libfb0\FastLED\FastLED.cpp.o:(.bss.ledSPI+0x0): first defined here
.pio\build\esp32thing\libfb0\FastLED\noise.cpp.o:(.bss.ledSPI+0x0): multiple definition of `ledSPI'
.pio\build\esp32thing\libfb0\FastLED\FastLED.cpp.o:(.bss.ledSPI+0x0): first defined here
.pio\build\esp32thing\libfb0\FastLED\power_mgt.cpp.o:(.bss.ledSPI+0x0): multiple definition of `ledSPI'
.pio\build\esp32thing\libfb0\FastLED\FastLED.cpp.o:(.bss.ledSPI+0x0): first defined here
.pio\build\esp32thing\libfb0\FastLED\wiring.cpp.o:(.bss.ledSPI+0x0): multiple definition of `ledSPI'
.pio\build\esp32thing\libfb0\FastLED\FastLED.cpp.o:(.bss.ledSPI+0x0): first defined here
.pio\build\esp32thing\src\main.cpp.o:(.bss.ledSPI+0x0): multiple definition of `ledSPI'
.pio\build\esp32thing\libfb0\FastLED\FastLED.cpp.o:(.bss.ledSPI+0x0): first defined here
.pio\build\esp32thing\lib06c\libArdUtils.a(ard_led.cpp.o):(.bss.ledSPI+0x0): multiple definition of `ledSPI'
.pio\build\esp32thing\libfb0\FastLED\FastLED.cpp.o:(.bss.ledSPI+0x0): first defined here
```

I think `ledSPI` needs to be created once in a .cpp source file, and header can have

    extern SPIClass ledSPI;

ChisholmKyle:
It's extremely easy to use PlatformIO. I'm developing using SparkFun ESP32 "Thing" and the only thing needed to test your library fork is to use VSCode with PlatformIO extension installed and this platformio.ini file: 

```
[platformio]
default_envs =
  esp32thing

[env:esp32thing]
framework = arduino
platform = espressif32
board = esp32thing

lib_deps =
  SPI
  https://github.com/ntwallace/FastLED
```

check out platformIO available boards if you're using another one beside the SparkFun "Thing"

skandragon:
Two questions:  One, does this work well with multiple SPI pins, or does it gracefully handle if the ESP32 runs out of SPI pin mappings?

trandi:
Just wanted to say THANK YOU @ntwallace , you've just saved my life with this change !
I've just lost 2-3 hours going crazy about why neither FastLED nor NeoPixelBus libraries were working with my modest, WS2801 based, 30 LED strip (random colours appearing on some of the LEDs) and it turns out it was the timing on ESP32 when using software SPI.
I've just manually applied this diff, using the HSPI bus, and everything seems fine !!!

samguyer:
@ntwallace I'm looking at this PR, and I'd like to merge it as soon as possible. Two questions: first, do we need to add instructions on how to set up the SPI library? Can we guard the code that won't compile with #ifdefs? Second, is there any issue with the limited number of SPI busses?

bmagistro:
@ntwallace question on the SPI bus, if I am reading the ESP32 datasheet right, it looks like you can use any GPIO pin for the SPI unless you need parallel SPI and then it has some pin restrictions.  I'm still getting things started for my project but any thoughts on if the general purpose SPI is sufficient for driving the APA102s, if so how could someone override the predefined pins?

https://www.espressif.com/sites/default/files/documentation/esp32_datasheet_en.pdf

samguyer:
@ntwallace We finally merged your SPI work! Thank you so much for this contribution.

Benik3:
@stickperson add `#include <SPI.h>` before fastled include :)
I will leave it here for other people which search for a solution...

gitTinker:
Maybe this will be convenient for others; an example of using the above with FastLed
MPU: [ESP32-PICO](https://www.amazon.com/gp/product/B08VGST8LJ)
Arduino IDE: 1.8.19
[FastLed](https://github.com/FastLED/FastLED): 3.5.0

`
#define HSPI_MISO   27
#define HSPI_MOSI   26    #this is the only IO pin used in this example (master out, slave in)
#define HSPI_SCLK   25
#define HSPI_SS     32
#define FASTLED_ALL_PINS_HARDWARE_SPI
#define FASTLED_ESP32_SPI_BUS HSPI

#include <SPI.h>
a#include "FastLED.h"

#define Neopixel_PIN 26
#define NUM_LEDS 50   
CRGB leds[NUM_LEDS];

void setup() {
FastLED.addLeds<WS2811, Neopixel_PIN, RGB>(leds, NUM_LEDS);
}

void loop() { }
`

criptych:
@Benik3 Should note that it must be added in a particular file (see @ChisholmKyle's comment earlier); adding it to your own code does bupkis. And even then I'm getting the linker error he mentioned as well.  Unfortunately something must have changed between this PR and 3.5.0 because now even his additional changes aren't working for me.

@gitTinker WS2811s don't use SPI, though.

100ideas:
@gitTinker, I think the pins you are using for SPI do not interface with the higher-speed `IO_MUX` bus. Yes you can define non-default SPI pins, but if any of them DO NOT connect with the `IO_MUX` bus then the entire SPI channel will run via the `GPIO matrix` which puts more load on the CPU core ("software mode"). You can get better performance by choosing a set of pins for your SPI connection that are all supported by `IO_MUX` (maybe worthwhile if you are driving a lot of LEDs or reacting to interrupts).

> GPIO Matrix and IO_MUX
> 
> Most of ESP32’s peripheral signals have a direct connection to their dedicated IO_MUX pins. However, the signals can also be routed to any other available pins using the less direct > GPIO matrix. If at least one signal is routed through the GPIO matrix, then all signals will be routed through it.
> 
> The GPIO matrix introduces flexibility of routing but also brings the following disadvantages:
> - Increases the input delay of the MISO signal, which makes MISO setup time violations more likely. If SPI needs to operate at high speeds, use dedicated IO_MUX pins.
> - Allows signals with clock frequencies only up to 40 MHz, as opposed to 80 MHz if IO_MUX pins are used.
>
> source: https://docs.espressif.com/projects/esp-idf/en/latest/esp32/api-reference/peripherals/spi_master.html#gpio-matrix-and-io-mux

It's kinda tedious to figure out what pins are supported by IO_MUX, what their names are, and how accessable they are on each particular esp32 board. This header file is helpful [`esp-idf/components/soc/esp32/include/soc/spi_pins.h`](https://github.com/espressif/esp-idf/blob/master/components/soc/esp32/include/soc/spi_pins.h) (notice legacy naming conventions):

```C
#pragma once

#define SPI_FUNC_NUM            1
#define SPI_IOMUX_PIN_NUM_MISO  7
#define SPI_IOMUX_PIN_NUM_MOSI  8
#define SPI_IOMUX_PIN_NUM_CLK   6
#define SPI_IOMUX_PIN_NUM_CS    11
#define SPI_IOMUX_PIN_NUM_WP    10
#define SPI_IOMUX_PIN_NUM_HD    9

//For D2WD and PICO-D4 chip
#define SPI_D2WD_PIN_NUM_MISO  17
#define SPI_D2WD_PIN_NUM_MOSI  8
#define SPI_D2WD_PIN_NUM_CLK   6
#define SPI_D2WD_PIN_NUM_CS    16
#define SPI_D2WD_PIN_NUM_WP    7
#define SPI_D2WD_PIN_NUM_HD    11

#define SPI2_FUNC_NUM           HSPI_FUNC_NUM
#define SPI2_IOMUX_PIN_NUM_MISO HSPI_IOMUX_PIN_NUM_MISO
#define SPI2_IOMUX_PIN_NUM_MOSI HSPI_IOMUX_PIN_NUM_MOSI
#define SPI2_IOMUX_PIN_NUM_CLK  HSPI_IOMUX_PIN_NUM_CLK
#define SPI2_IOMUX_PIN_NUM_CS   HSPI_IOMUX_PIN_NUM_CS
#define SPI2_IOMUX_PIN_NUM_WP   HSPI_IOMUX_PIN_NUM_WP
#define SPI2_IOMUX_PIN_NUM_HD   HSPI_IOMUX_PIN_NUM_HD

#define SPI3_FUNC_NUM           VSPI_FUNC_NUM
#define SPI3_IOMUX_PIN_NUM_MISO VSPI_IOMUX_PIN_NUM_MISO
#define SPI3_IOMUX_PIN_NUM_MOSI VSPI_IOMUX_PIN_NUM_MOSI
#define SPI3_IOMUX_PIN_NUM_CLK  VSPI_IOMUX_PIN_NUM_CLK
#define SPI3_IOMUX_PIN_NUM_CS   VSPI_IOMUX_PIN_NUM_CS
#define SPI3_IOMUX_PIN_NUM_WP   VSPI_IOMUX_PIN_NUM_WP
#define SPI3_IOMUX_PIN_NUM_HD   VSPI_IOMUX_PIN_NUM_HD

//Following Macros are deprecated. Please use the Macros above
#define HSPI_FUNC_NUM           1
#define HSPI_IOMUX_PIN_NUM_MISO 12
#define HSPI_IOMUX_PIN_NUM_MOSI 13
#define HSPI_IOMUX_PIN_NUM_CLK  14
#define HSPI_IOMUX_PIN_NUM_CS   15
#define HSPI_IOMUX_PIN_NUM_WP   2
#define HSPI_IOMUX_PIN_NUM_HD   4

#define VSPI_FUNC_NUM           1
#define VSPI_IOMUX_PIN_NUM_MISO 19
#define VSPI_IOMUX_PIN_NUM_MOSI 23
#define VSPI_IOMUX_PIN_NUM_CLK  18
#define VSPI_IOMUX_PIN_NUM_CS   5
#define VSPI_IOMUX_PIN_NUM_WP   22
#define VSPI_IOMUX_PIN_NUM_HD   21
```

So looking at the espressif docs for [ESP-PICO-KIT-1](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/hw-reference/esp32/get-started-pico-kit-1.html#header-j2), [ESP-PICO-KIT-V4](https://docs.espressif.com/projects/esp-idf/en/latest/esp32/hw-reference/esp32/get-started-pico-kit.html#header-j2) pin description list, it indicates that

| NAME | Function |
| ------ | ---------- |
| IO14 | ADC2_CH6, TOUCH6, RTC_GPIO16, MTMS, **HSPICLK**, HS2_CLK, SD_CLK, EMAC_TXD2 |
| IO12 | ADC2_CH5, TOUCH5, RTC_GPIO15, MTDI (See 3), **HSPIQ**, HS2_DATA2, SD_DATA2, EMAC_TXD3 |
| IO13 | ADC2_CH4, TOUCH4, RTC_GPIO14, MTCK, **HSPID**, HS2_DATA3, SD_DATA3, EMAC_RX_ER | 
| IO15 | ADC2_CH3, TOUCH3, RTC_GPIO13, MTDO, **HSPICS0**, HS2_CMD, SD_CMD, EMAC_RXD3 |
| ...  |  ... |

the venerable @kolban asked what these unusual Qs and Ds indicate on an esp32 forum (see bottom) and the response was: 

```
"GPIO pin for spi_d (=MOSI), spi_q (=MISO) signal". 
```

So then, if you wanted to try IO_MUX enabled SPI definitions, you might try:

```C
#define HSPI_MISO   12
#define HSPI_MOSI   13    #this is the only IO pin used in this example (master out, slave in) 
#define HSPI_SCLK   14
#define HSPI_SS     15
```

I could have this backwards, I've just been doing a lot of research into how to run dual `IO_MUX` enabled SPI channels in parallel. So far I think this is possible by using the designated `HSPI/SPI3` and `VSPI/SPI3` GPIO groups.

One thing I'm not sure about is what happens if `SPI2` and `SPI3` channels are run on the same cpu core? I think I read in the espressif docs that it can work, but each SPI channel needs to be run in a separate RTOS task...

---

**[[Answered] SPI: Meaning of some of the pins for spi_bus_config_t](https://esp32.com/viewtopic.php?f=13&t=918)**
Regarding espressif's confusing new SPI names (`VSPIQ`, `HSPID`, `HSPIWP` etc):

> kolban » Tue Jan 10, 2017 5:24 am:
>
> Studying the SPI master driver, I find that we must configure a data structure called "spi_bus_config_t". Reading the documentation on it found here:
> 
> [http://esp-idf.readthedocs.io/en/latest ... s_config_t](http://esp-idf.readthedocs.io/en/latest/api/spi_master.html#_CPPv216spi_bus_config_t)
> 
> I find 5 pins need be configured:
> ```
> spid_io_num
> spiq_io_num
> spiclk_io_num
> spiwp_io_num
> spihd_io_num
> ```
> Here comes the list of questions related to these names:
> 
> 1) What does "spi_d" mean? I understand the SPI concept of "MOSI" ... but the concept of "d" means nothing to me.
> 2) What does "spi_q" mean? I understand the SPI concept of "MISO" ... but the concept of "q" means nothing to me.
> 3) What does "spi_wp" mean? I can't relate that to anything I am familiar with in the concept of SPI.
> 4) What does "spi_hd" mean? I can't relate that to anything I am familiar with in the concept of SPI.
>
> .....................
> 
> ESP_Sprite » Tue Jan 10, 2017 6:21 am:
>
> The SPI flash peripheral actually was initially created to interface with SPI flash chips, which have a d, q, clk, wp and hd pin. I can imagine the SPI port pins are somewhat weird if you aren't familiar with that.
> 
> For MISO and MOSI, it is explained in the page you quoted:
> ```
> int spid_io_num
> 
> GPIO pin for spi_d (=MOSI)signal, or -1 if not used.
> 
> int spiq_io_num
> 
> GPIO pin for spi_q (=MISO) signal, or -1 if not used.
> ```
> spi_hd is the hold pin pf the flash chip, while spi_wp is for write protect. These pins aren't really used in the 1-bit SPI mode you'd normally use, but the 4-bit mode uses them as > data lines.
> 
> I agree that the documentation can be slightly better wrt explaining this.

SCarter858:
Has anyone used this ESP32 Hardware SPI support successfully for effective transfer rates higher than about 5MHz?  It looks to me like it's using the DMA channel for only single byte transfers - very inefficient.  

I'm using a simple test program that just continually calls FastLED.show() to rewrite a string of 128 APA102's (so each "show" should write about 4k bits to the string).  On an ESP32S3 and FastLED 3.6.0, I'm using pins 14 & 13, the correct hw pins for the HSPI.   I get the #pragma message that ESP32 hardware support was added, and indeed, I see the stable 20MHz default clock.  But there's a microsecond pause between each 8 SPI clocks.  

The Espressif docs talk about how important it is to set up the DMA channel correctly for the SPI driver, as each transfer setup takes a uSec.  Looks to me like this code sends each individual byte for its own DMA transfer rather than sending the whole 4k at once?

Am I missing something?  Seems a shame to not be able to use the 40-80MHz rates this chip is capable of for the newer faster LED strips.  I see this driver calling ledSPI.transfer(byte) continually.  Is all FastLED transfer byte-wise without digging down into the fundamentals?

PR #1048, opened 2020-07-10
non linear power scaling

L0laapk3:
I was having issues with power limiting: when I did my full brightness current test on 88x sk6805-1515, I measured R: .55A, G: .33A, B: .56A, R+G+B: 1.46A. I am trying to comply to USB spec so a current limit of 0.5A is needed. After changing the mw constants for each color and setting the appropriate current limit, I obtained results of R: .48A, G: .33A, B: .48A. All good. However, when I use mixed colors such as white, the current draw went up to as much as .62A.
There seems to be non-linear current scaling coming into play, this PR attempts to fix this. With the exponent I've implemented of 0.77, the white current draw nicely sits at .48A in the same setup without affecting the other colors.

I've attempted to implement non linear power scaling in a way that shouldn't impact performance too much. Additional testing is needed as I only measured power on sk6805-1515's.

### DIFF
--- a/FastLED.h
+++ b/FastLED.h
@@ -490,7 +490,7 @@ public:
 
 	/// Set the maximum power to be used, given in milliwatts
 	/// @param milliwatts - the max power draw desired, in milliwatts
-	inline void setMaxPowerInMilliWatts(uint32_t milliwatts) { m_pPowerFunc = &calculate_max_brightness_for_power_mW; m_nPowerData = milliwatts; }
+	inline void setMaxPowerInMilliWatts(uint32_t milliwatts) { setupPowerScaleTable(); m_pPowerFunc = &calculate_max_brightness_for_power_mW; m_nPowerData = milliwatts; }
 
 	/// Update all our controllers with the current led colors, using the passed in brightness
 	/// @param scale temporarily override the scale
--- a/power_mgt.cpp
+++ b/power_mgt.cpp
@@ -23,17 +23,17 @@ FASTLED_NAMESPACE_BEGIN
 // be compile-time constants.
 
 constexpr float POWER_EXPONENENT = 0.87; // 50% pwm will result in 0.5^0.87 = 54.7% of power usage
-//constexpr uint8_t gRed_mW   = 16 * 5; // 16mA @ 5v = 80mW
-//constexpr uint8_t gGreen_mW = 11 * 5; // 11mA @ 5v = 55mW
-//constexpr uint8_t gBlue_mW  = 15 * 5; // 15mA @ 5v = 75mW
-//constexpr uint8_t gDark_mW  = 1 * 5;  // 1mA @ 5v = 75mW
+constexpr uint8_t gRed_mW   = 16 * 5; // 16mA @ 5v = 80mW
+constexpr uint8_t gGreen_mW = 11 * 5; // 11mA @ 5v = 55mW
+constexpr uint8_t gBlue_mW  = 15 * 5; // 15mA @ 5v = 75mW
+constexpr uint8_t gDark_mW  = 1 * 5;  // 1mA @ 5v = 75mW
 
 // Alternate measurements for sk6805-1515
 // constexpr float POWER_EXPONENENT = 0.87;
- constexpr uint8_t gRed_mW   = 25;
- constexpr uint8_t gGreen_mW = 18;
- constexpr uint8_t gBlue_mW  = 24;
- constexpr uint8_t gDark_mW  = 4;
+// constexpr uint8_t gRed_mW   = 25;
+// constexpr uint8_t gGreen_mW = 18;
+// constexpr uint8_t gBlue_mW  = 24;
+// constexpr uint8_t gDark_mW  = 4;
 
 // Alternate calibration by RAtkins via pre-PSU wattage measurments;
 // these are all probably about 20%-25% too high due to PSU heat losses,
--- a/power_mgt.cpp
+++ b/power_mgt.cpp
@@ -67,21 +67,21 @@ static uint8_t gMaxPowerIndicatorLEDPinNumber = 0; // default = Arduino onboard 
 
 // if defined, use linear power scaling.
 #ifdef FASTLED_POWER_SCALING_FAST
-inline uint8_t getPowerValue(uint8_t brightness) {
+inline constexpr uint8_t getPowerValue(uint8_t brightness) {
     return brightness;
 }
-inline uint8_t getReversePowerValue(uint8_t brightness_scaled) {
+inline constexpr uint8_t getReversePowerValue(uint8_t brightness_scaled) {
     return brightness_scaled;
 }
 void setupPowerScaleTable() {}
 #else
 #define calcPowerValue(brightness) static_cast<uint8_t>(pow(brightness, POWER_EXPONENENT) * pow(256, 1 - POWER_EXPONENENT))
-inline uint8_t getReversePowerValue(uint8_t brightness_scaled) {
+inline constexpr uint8_t getReversePowerValue(uint8_t brightness_scaled) {
     return static_cast<uint8_t>(pow(brightness_scaled / pow(256, 1 - POWER_EXPONENENT), 1.f/POWER_EXPONENENT));
 }
 // if defined, compute power value each time.
 #ifdef FASTLED_POWER_SCALING_COMPUTE
-inline uint8_t getPowerValue(uint8_t brightness) {
+inline constexpr uint8_t getPowerValue(uint8_t brightness) {
     return calcPowerValue(brightness);
 }
 void setupPowerScaleTable() {}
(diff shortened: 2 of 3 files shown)


L0laapk3:
As a sidenote, it would be nice to have the power draw definions somewhere tied to the led type. (maybe in `CLEDController`?) Anyways thats out of scope for this PR.

5chmidti:
I don't know about power scaling, but here some remarks:

[don't pass literals by const &](https://godbolt.org/z/s7P165)

For people that rely on power scaling this could be really great, but there is one thing you overlooked.
When using controllers with little ram, your array chips off 256 bytes of memory. For an Arduino Nano, that only has 2kB, this would be 1/8 of all memory.

I think making this opt in / configurable would be better. Adding a define somewhere (in power_mgt.h or fastled_config.h I would guess) that can switch between implementations, with the way it is now as the default.

Maybe like this?: 
```cpp
#ifdef FASTLED_POWER_VALUE_FAST
constexpr uint8_t getTargetBrightness(uint8_t target_brightness) { return target_brightness; }
constexpr uint8_t getPowerValue(uint8_t brightness) { return brightness; }
constexpr uint8_t getResultBrightness(uint8_t recommended_brightness) { return recommended_brightness; }
constexpr uint8_t getPowerVal(uint8_t brightness) { return brightness; }

#else
constexpr float POWER_EXPONENENT = 0.77; // 50% pwm will result in 0.5^0.75 = 58.6% of power usage
constexpr uint8_t getPowerValue(uint8_t brightness) { return pow(brightness, POWER_EXPONENENT) * pow(256, 1 - POWER_EXPONENENT); }
constexpr uint8_t getTargetBrightness(uint8_t target_brightness) { return getPowerValue(target_brightness); }
constexpr uint8_t getResultBrightness(uint8_t brightness) { return pow(brightness / pow(256, 1 - POWER_EXPONENENT), 1.f / POWER_EXPONENENT); }
#ifdef FASTLED_POWER_VALUE_MEMORY
// array defs
constexpr uint8_t getPowerVal(uint8_t brightness) { return POWERTABLE[brightness]; }
#elif defined(FASTLED_POWER_VALUE_COMPUTE)
constexpr uint8_t getPowerVal(uint8_t brightness){ return getPowerValue(brightness); }
#endif
#endif

// substituting the calls in the function bodies where necessary
```

One could argue that making the `FASTLED_POWER_VALUE_COMPUTE` version the default could be fine too, as the amount of times one updates the power management is probably once, at setup.



And in line 26 you use 0.77 but write in your comment 0.75, that should be  0.77.

Minor stuff/maybe's: All of those functions could be marked `always_inline`, make `POWER_EXPONENENT` static and (this part is not yours) make `gRed_mW, gGreen_mW, gBlue_mW` `constexpr` instead of just `const`.

L0laapk3:
Thanks for the feedback. I'll make some adjustments.

To my understanding `calculate_unscaled_power_mW()` does get called on every `show()`, which is why I made an attempt to optimise it.

L0laapk3:
I've added options to `#DEFINE FASTLED_POWER_SCALING_FAST` and `#DEFINE FASTLED_POWER_SCALING_COMPUTE`.

I've tweaked the parameters a bit. Currently, I'm happy to say that when I did a full RGB sweep, the current stayed exactly to 0.5A to within tolerance of my meter (+-0.01A). (on the portions where the power limiting is active ofcourse). For future reference, the parameters I used for my 88x sk6805-1515's are:
```
constexpr float POWER_EXPONENENT = 0.87;
constexpr uint8_t gRed_mW   = 25;
constexpr uint8_t gGreen_mW = 18;
constexpr uint8_t gBlue_mW  = 24;
constexpr uint8_t gDark_mW  = 4;
```

5chmidti:
At now ln. 220 it should return `return getReversePowerValue(target_brightness);`, otherwise looks good to me (didn't measure amps though).

`show()` doesn't call  `calculate_unscaled_power_mW()` or any of the power_mgt functions as far as I can see.

L0laapk3:
Line 220 simply returns the original target_brightness, not the one that was scaled according to power consumption.

It seems like `show()` calls it at https://github.com/FastLED/FastLED/blob/5608ecf7/FastLED.cpp#L52 .

5chmidti:
I just mentioned the return, as you had something else there before. 

Oh, you're right, I didn't see that. That means that `FASTLED_POWER_SCALING_FAST` should be the default to make sure the same performance is there as before.
I noticed there isn't any of the power scaling macros defined anywhere (just the conditions), so `#define FASTLED_POWER_SCALING_FAST` would have to be somewhere.

Now the rest is up to the maintainers or people that can do more measurements I guess

L0laapk3:
I disagree, since the FAST option will likely result in power draw exceeding the maximum power draw it is in my opinion not a good default option. In my opinion the array based approach is the best, however in the case the extra used memory is a problem, thats when users can define `#FASTLED_POWER_SCALING_COMPUTE` or `#FASTLED_POWER_SCALING_FAST`.

However, its probably a good idea to only allocate the memory when the user actually uses the power limitation features.

5chmidti:
I was wrong about something, the array won't blow up ram usage, as it is just values in a symbol table inside the compiled binary. Should've thought more about where the array is stored. 
Because the array version would be more accurate it would be better as a default, sorry for the inconvenience :/

Two important things though. I tested this with an Arduino nano I had lying around, and it seems like the compiler for the nano has no array file to include. So you should make those arrays c-style arrays, i.e. `constexpr uint8_t POWERTABLE[256]`.
Also there are narrowing conversions happening with the array initialization, casting the result to `uint8_t` by, for example, changing the `calcPowerValue` macro to
`#define calcPowerValue(brightness) static_cast<uint8_t>(pow(brightness, POWER_EXPONENENT) * pow(256, 1 - POWER_EXPONENENT))` does the trick.

L0laapk3:
Memory does seem to be a valid concern as the table will live on the heap. It seems like a good idea to only allocate the table if the power functions are actually used. See my last commit.

kriegsman:
Hi all- thanks for the help with this issue; the FastLED power management feature is great when it works right, so I always appreciate help tuning it up a bit.

A couple of questions and thoughts come to mind, and I'd appreciate hearing your thoughts on them.

First, I'm very curious where the non-linearity is coming from.  Did I understand that that you were testing on 88 pixels of sk6805-1515?  If so... were you applying power just from one end of the string?  Or were you injecting power periodically, say, every 30 pixels or so?    In your test rig, what's the measured voltage at the 'far end' of the pixel array? There's definitely a voltage drop along the string as the power consumption goes up, and that leads to nonlinearity of current draw, but often in that case the right answer is to make the power distribution more even along the strip.

I'd also be curious to hear if you find the same nonlinearity if you only light up, say, ten pixels at the near-end of the strip.  That would basically take the voltage sag -along-the-string out of the equation.

So I'd like to investigate that a bit more, if you would be willing -- the source of the nonlinearity.  In the past, I've found the fully-powered LED pixels tend to be pretty linear, which is why I'm asking all these questions about this test setup.

Once we figure out where the non-linearity is coming from, then we can look at this code.  Overall I like the approach, but I'm also a bit wary of using "float" values and pow(...), which will add to code size and reduce speed a little bit.  I wonder if there's a fixed point option... but we can get to that discussion once we understand the power draw situation.

Thanks again for helping to investigate this!

kriegsman:
One other thought just came up now as I was talking with @samguyer, too:  he suggested testing to see if the power consumption increased linearly with the number of all-white LEDs.  E.g. light up just 1 LED pure 0xFFFFFF white, and measure power draw.  Then try 10 and measure current, then try 20 and measure current, 30, 40, 50, 60, 70, and 80, measuring current at each step, and see if the overall current draw goes up linearly there, or if something else is going on.

Might be interesting data, and might help understand what's going on.

Alexandr180:
I broke my head, but I don't understand how to get information about the current consumption, for example, (Serial.print) with a given frequency. "calculate_max_brightness_for_power_vmA" how to work with it? 
The accuracy suits me, the current limitation works correctly, and the network Watt meter shows about 10% deviation.

zackees:
I like this change and will keep this open. The power function is starting to rot as APA102HD is no longer RGB8 but RGB8 + 5brightness. APA102HD mode does gamma correction in the chip itself and this isn't reflected in the power feature. Neither is RGBW with the extra component. This is a big job to fix to make it work everywhere. Also I've seen WS2812 that is ultra low power so the user code must be able to set an adjustment because not all led's are regulated the same.

zackees:
I couldn't update this PR in place because its head branch is `L0laapk3:powertweaks`, not a branch in the main FastLED repo.

I rebased and re-implemented the change against current `master` on a new branch and opened a replacement PR here:
- #2362 https://github.com/FastLED/FastLED/pull/2362

That PR carries the modernized implementation and current focused test coverage.

zackees:
this is in

PR #1063, opened 2020-08-02
added support for LGT8F arduino nano and pro clones

jayzakk:
As more and more Arduino nano/pro clones with LGT8F328-chips show up, they should get supported.
It's some more updated version of the Atmega328, so many ops have lower cycle count. This PR adds support for the current timing up to max (32MHz) chip speed.


leifclaesson:
Tested and works perfectly for me with a WS2813 strip on a WAVGAT LGT8F board at 16 MHz, while the main unmodified FastLED flickers severely due to the timing differences. I for one would love to see this merged into the main line.

kriegsman:

PR #1079, opened 2020-09-12
implemented the option to place the matrix vertically

c0decaps:
In the XYMatrix example one now may chose whether the matrix is placed vertically by rotating it 90 degrees clockwise with the  boolean kMatrixVertical option.


kriegsman:

PR #1089, opened 2020-09-27
Update fastpin_apollo3.h

skyfly200:
Add support for Artemis Dev Kit
Still only supporting the Artemis Arduino core v1.2.1


kriegsman:

PR #1091, opened 2020-09-28
Add support for non-P ATMega1284 device

wirehead:
FastLED already supports the ATMega1284P, but I happen to only have plain ATMega1284 parts in my geekroom.

The only difference between an ATMega1284P and a regular ATMega1284 is that the P is the picopower version that uses less power, so all one needs to do is add one more define and it works just fine.


kriegsman:
Thanks for the contribution!

PR #1096, opened 2020-10-12
add qt py support!

ladyada:
a new samd21 variant

### DIFF
--- a/platforms/arm/d21/fastpin_arm_d21.h
+++ b/platforms/arm/d21/fastpin_arm_d21.h
@@ -215,6 +215,19 @@ _FL_DEFPIN( 3, 7, 0); _FL_DEFPIN( 4, 6, 0); _FL_DEFPIN( 7, 0, 0); _FL_DEFPIN( 8,
 
 #define HAS_HARDWARE_PIN_SUPPORT 1
 
+#elif defined(ADAFRUIT_QTPY_M0)
+
+#define MAX_PIN 10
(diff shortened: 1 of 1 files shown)


marmilicious:
Adding more info for anyone interested.
Product page:  https://www.adafruit.com/product/4600
ATSAMD21 Datasheet:  https://cdn-learn.adafruit.com/assets/assets/000/044/363/original/samd21.pdf?1501106093

marmilicious:
@ladyada Yes, in little bits here and there.  Loosing Dan Garcia was a pretty big blow to development and updates.  It's still slow going.

ladyada:
im sorry to hear that :( would it be easier if we released a FastLED Adafruit Fork so we're not waiting on y'all?

marmilicious:
@ladyada  That would be a good question for Mark @kriegsman

kriegsman:
@samguyer and I (and @uberjay) are getting back into the groove a bit more now.  We've got weekly development sessions on our calendars, and we're hoping to also start including more contributors.

The idea of an Adafruit fork is a good one to think about, but let's see if we can keep up a better pace for now with just the original version.  We'll do what we can to prioritize support for the Adafruit boards since so many people use (and love) them.

ladyada:
thank ya! one thing that may help improve contributions is a CI system. i can help implement the one we use for adafruit - it can be easily adapted to support a vast number of boards to make sure no compilation errors sneak in :)

kriegsman:
THAT would be GREAT... you know better than anyone else really what a huge range of boards are out there, and how easy it is to have an oops.

PR #1108, opened 2020-11-02
Major updates to ESP32 support

samguyer:
This update includes significant updates to the ESP32 support based on the RMT peripheral. All methods/functions used by the interrupt handler have been refactored to a non-template class so that they can be placed in IRAM. This change allows show() to continue to operate even when flash memory operations are happening. The other big change is that we now use two RMT memory blocks per LED strip, which allows the device to go longer without generating an interrupt for data refills. This change seems to improve reliability in the presence of WiFi and other high-priority interrupt tasks. Thanks to @bbulkow for this invaluable work.


samguyer:
The other major change is moving the code into a src directory, which allows us to have cpp files wherever we need them, and a library.json file, which allows FastLED to work more smoothly with platform.io

porkyneal:
@samguyer I just wanted to check to see if this solves a problem of mine, but I can't compile due to line 400 of clockless_rmt_esp32.cpp. 

```mBufferSize size = size_in_bytes * 8 * 4;``` maybe it should be ```mBufferSize = size_in_bytes * 8 * 4;```?

Thanks for your efforts, testing now..

**Comment by samguyer** [comment]
@porkyneal Sorry about that! I just forgot to push my changes. Should be good now.

**Comment by elliottkember** [comment]
@samguyer Should the `component.mk` file now be changed?

```
COMPONENT_ADD_INCLUDEDIRS := .
```

to

```
COMPONENT_ADD_INCLUDEDIRS := src
```

Asking because I updated the FastLED lib and hit a bunch of `fatal error: FastLED.h: No such file or directory`

samguyer:
@elliottkember Oh, you're probably right. Let me take a look. Did your change to the component.mk file fix it?

elliottkember:
@samguyer it did for me! 

I wonder if I can make a test case for it. Has anybody considered using GitHub actions to smoke-test compilation? I’d be happy to look into it.

Issue #1118, opened 2020-11-24
ESP32 (ld): undefined reference to `ESP32RMTController::ESP32RMTController


elliottkember:
Updated FastLED to get the new updates! Very excited about the new ESP32 functionality as that's the chipset I use.

I made a change to the `component.mk` as explained in https://github.com/FastLED/FastLED/pull/1108#issuecomment-732538248 - including the src directory. However, I'm now hitting this problem in the linking step:

```
Generating esp32.project.ld
LD /Users/elliott/Sites/soulmate/soulmate-builder-server/output/sketch/build/soulmate.elf
/Users/elliott/Sites/soulmate/soulmate-builder-server/output/sketch/build/main/libmain.a(main.o):(.literal._ZN15SoulmateLibrary5setupEv[SoulmateLibrary::setup()]+0x5c): undefined reference to `ESP32RMTController::ESP32RMTController(int, int, int, int)'
/Users/elliott/Sites/soulmate/soulmate-builder-server/output/sketch/build/main/libmain.a(main.o):(.literal._ZN19ClocklessControllerILi32ELi60ELi150ELi90EL6EOrder66ELi0ELb0ELi5EE13loadPixelDataER15PixelControllerILS0_66ELi1ELj4294967295EE[ClocklessController<32, 60, 150, 90, (EOrder)66, 0, false, 5>::loadPixelData(PixelController<(EOrder)66, 1, 4294967295u>&)]+0x0): undefined reference to `ESP32RMTController::getPixelBuffer(int)'
/Users/elliott/Sites/soulmate/soulmate-builder-server/output/sketch/build/main/libmain.a(main.o):(.literal._ZN19ClocklessControllerILi32ELi60ELi150ELi90EL6EOrder66ELi0ELb0ELi5EE10showPixelsER15PixelControllerILS0_66ELi1ELj4294967295EE[ClocklessController<32, 60, 150, 90, (EOrder)66, 0, false, 5>::showPixels(PixelController<(EOrder)66, 1, 4294967295u>&)]+0x0): undefined reference to `ESP32RMTController::showPixels()'
/Users/elliott/Sites/soulmate/soulmate-builder-server/output/sketch/build/main/libmain.a(main.o): In function `CLEDController::CLEDController()':
/Users/elliott/Sites/soulmate/soulmate-builder-server/soulmate-core/./SoulmateMatrix.h:130: undefined reference to `ESP32RMTController::ESP32RMTController(int, int, int, int)'
[...]

PR #1119, opened 2020-11-24
Update component.mk to work with ESP-IDF 3.3.x

elliottkember:
I found, when updating FastLED to the latest `src` directory change, that I was getting a lot of errors about missing files!

I think this is due to the new `src` directory, which I think is a very good idea and cleans up the repository a lot. Seems as though `component.mk`'s list of paths is not searched recursively for files.

I also had to add `src/platforms/esp32` to the paths for the new clockless files to be found, otherwise I had a lot of these errors in the linker:

```
libmain.a(main.o):(.literal._ZN15SoulmateLibrary15lightPercentageEf[SoulmateLibrary::lightPercentage(float)]+0xc): undefined reference to `FastLED'
libmain.a(main.o):(.literal._ZN15SoulmateLibrary15lightPercentageEf[SoulmateLibrary::lightPercentage(float)]+0x10): undefined reference to `fill_solid(CRGB*, int, CRGB const&)'
libmain.a(main.o):(.literal._ZN15SoulmateLibrary15lightPercentageEf[SoulmateLibrary::lightPercentage(float)]+0x18): undefined reference to `CFastLED::show(unsigned char)'
libmain.a(main.o):(.literal._ZN15SoulmateLibrary18playCurrentRoutineEv[SoulmateLibrary::playCurrentRoutine()]+0x0): undefined reference to `hsv2rgb_rainbow(CHSV const&, CRGB&)'
libmain.a(main.o):(.literal._ZN15SoulmateLibrary10showPixelsEv[SoulmateLibrary::showPixels()]+0x1c): undefined reference to `blend(CRGB const&, CRGB const&, unsigned char)'
libmain.a(main.o):(.literal._ZN15SoulmateLibrary5setupEv[SoulmateLibrary::setup()]+0x24): undefined reference to `FastLED'
libmain.a(main.o):(.literal._ZN15SoulmateLibrary5setupEv[SoulmateLibrary::setup()]+0x28): undefined reference to `calculate_max_brightness_for_power_mW(unsig
```

I found also that I had to add `COMPONENT_SRCDIRS` for this to work. Removing either of these lines causes compilation to fail. 

If there is a simpler fix, I'd love to know what it is, but this made everything work for me!

### DIFF
--- a/component.mk
+++ b/component.mk
@@ -1,1 +1,3 @@
-COMPONENT_ADD_INCLUDEDIRS := .
+COMPONENT_ADD_INCLUDEDIRS := ./src src/platforms/esp/32
+COMPONENT_SRCDIRS := ./src src/platforms/esp/32
+
--- a/component.mk
+++ b/component.mk
@@ -1,3 +1,2 @@
 COMPONENT_ADD_INCLUDEDIRS := ./src src/platforms/esp/32
 COMPONENT_SRCDIRS := ./src src/platforms/esp/32
-


elliottkember:
This fixes  https://github.com/FastLED/FastLED/issues/1118

uberjay:
Thanks for the PR -- this looks like it'll take care of the esp32, but I'm worried it'll not play nice with the esp8266. (since you've got a couple esp/32 specific paths in there now.)

elliottkember:
@uberjay Good point. In fact it won't work for ESP8266 anyway because there is a `src/platforms/esp/8266` directory that we need to add.

I think the answer is to wrap the code in each file with `#ifdef ESP32` and `#ifdef ESP8266` blocks?

uberjay:
Looking at this a bit further, it seems like ESP-IDF only targets the esp32, so maybe there's no issue and your PR is correct?

elliottkember:
> Looking at this a bit further, it seems like ESP-IDF only targets the esp32, so maybe there's no issue and your PR is correct?

Yes I think you're right. It seems there is a plan to use ESP-IDF for ESP8266, but no set timeline for this.

uberjay:
Before the repo reorg, the include path was ".", because all of the files were in the top-level directory. But the esp platform support includes where still in `platforms/esp/{32,8266}`. How did it work before? Is it possible that you don't need to add the src/platforms/esp/32 path to the `COMPONENT_ADD_INCLUDEDIRS`, but _do_ need it for `COMPONENT_SRCDIRS`? Before @samguyer reworked the esp32 support, there were only headers in the two esp platform directories. Perhaps I should try building this with ESP-IDF myself.

uberjay:
Thanks for the PR! If it turns out this needs be tweaked some more, we can fix it up later.

PR #1127, opened 2020-11-30
Compilation fix when using I2S on ESP32

drzony:


kriegsman:
@samguyer could you take a look at this?  It looks like a quick fix, and if you like it, please just merge it in.  If it needs something additional/else we can also do that.

samguyer:
OK, this looks like a reasonable change, although I'm not entirely sure why it's necessary.

drzony:
clockless_rmt_esp32.cpp is always compiled on ESP32, if FASTLED_ESP32_I2S is defined then the clockless_rmt_esp32.h header is not included (see fastled_esp32.h) and weird things happen when compiling.

Issue #1128, opened 2020-12-01
Bug in ESP32RMTController::fillNext causing end of LEDs to misbehave for certain chain lengths


hazcat:
I have an ESP32 LOLIN board attached to a chain of ws2812b LEDs 
The following code causes LED 4 to not light and LED 3 to light as yellow instead of white.
If I checkout tag 3.3.3 then the issue does not occur.

    #include <Arduino.h>
    #include <FastLED.h>
    #define NUM_LEDS 4
    #define DATA_PIN 25
    CRGB leds[NUM_LEDS];
    
    void setup()
    {
        // Initialize LEDs
        FastLED.addLeds<WS2812B, DATA_PIN, GRB>(leds, NUM_LEDS);
        FastLED.clear();
        FastLED.setBrightness(128);
    }
    
    void loop()
    {
        CRGB colors[3] = {CRGB::White, CRGB::Green, CRGB::Black};
        static int colorCtr = 0;
        colorCtr = (colorCtr + 1) % 3;
        
        fill_solid(leds, NUM_LEDS, colors[colorCtr]);
        FastLED.show();
        delay(1000);
    }

In ESP32RMTController::fillNext this code is overwrites the pixel values that were just written when (mCur < mSize)

        } else {
            // -- No more data; signal to the RMT we are done
            for (uint32_t j = 0; j < 32; j++) {
                * mRMT_mem_ptr++ = 0;
            }
        }

If I change the code as follows my LEDs function as expected
        } else {
            // -- No more data; signal to the RMT we are done
            for (uint32_t j = 0; j < 32; j++) {
                *pItem++ = 0;
            }
        }

hazcat:
Should be fixed by #1129

PR #1129, opened 2020-12-01
Write End of Data to pItem as that prevents previously written data f…

hazcat:
…rom being stomped


hazcat:
Fixes #1128

samguyer:
@hazcat Thank you so much! That was such a dumb error on my part. :-)

PR #1140, opened 2020-12-15
Inline RMT functions needed to get code in IRAM

samguyer:
These changes should ensure that all code needed to push the bits out during FastLED.show() are in IRAM. Some of the ESP-IDF function we rely on do not have the IRAM_ATTR, so if they are executed during a flash memory operation, the program fails. The solution is to inline the code (basically, direct access to RMT registers) into our functions, which do have the IRAM attribute.


samguyer:
Merging my own pull request. I guess if anything goes wrong I'll have no one to blame!

Issue #1130, opened 2020-12-02
ESP32 GPIO0 gets unuseable


Jueff:
I discovered an issue in the current trunk, that GPIO0 pin of ESP32 can't be used as output anymore, after calling FastLed.show the first time.

I found out that it must have something to do with
void ESP32RMTController::init()
...
        rmt_tx.gpio_num = gpio_num_t(0);  // The particular pin will be assigned later
....
If I change to e.g.  gpio_num_t(22) GPIO 0 can be used as output again.
Seems the pin is set to another kind of operation, but isn't reset to default  when assigning the correct pin.

Jueff:
As a workaround I added 

`gpio_matrix_out(gpio_num_t(0), 0x100, false, false);`

in my code after the first FastLED.show call.

samguyer:
@Jueff Thanks for the issue report! I should be able to fix that problem in a more permanent way.

samguyer:
@Jueff I just pushed a fix for this issue, so I'm closing it. Thanks!

Jueff:
Thanks for the quick fix!

PR #1141, opened 2020-12-15
Fix for issue #1130: don't use pin 0 to configure the RMT device

samguyer:
The RMT init code only needs to be invoked one time, the first time show() is called, so I chose pin 0 as a stand-in for whatever pins would be used later. Unfortunately, the rmt_config code does configure the pin, so choosing pin 0 is not benign. Instead, I just pass in one of the pins that will actually be used (it doesn't matter which one).

In the long run I'd like to change the setup code to avoid this problem altogether.


samguyer:
Merge in this fix for #1130

PR #1144, opened 2020-12-20
Rolling back the bailout code

samguyer:


samguyer:
This is a safer way to bail out when the interrupt misses the deadline to refill the buffer.

PR #1149, opened 2020-12-23
Blur2d update for irregular matrix

80Stepko08:


kriegsman:
Hi and thank you for the contribution!  I'm curious what you're doing with an 'irregular matrix' -- and just how irregular is it?

80Stepko08:
> Hi and thank you for the contribution! I'm curious what you're doing with an 'irregular matrix' -- and just how irregular is it?

Im using 16x8 matrix, but with mapping from 16x16

80Stepko08:
> Im using 16x8 matrix, but with mapping from 16x16
Video of problem
https://youtu.be/hNOfrfGZsb8?t=16

kriegsman:
OK, thank you for the description!  I'm looking at this PR right now.

kriegsman:
Thank you -- nice PR, and allows for all sorts of different pixel layouts, because it calls XY(i,j) for each pixel.  There's some minor performance penalty for rectangular matrixes, but I think overall it's worth it here so that it works correctly in the general case.  In theory, we could also offer the original version as a special case for people who really really need the extra few clock cycles back -- but I think we'll leave that as a future piece of work, if it is needed then.

Thank you again for the PR, and the contribution to FastLED!

PR #1152, opened 2020-12-25
include optimized sin/cos code for AVR only when on this platform

jandelgado:
code optimized for AVR platform containing AVR assembly should only be included when compiling for AVR. 
Reason for change: I was experimenting to compile for a new platform, and got compilation errors.

### DIFF
--- a/src/lib8tion/trig8.h
+++ b/src/lib8tion/trig8.h
@@ -16,9 +16,6 @@
 
 #if defined(__AVR__)
 #define sin16 sin16_avr
-#else
-#define sin16 sin16_C
-#endif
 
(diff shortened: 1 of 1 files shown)


kriegsman:
Hi and thank you for reporting this issue!  Much appreciated.  
I think I'm going to try to make the same fix, but with a slightyl different arrangement of handling of the #defines and #ifdefs.

kriegsman:
Ha, nope! I take back what I just said.

I'm going to accept your PR exactly as written- and thank you very much!

jandelgado:
Excellent, thanks for merging. I made the change as compatible as possible. Firstly, I was also thinking on discarding the `#defines` but this could have broken existing code.

PR #1155, opened 2020-12-26
Build examples in CI using Github Actions

jandelgado:
Reason for Request: make sure code compiles when new PR's are submitted. This PR could later be used as a base for adding host-based unit tests for e.g. platform-independent code.

The examples are built for various different boards/platforms. Where necessary, examples were slightly changed to enable compilation also for ESP8266 and ESP32 platforms (to pass `static_asserts` checking for certain pin's).

The script `ci/ci-compile` performs the build and is run during a Github Actions workflow. The script can also be run locally.

See https://github.com/jandelgado/FastLED/actions/runs/445634210 for an example

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,28 @@
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
+
+name: build
+jobs:
+
+  build:
+    runs-on: ubuntu-20.04
+    steps:
+      - name: checkout code
+        uses: actions/checkout@v2
+
+      - name: install python
+        uses: actions/setup-python@v2
+        with:
+          python-version: '3.8'
+
+      - name: install platformio
+        run: |
+          pip install platformio==5.0.3
+
+      - name: build FastLED examples
+        run: ./ci/ci-compile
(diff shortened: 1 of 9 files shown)


kriegsman:
Thank you for this PR -- look extremely useful, and will be very helpful as we add more examples and more platforms.  I'm reviewing the code changes right now.

jandelgado:
Would also be great to "streamline" the examples, making them use the same set of `#defines/#ifdef` so that they can be more or less all configured the same, without changing the code. I can support here you think it's useful. See also #1158 ;-)

kriegsman:
OK, so @samguyer and I just had a conversation about this, and one of the things we kept coming back to was wanting to keep the examples as concise as possible so that new users didn't have to learn more than absolutely necessary to get started.

What do you think about just changing all the examples to just #define DATA_PIN 2 , so that it works on all the boards that you enumerated in the CI script -- but leave out the #ifndef / #endif around those #defines?

jandelgado:
That's fine - should I change it with this PR? This would also help ESP32/ESP8266 getting the examples run. I'll change it...

jandelgado:
Done - I minimized the changes by only changing the examples which did not compile for ESP32/8266.

kriegsman:
Thanks- this looks great!

kriegsman:
This PR is very much appreciated.  Now I have to go and learn how to use 'act' so that I can kick off these tests locally, too!  (Also going to look at that 'host' PR and idea.)

Thank you for the PR and for your contributions to FastLED!

samguyer:
@jandelgado and @kriegsman I just submitted a PR from my fork, and it looks like the tests are running automatically. Cool!

PR #698, opened 2018-12-13
Add WEB chipset and Unix platform

kw217:
This PR adds support for the following to FastLED:
* The `DUMMY` chipset. This performs no output. It's useful for stubbing out code, or for testing a new platform.
* The `__unix__` platform. This allows building the library for Linux, without requiring AVR or Arduino specials.
* The `WEB` chipset. This provides live output to a browser. It's only supported on `__unix__`.

I wanted to test my FastLED patterns without having to flash them to my Arduino or look at my LED array (which is fixed to the outside of my house and hence difficult to view from my workstation). So I added this new chipset, which runs a small webserver to which you can connect your browser. When you do so, the LEDs are displayed in the browser.

Use it like this:
```cpp
  FastLED.addLeds<WEB, 9002, 0>(leds, NUM_LEDS);
```
and then connect your browser to http://localhost:9002/ to see the show. Multiple LED strings are supported - just specify an index in place of `0`. Any port can be used - `9002` is just an example.

Obviously to run typical Arduino code in this way you'll need to provide a `main()` which calls `setup()` and `loop()`, and define any Arduino functions you use. If you guard these things and the `addLeds` all above with `#ifdef __unix__` then the same code will work for Arduino and Linux. To build for Linux you will need to install libwebsocketpp-dev (https://github.com/zaphoyd/websocketpp), and link with `-lboost_system -lpthread`.

Please let me know if there is anything I need to add to or change in this code to meet the FastLED project requirements.

Enjoy!

**How it works:** It runs a small combined web and websocket server. For web requests it serves a page with some Javascript, which then makes a websocket request back to the origin server. Each call to `showPixels` renders an n x 1 PNG image of the LEDs and pushes it over the websocket to the browser, which displays it.

PR #1158, opened 2020-12-29
add a "host" platform

jandelgado:
I added a "host" platform to compile FastLED on your host system, without cross-compilation.

The platform is selected by `#define` "FASTLED_HOST".

This allows e.g. to compile FastLED for the Web Assembly (WASM) target and run FastLED code directly in the browser (for demonstration and testing purposes).

Visit https://jandelgado.github.io/fastled-wasm for a demo. I had only to do slight modifications to the original FastLED examples. By adding a few `#ifdefs` and less use of global variables,  it would be also easily possible to directly compile the examples for WASM, without modifications.

image

BR
Jan

### DIFF
--- a/src/led_sysdefs.h
+++ b/src/led_sysdefs.h
@@ -5,7 +5,9 @@
 
 #include "fastled_config.h"
 
-#if defined(NRF51) || defined(__RFduino__) || defined (__Simblee__)
+#if defined(FASTLED_NO_MCU)
+#include "platforms/no_mcu/led_sysdefs_no_mcu.h"
+#elif defined(NRF51) || defined(__RFduino__) || defined (__Simblee__)
 #include "platforms/arm/nrf51/led_sysdefs_arm_nrf51.h"
 #elif defined(NRF52_SERIES)
 #include "platforms/arm/nrf52/led_sysdefs_arm_nrf52.h"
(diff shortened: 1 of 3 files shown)


kriegsman:
This is a great idea, and a nice concise PR for implementing it.  I have to admit that I've done this myself, by hand, a few times over the years, but this is a nice clean little bundle.

My only question about it is if we can come up with a more specific name than "HOST".  "Host" can mean a couple of different things, and I'd like to see if we can come up with a different name.  

What do you think of FASTLED_NO_MCU ?  Or alternatively, FASTLED_NATIVE ?  

Thank again for this idea and this implementation.  It's going to make a bunch of other future things easier, and I appreciate the 'incremental changes' approach here.

marmilicious:
FASTLED_NO_MCU seems nicely descriptive.

jandelgado:
Great! I'll change it to `FASTLED_NO_MCU`

jandelgado:
Done

kw217:
See also #698 for another approach to solving this issue.

zackees:
There is now a "stub" implementation that does exactly this but is more detailed. I think you'll want to just use that instead.

It's already part of our build suite so tested to work on every cl.

[linux_native](https://github.com/FastLED/FastLED/actions/workflows/build_linux.yml)

PR #1166, opened 2021-01-03
Add hardware SPI support for Arduino Nano Every

tom-miller1:
Tested working with official Arduino Nano Every and an APA102 LED strip.

The Arduino Nano Every pinout sheet (https://content.arduino.cc/assets/Pinout-NANOevery_latest.pdf),  lists the SPI0 ports in the ALT2 portmux position (PE[3:0]), so that's where they're set in this PR using the PORTMUX_TWISPIROUTEA register (page 138 in http://ww1.microchip.com/downloads/en/DeviceDoc/ATmega4808-4809-Data-Sheet-DS40002173A.pdf). This could be configurable with some more effort.


kriegsman:
Looks good - thank you for the PR.  If we need to make it more configurable later, we can look at that, but this is great for now for the Nano Every.  Thank you for helping out with FastLED!

PR #1167, opened 2021-01-04
Add interrupt enable and disable points between each tightly timed clockless signal.

ben-xo:
This enables very fast interrupts (such an audio sampler) to work for e.g. sound reactive light setups, with minimal jitter.

cc @samguyer - this is what I was discussing with you on Reddit at https://www.reddit.com/r/FastLED/comments/kpvni2/fastled_soundreactive_visualisers_beat_detection/gi0er9w/

There is an explanation in that thread of the limitations / the timing flexibility that can be afforded by neopixels even though it is out of spec. The key information on what timings are critical and which are not is here: https://wp.josh.com/2014/05/13/ws2812-neopixels-are-not-so-finicky-once-you-get-to-know-them/

This patch doesn't take the theory from that article to its limit; each bit is is still in its own critical section. But as long as your ISRs don't take more than about 90cycles (for example the `millis()` ISR takes around 50) then the lights won't latch and neopixels will happily show at a stable framerate even with jittery data.

Obviously with ISRs enabled like this I defined NO_CORRECTION to disable the unnecessary clock skew correction.

I can't promise this will work nicely with the LED dithering feature, however.

### DIFF
--- a/src/controller.h
+++ b/src/controller.h
@@ -148,7 +148,7 @@ public:
     }
 
     static CRGB computeAdjustment(uint8_t scale, const CRGB & colorCorrection, const CRGB & colorTemperature) {
-      #if defined(NO_CORRECTION) && (NO_CORRECTION==1)
+      #if defined(NO_COLOR_CORRECTION) && (NO_COLOR_CORRECTION==1)
               return CRGB(scale,scale,scale);
       #else
               CRGB adj(0,0,0);
--- a/src/platforms/avr/clockless_trinket.h
+++ b/src/platforms/avr/clockless_trinket.h
@@ -406,31 +406,31 @@ protected:
 				// Inline scaling - RGB ordering
 				// DNOP
 				
-				cli(); HI1 _D1(1) QLO2(b0, 7) LDSCL4(b1,O1) 	_D2(4)	LO1	sei();	PRESCALEA2(d1)	_D3(2) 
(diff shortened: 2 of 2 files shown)


kriegsman:
Hi @ben-xo - this, and the interleaved high-performance audio sampling and beat detection, are a brilliant accomplishment.  Kudos!  @samguyer and I had a long conversation about this PR this morning, and how we want to incorporate it into FastLED.  I have a couple of things that I want to think about a little bit, but this is great, and I wanted to let you know right away that I'm excited for what you've been able to do here, and for how it might open up more projects for other FastLED users.  Thank you for this, and I'll write more once I've had a chance to go for a walk and see if I can think of any other implications that we should be planning for with this change.  Thanks again!

ben-xo:
> Hi @ben-xo - this, and the interleaved high-performance audio sampling and beat detection, are a brilliant accomplishment. Kudos!

Thank you, totally made my day to hear that! 😊

ben-xo:
Just today I discovered one small issue worth noting (although not necessarily a problem). FastLED bitbangs en entire PORT rather than just sbi / cbi on individual pins. This means that if you are also toggling outputs on the same port in an ISR, that FastLED will reset those pins to whatever state they were in when it saves the mask.

Obviously with interrupts disabled this is fine, as no interrupt could change those pins; and also my complaint is esoteric, as I'm already  in uncharted territory. But it was something I noticed!

ben-xo:
I've been working on this branch for my personal project so it's a little out of date now. but #1263 and #1228 are related.

ben-xo:
This has been superceded by #1268

Issue #1034, opened 2020-06-06
Support for ESP32-S2


fabianoriccardi:
Hi,
I'm just testing the new [ESP32-S2 dev module](https://docs.espressif.com/projects/esp-idf/en/latest/esp32s2/hw-reference/esp32s2/user-guide-saola-1-v1.2.html). Is ESP32-s2 supported? Because the code compiles with [esp32s2 branch](https://github.com/espressif/arduino-esp32/tree/esp32s2), but the led doesn't turn on.

EDIT: the Adafruit Neopixel library seems already working, even if support to esp32-s2 is not yet stated.

DrJohnM61:
I have tried the Adafruit Neopixel library and it works, however, it does not work if you include the WiFiMulti and issue a WiFiMulti.run() command. This may be related to https://www.eevblog.com/forum/beginners/esp12f-with-ws2812-(neopixels)-within-arduino/ and the suggestion is to use FastLED. Any suggestion of how to use FastLED on a ESP32S2 would be appreciated.

roblatour:
I have it part-way working .with this code and an espressif  esp32-s2-saola-1 board

`#include <FastLED.h>

#define LED_PIN     18
#define NUM_LEDS    1
#define BRIGHTNESS  10
#define LED_TYPE    WS2812
CRGB leds[NUM_LEDS];

enum colours { Green, Blue, Red };

void setNeoPixelColour(colours choosenColour) {

  // https://github.com/FastLED/FastLED/blob/master/examples/RGBSetDemo/RGBSetDemo.ino
  // https://github.com/FastLED/FastLED/wiki/FastLED-HSV-Colors

  switch (choosenColour) {
    case Green:
      Serial.println("green");
      leds[0] = CRGB::Green;
      break;
    case Blue:
      Serial.println("blue");
      leds[0]  = CRGB::Blue;
      break;
    case Red:
      Serial.println("red");
      leds[0] = CRGB::Red;
      break;
    default:
      // statements
      break;
  }

  FastLED.show();   // <<<< problem here
  
}
`

`FastLED.addLeds<WS2812, LED_PIN, RGB>(leds, NUM_LEDS);
  FastLED.setBrightness(BRIGHTNESS );`


`setNeoPixelColour(Green);`


It will turn on the led to the desired colour, but after the FastLED.show(); command is executed the board reboots.

Here is the dump:

Blue
E (7363) rmt: rmt_set_pin(480): RMT CHANNEL ERR
[...]

PR #1181, opened 2021-01-22
Fixed ESP32 clockless RMT to work on ESP32S2

shlomozippel:
ESP32S2 has only 4 channels in the RMT peripheral compared to the 8 channels in ESP32. Some interrupt bits need to fixed to ensure it works properly on both platforms.

Sanity tested this fix on both ESP32 and ESP32S2.

I'm not sure this is the best way to test for S2 vs. vanilla - happy to change it if anyone suggests a better option


shlomozippel:
This fixes #1034 - at least partially. I2S still doesn't work and I didn't dive too deep into it

samguyer:
@shlomozippel Thank you for this contribution! I want to do a quick review and test, but I'll plan to merge it in soon.

PR #1193, opened 2021-02-22
Add support for Adafruit MatrixPortal M4

Talljoe:
Pulled the definitions from the [variants](https://github.com/adafruit/ArduinoCore-samd/blob/master/variants/matrixportal_m4/variant.cpp) file. Tested working (for pin 4/`PIN_NEOPIXEL`, at least). For SPI I chose the secondary SPI port, since the first is in use by the ESP32 coprocessor. It's rather academic as neither of the SPI ports are broken out to the board.

As one of the pins is bit 31, I needed to [change the definition](https://github.com/FastLED/FastLED/compare/master...Hevi-dev:matrixportal-m4?expand=1#diff-f18dce76c03225559d8bfb64c2b8ac4ce0aea030652e04f39599fb4a889f731eR62) of the macro to unisgned long. I hopefully did so in a manner that is compatible with other boards/compilers.

### DIFF
--- a/src/platforms/arm/d51/fastpin_arm_d51.h
+++ b/src/platforms/arm/d51/fastpin_arm_d51.h
@@ -59,7 +59,7 @@ public:
 
 #define _FL_IO(L) _RD32(GPIO ## L)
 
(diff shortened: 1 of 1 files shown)


kriegsman:
Thank you for the help and for your contribution to FastLED !

PR #1194, opened 2021-02-22
Add support for Seeed Wio Terminal

craiglink:
adds FastLED support to SAMD51-based  Seeed Wio Terminal


craiglink:
debugging an issue with hardware support

PR #1197, opened 2021-02-28
Ardunio Nona Every optimizations

craiglink:
Enables Hardware SPI interface by correcting pin number
Leverages the ATMega4809's VPORTs to have single cycle writes


kriegsman:
Nice optimizations, thank you!  Thanks for the PR and thanks for helping out with FastLED!

PR #1214, opened 2021-03-26
I encountered this problem, what is the reason? //fatal error: avr/io.h: No such file or directory//

loiswen:
Hello , I am a beginner of CC3200, I use CC3200 and WS2812B, I want to use this library to light up the led. But i have been unsuccessful, I encountered this problem, what is the reason? How to solve it?
E:\energia\energia\hardware\cc3200\libraries\FastLED/platforms/avr/led_sysdefs_avr.h:12:20: fatal error: avr/io.h: No such file or directory

Could someone help me?


toine512:
Please post an issue.

Issue #1208, opened 2021-03-17
Modernize the examples to help new users 


kriegsman:
There are a few little modernizations we could make to the example sketches that would help new users get started faster, and develop good habits from the start.

1. Switch from `FastLED.setBrightness(...)` to `FastLED.setMaxPowerInVoltsAndMilliamps( ... )`
2. [done!] Switch `LEDS.[function]` to `FastLED.[function]`
3. [done!] Switch from `byte` to `uint8_t`
4. [done!] Use `uint16_t` for all LED index variables, so that the examples all work with > 255 LEDs
5. Use `FastLED.showAndHold( ... )` (when that exists) instead of `FastLED.show(); FastLED.delay( ...);`, with an explicit note that it will keep actively refreshing the LEDs to make temporal dithering / fractional brightnesses work.  (See discussion in https://github.com/FastLED/FastLED/issues/1206, in terms of educating new users about the effects of these choices.) 
6. Try to make all examples run at about the same visual speed regardless of microcontroller speed, e.g. use `FastLED.showAndHold(...)`, or have animations be a function of _time_, rather than a function of _number-of-iterations_ .
7. Switch from manifest constants (`#define`) to `static const` where possible and more legible.
8. Switch from a startup `delay(3000);` to perhaps a warmup function, like `FastLED.fadeIn( 3000 )`, which doesn't exist yet in the library as of 3/17/2021, but which I have code for.
9. [done!] Remove messages about minimum version of FastLED to use
10. Unify overall structure of each sketch so they're easier to read as a group

There are other larger pieces of rework we could do, e.g. make the demo animations like Fire2012 into their own classes, with proper member variables to control parameters, etc., but this Issue is not for tracking that work; this issue is just for tracking the small modernizations listed above.

yesyesuk:
>> 7. Switch from manifest constants (#define) to static const where possible and more legible.
Just for my understanding, why is that better? It is my understanding that "static const" would take up memory whereas a "#define" would not.

kriegsman:
This is a great question!  Twenty (or thirty!) ago, most C or C++ compilers would have actually created the variable for the `static const`, which _would_ take up RAM at runtime, as you suggested that it might.  However, when any modern compilers see a `static const` value, they literally forward the value of the constant into the surrounding code, as if you had just typed the constant value inline yourself.  

[...]

PR #1218, opened 2021-04-04
Some Modernization updates for #1208

marmilicious:
As suggested in #1208 this updates:
LEDS.[function] to FastLED.[function]
Switch from byte to uint8_t
Use uint16_t for all LED index variables
Remove messages about minimum version

### DIFF
--- a/examples/Cylon/Cylon.ino
+++ b/examples/Cylon/Cylon.ino
@@ -15,8 +15,8 @@ CRGB leds[NUM_LEDS];
 void setup() { 
 	Serial.begin(57600);
 	Serial.println("resetting");
-	LEDS.addLeds<WS2812,DATA_PIN,RGB>(leds,NUM_LEDS);
-	LEDS.setBrightness(84);
+	FastLED.addLeds<WS2812,DATA_PIN,RGB>(leds,NUM_LEDS);
+	FastLED.setBrightness(84);
 }
 
 void fadeall() { for(int i = 0; i < NUM_LEDS; i++) { leds[i].nscale8(250); } }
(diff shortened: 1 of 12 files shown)


kriegsman:
This is great.  Thank you so much for moving this forward, and it's so nice not to see things like 'byte' in there any more.  

The only question I'm thinking about is whether in the XYmatrix examples, whether the X and Y indicies should be 16-bit themselves, or just the "which pixel number" indicies.   If we leave the X and Y indicies at 8 bits each, we can still handle a matrix up to 256x256 pixels with no code change, and everyone with smaller matrices would still get the speed benefits, at least on 8-bit microcontrollers like AVR.  OK, now that I've talked it through, I think that's the way to go: use 8-bit indicies for X and Y values, but make sure that the 'pixel number' indicies are all 16-bit.  If you would make that change, I'll accept and merge the PR right away.  

Thanks again for all the help.  It really makes a difference.

marmilicious:
I did wonder about that and was trying to think what a more edge like case might be.  Perhaps someone making something like a 400x8 matrix for a scrolling text marque?  I'll switch that back though.

marmilicious:
@kriegsman  I wasn't sure if this was ok, or if you would like me to delete my first PR and submit a new one?  If you have a preference or suggestion for the future please let me know.  I'm still a bit green when it comes to doing PRs, but am happy to learn and help out where I can.

kriegsman:
Looks great!  Thank you so much for helping out with this, especially because the example sketches are where everyone starts out, and so improvements there help them get started more quickly and more easily and on the right track.  Very much appreciated.

elliottkember:
> I did wonder about that and was trying to think what a more edge like case might be. Perhaps someone making something like a 400x8 matrix for a scrolling text marque? I'll switch that back though.

@marmilicious This has actually come up for me when writing resolution-independent patterns. Sometimes my matrix is 20x20 and everything is fine, but then other times it's 400x1 for a single strip. 

This is fine for me as I redeclare XY() with 16-bit x and y parameters, but might be worth considering!

PR #1222, opened 2021-04-12
Use FASTLED_INTERNAL define to suppress pragma message from the ESP32 I2S driver

markuslang79:
A simple change to check for a set `FAST_INTERNAL` define to suppress the _"NOTE: ESP32 support using I2S parallel driver. All strips must use the same chipset"_ message from the ESP32 I2S driver if you know what you're doing and want to clean up your compile output a bit :wink:

### DIFF
--- a/src/platforms/esp/32/clockless_i2s_esp32.h
+++ b/src/platforms/esp/32/clockless_i2s_esp32.h
@@ -86,7 +86,9 @@
 
 #pragma once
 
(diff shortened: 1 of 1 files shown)


samguyer:
@markuslang79 Thanks for your contribution! I just merged it into the main code base. And yes, we are trying to cut down on the number of pragma compiler warnings, which seem to just confuse people.

Issue #1224, opened 2021-04-21
MS_COUNTER undefined


i164j9:
in file block_clockless_arm_mxrt1062.h when compiling for use on the teensy 4.1 MS_COUNTER is undefined at line 103.

i164j9:
The version is 3004000 downloaded through platformio, the same applies in the code here in github.

zackees:
Reopening: this issue was closed automatically by a stale-bot run, not by a maintainer decision. That auto-close behavior was a mistake and has been disabled — FastLED issues will no longer be closed for inactivity. Apologies for the noise.

PR #1226, opened 2021-04-23
Defined MS_COUNTER

i164j9:
issue #1224
Declared MS_COUNTER and initialized to 0.
previously undefined when using with teensy 4.1


kriegsman:
Thank you for the contribution and thanks for the help with FastLED.

PR #1240, opened 2021-05-26
Add missing const in unary operator-

ChristofKaufmann:
The unary minus operator of `CRGB` returns a new `CRGB` object and should therefore be `const`. In fact, this is a bug. Having a function like the following cannot call the unary minus.
```c++
CRGB complement(CRGB const& color) {
    return -color; // yields CRGB(255, 255, 255)
}
```
Adding a `const` qualifier should resolve the issue.

### DIFF
--- a/src/pixeltypes.h
+++ b/src/pixeltypes.h
@@ -412,7 +412,7 @@ struct CRGB {
     }
 
     /// invert each channel
-    inline CRGB operator- ()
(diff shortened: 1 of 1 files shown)


kriegsman:
Thank you for finding this, for submitting the PR, and for the help with FastLED.  Much appreciated.

PR #1251, opened 2021-06-22
Fix incorrectly calling perlin noise "Simplex Noise"

aykevl:
Simplex noise is something different and is not implemented by the FastLED library (yet!). Therefore, don't confuse people by calling it Simplex Noise.

It appears there are 3 variants of noise developed by Ken Perlin:

- Classical Perlin noise.
- An "improved" noise - AFAIK just small improvements over classical Perlin noise.
- Simplex Noise. This algorithm is a bit more complicated and works in a different way. It allows noise up to many dimensions without too much computational overhead: O(n+1) instead of O(2^n). It also has fewer artifacts, that are sometimes quite visible in classical/improved noise. Downside: it's patented until early 2022.

I've looked at the noise.cpp file and it looks a lot more like the improved noise (https://mrl.cs.nyu.edu/~perlin/noise/) than simplex noise (https://github.com/stegu/perlin-noise/blob/master/src/simplexnoise1234.c).

Note: I have converted the simplex noise algorithm (1D, 2D, 3D, 4D) over to a fixed point implementation (written in Go) that I'm now converting to C++ for integration in FastLED. I hope to make a PR soon. This PR is just to fix some issues with the docs.

### DIFF
--- a/examples/NoisePlusPalette/NoisePlusPalette.ino
+++ b/examples/NoisePlusPalette/NoisePlusPalette.ino
@@ -15,7 +15,7 @@ const bool    kMatrixSerpentineLayout = true;
 
 // This example combines two features of FastLED to produce a remarkable range of
 // effects from a relatively small amount of code.  This example combines FastLED's 
-// color palette lookup functions with FastLED's Perlin/simplex noise generator, and
+// color palette lookup functions with FastLED's Perlin noise generator, and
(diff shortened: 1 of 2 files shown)


kriegsman:
Thanks for the help with FastLED -- and with the noise functions in particular.  Much appreciated.

PR #1252, opened 2021-06-22
Add simplex noise [revisit this PR in 2022]

aykevl:
Simplex noise has some benefits over classical (or improved) Perlin noise. It generally has fewer artifacts than previous versions. It still has artifacts, just a lot less noticeable. It is also possible to easily extend the algorithm to higher dimensions without exponentially increasing the computational requirements. One practical advantage is that this implementation also includes a 4D version while the other noise functions only implement 1D, 2D and 3D versions.

The implementation has been optimized for 32-bit microcontrollers. I've tried to keep all calculations in 32-bit integers, only going to 64-bit when really necessary.

The API is a bit different because I converted to fixed-point separately. I also didn't care about possible benefits for microcontrollers for a 16.16 fixed point format, so I picked what seemed better to me at the time: a 19.12 format (or 20.12 if you count the sign: it's a signed integer).

Note that this is covered by a patent, probably until early 2022: https://patents.google.com/patent/US6867776
However, it might not apply to simple LED animations: https://patents.stackexchange.com/questions/18573
The patent is currently owned by a patent troll so it would be rather risky to bring any products to market that uses this algorithm. If this is a problem, it may be better to just wait with merging this PR until the patent has expired.

Also see #1251.

### DIFF
--- a/src/simplex.cpp
+++ b/src/simplex.cpp
@@ -1,0 +1,461 @@
+#define FASTLED_INTERNAL
+#include "FastLED.h"
+
+// This file implements simplex noise, which is an improved Perlin noise. This
+// implementation is a fixed-point version that avoids all uses of floating
+// point while still being compatible with the floating point version.
+
+// Original author: Stefan Gustavson, converted to Go by Lars Pensjö, converted
+// to fixed-point Go and then to C++ by Ayke van Laethem.
+// https://github.com/larspensjo/Go-simplex-noise/blob/master/simplexnoise/simplexnoise.go
+// https://github.com/aykevl/ledsgo/blob/master/noise.go
+//
+// The code in this file has been placed in the public domain. You can do
+// whatever you want with it. Attribution is appreciated but not required.
+
+// Notation:
+// Every fixed-point calculation has a line comment saying how many bits in the
+// given integer are used for the fractional part. For example:
+//
+//     uint32_t n = a + b; // .12
+//
+// means the result of this operation has the floating point 12 bits from the
+// right. Specifically, there are 20 integer bits and 12 fractional bits. It
+// can be converted to a floating point using:
+//
+//     double nf = (double)n / (1 << 12);
+
+FASTLED_NAMESPACE_BEGIN
+
+#define P(x) FL_PGM_READ_BYTE_NEAR(p + (x))
+
+// Permutation table. This is just a random jumble of all numbers.
+// This needs to be exactly the same for all instances on all platforms,
+// so it's easiest to just keep it as static explicit data.
+FL_PROGMEM static uint8_t const p[] = {
+    151, 160, 137, 91, 90, 15,
+    131, 13, 201, 95, 96, 53, 194, 233, 7, 225, 140, 36, 103, 30, 69, 142, 8, 99, 37, 240, 21, 10, 23,
+    190, 6, 148, 247, 120, 234, 75, 0, 26, 197, 62, 94, 252, 219, 203, 117, 35, 11, 32, 57, 177, 33,
+    88, 237, 149, 56, 87, 174, 20, 125, 136, 171, 168, 68, 175, 74, 165, 71, 134, 139, 48, 27, 166,
+    77, 146, 158, 231, 83, 111, 229, 122, 60, 211, 133, 230, 220, 105, 92, 41, 55, 46, 245, 40, 244,
+    102, 143, 54, 65, 25, 63, 161, 1, 216, 80, 73, 209, 76, 132, 187, 208, 89, 18, 169, 200, 196,
+    135, 130, 116, 188, 159, 86, 164, 100, 109, 198, 173, 186, 3, 64, 52, 217, 226, 250, 124, 123,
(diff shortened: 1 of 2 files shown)


src/simplex.cpp
```diff
@@ -0,0 +1,466 @@
...
+    223, 183, 170, 213, 119, 248, 152, 2, 44, 154, 163, 70, 221, 153, 101, 155, 167, 43, 172, 9,
+    129, 22, 39, 253, 19, 98, 108, 110, 79, 113, 224, 232, 178, 185, 112, 104, 218, 246, 97, 228,
+    251, 34, 242, 193, 238, 210, 144, 12, 191, 179, 162, 241, 81, 51, 145, 235, 249, 14, 239, 107,
+    49, 192, 214, 31, 181, 199, 106, 157, 184, 84, 204, 176, 115, 121, 50, 45, 127, 4, 150, 254,
+    138, 236, 205, 93, 222, 114, 67, 29, 24, 72, 243, 141, 128, 195, 78, 66, 215, 61, 156, 180,
+};
+
+// A lookup table to traverse the simplex around a given point in 4D.
+// Details can be found where this table is used, in the 4D noise method.
+// TODO: This should not be required, backport it from Bill's GLSL code!
```
aykevl, line 58: I actually tried this, but the GLSL code is not faster.

1. I tried it in my Go implementation (running on my laptop), where it was actually a bit slower.
2. I tried roughly the same code on an ESP32, but it didn't affect the speed (the difference was well within the noise).

For those interested, see: https://gist.github.com/aykevl/dddc20182537f7c00f8b6358de97fc6d

kriegsman:
Thank you for your work on this -- including the 32-bit-friendly fixed-point implementation!  
I think I agree with you that if there are 'patent trolls' involved, I'd rather wait out the patent expiration, especially because it's not _very_ far away.
In the meantime, I'm going to accept your other contribution - to change the terminology to just Perlin noise in the existing code.

The existing noise functions are overdue for a little careful code review and cleanup; I think we're losing a little precision here and there that we don't need to be.  But that's another problem for another day, I think.

Anyway, thanks again for all the contributions, and the help with FastLED!

aykevl:
Actually, this was never specifically optimized for 32-bit MCUs. I just wanted something that was better than improved Perlin noise and I happened to be using 32-bit MCUs (because I absolutely needed the extra performance for what I was animating: 6144 LEDs at a usable frame rate).

>  I think I agree with you that if there are 'patent trolls' involved, I'd rather wait out the patent expiration, especially because it's not _very_ far away.

Okay, sounds reasonable to me. Note that the patent (from my reading) is only for 3D and up, so the 1D and 2D versions might be safe to merge.

Regarding the API: I'm now thinking it might be better to use unsigned integers for the API. That would be a bit more compatible with the existing noise functions, but with the input shifted 4 bits. I used signed integers because I adopted it from a floating point implementation (which are naturally signed), not as a conscious decision.

aykevl:
I have updated the PR:

* I've rebased it on the master branch.
* I've fixed some rare bugs (I think overflows) that sometimes happened in my Go variant of the code. I don't have similar tests for the C variant but I applied the same fix so I hope it'll work there.
* I've changed the API to use unsigned integers instead of signed integers. The output is still signed but I scaled it so it is an unsigned number averaging around 0x8000. (It would be possible to add `*_raw` variants like for Perlin noise but that can easily be done at a later time if needed).

Also, it appears that the patent has expired. The legal status of [US6867776B2](https://patents.google.com/patent/US6867776B2) is listed as "Expired - Lifetime" so I guess we're in the clear now. Therefore, from my perspective this PR is ready.

aykevl:
Rebased to fix the merge conflict, and removed the patent warning (because the patent is expired).

This PR seems ready to me, is there anything missing?

---

Meanwhile I've found that there were some unintended division instructions in the code (I assumed they would be optimized away). This is fixed [here](https://github.com/aykevl/ledsgo/commit/f76cfa27fe12a2e8bdb0c722e336b12225a36b61) but not yet in this PR. It basically makes the noise functions a lot faster on chips that don't have a hardware divide instruction like Cortex-M0. I might update the PR eventually but I don't currently have an easy way to test FastLED changes.

zackees:
Patent has expired so no more issues with this being merged in

aykevl:
Thank you for merging!

Note that since I made this PR, I've optimized the original code to avoid a division instruction that wasn't optimized as I had expected. This would result in a 3.6x speedup in the 1D case (less so in 2D an up). The change is here: https://github.com/aykevl/ledsgo/commit/f76cfa27fe12a2e8bdb0c722e336b12225a36b61
It should be fairly easy to port this change over to FastLED (the code is basically the same except for the language) but I am not using FastLED at the moment so have no way to test this change. In any case, for anybody wanting to improve the performance, that's a way to do it.

zackees:
I've lifted your changed into a new PR. You were tagged so you should see it in your notifications.

aykevl:
Actually I wasn't notified, I don't see you tagged me?

zackees:
https://github.com/FastLED/FastLED/pull/1690

PR #1266, opened 2021-07-29
Remove clobbering of TCCR0A left over from ancient times

ben-xo:
Daniel Garcia commented out all of the other code which goes with this block in b9b23091 (2015). Remove this last remnant.

PR #1265, opened 2021-07-28
Remixing pixels on a single strip (and AVR clockless bugfixes)

ben-xo:
I present here a patch (which is currently only implemented fully for Clockless Trinket) which fixes a few long standing bugs as well as enabling new features.

Features
---------

* Interrupts can now be enabled on AVR. 
* Linked-list type "PixelCommand" (implemented only in Clockless Trinket at the moment) allows you to link together pointers within your LED array to render on the same strip without physically moving the data round. This allows for example:
-- free "reverse" (on a per sub-block basis)
-- mirroring 
-- repeating a short array of pixels onto a longer strip on a loop
-- rotation (showing N...END, 0..N, then N+1..END, 0..N+1, etc)  
-- offsetting
-- could also be used for scrolling vertically or horizontally on grids
* define to disable the minimum wait period (if you do your own FPS)

Bugs Fixed
-----------

1. clockless_trinket can now use negative skips. I jumped through some hoops to get this to work on gcc by freeing up 3 variables in the ASM constraints!
3. NO_CORRECTION in controller.h is renamed to NO_COLOR_CORRECTION, because it clashes with a NO_CORRECTION In clockless_trinket (used to disable the clock skew correction
3. removed trace of timer juggling - clockless_trinket should not be interfering with TCCR0A
4. pixel arrays of length 0 no longer hang.

If you want to see it in action… i can provide video.

PR #1267, opened 2021-07-29
Prevent FastLED from hanging on 0-length pixel arrays

ben-xo:
Without this patch, a 0 length pixel array is treated as if it's 65535 long. Whilst few people will encounter this bug in simple applications, it comes up frequently when manipulating ranges of pixels with multiple PixelControllers.

### DIFF
--- a/src/platforms/avr/clockless_trinket.h
+++ b/src/platforms/avr/clockless_trinket.h
@@ -115,7 +115,9 @@ protected:
 		mWait.wait();
 		cli();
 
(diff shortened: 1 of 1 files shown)


ben-xo:
Part two of the changes from my previous PR (#1265) which I'm separating out into separate pull requests, following from #1266

kriegsman:
Thanks, @ben-xo.  Nice catch, and nice change.

PR #1268, opened 2021-07-29
Allow interrupts on WS2811 etc on AVR without glitches

ben-xo:
For many WS2811 strips, we can relax the clearing of interrupts down to protecting the timing of the shortest pulse. Your interrupts must still be very very fast in order for the strip not to latch part-way through.

### DIFF
--- a/src/platforms/avr/clockless_trinket.h
+++ b/src/platforms/avr/clockless_trinket.h
@@ -31,11 +31,13 @@ FASTLED_NAMESPACE_BEGIN
 // condition that the shortest pulse in the specification (representing a 0 bit) is kept under a certain length. After exceeding
 // that length it would be interpreted as a 1, causing a glitch. 
 // 
-// If you set FASTLED_ALLOW_INTERRUPTS to 1, interrupts will only be enabled when necessary to keep this short pulse short.
+// If you set FASTLED_ALLOW_INTERRUPTS to 1, interrupts will only be disabled for a few cycles at a time, when necessary to keep
+// this signal pulse short.
 // 
 // Beware: even with FASTLED_ALLOW_INTERRUPTS enabled, you must ensure that your interrupt handlers are *very* fast. If they take
 // longer than about ~6000ns, which is roughly 90 clock cycles on a 16MHz AVR, the strip will latch partway through rendering, and
-// you will see big glitches.
+// you will see big glitches. If you are using multiple timers with interrupts, you can set them out of phase so they only fire
+// one at a time.
 //
 // See https://wp.josh.com/2014/05/13/ws2812-neopixels-are-not-so-finicky-once-you-get-to-know-them/ for more information on the
 // tolerances.
--- a/src/platforms/avr/clockless_trinket.h
+++ b/src/platforms/avr/clockless_trinket.h
@@ -35,9 +35,15 @@ FASTLED_NAMESPACE_BEGIN
 // this signal pulse short.
 // 
 // Beware: even with FASTLED_ALLOW_INTERRUPTS enabled, you must ensure that your interrupt handlers are *very* fast. If they take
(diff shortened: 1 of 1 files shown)


ben-xo:
Part three of the changes from my previous PR (#1265) which I'm separating out into separate pull requests, following from #1267

ben-xo:
@kriegsman at your request the new behaviour is only activated if FASTLED_ALLOW_INTERRUPTS is set to 1.

ben-xo:
I'll add a bit more detail / background here.

Even with this PR, the default platform behaviour for AVR Clockless remains unchanged (because `FASTLED_ALLOW_INTERRUPTS` is not set by derault). 

To make use of this at all, you must `#define FASTLED_ALLOW_INTERRUPTS 1` yourself, as if to assert that you really do want this, and that hopefully you understand the various caveats, consequences and shortcomings, which are all listed in source code (where you will bump into them if you found that you had to dig and figure out why interrupts weren't working, like I did).

The caveats are various. In principal, you are sacrificing specification conformance. In practice, different makes, models or manufacturers of LEDs may have different tolerances to the out-of-spec behaviour, although I have not so far met a strip on which this hasn't worked.

The principle being applied, [inspired by this blog post](https://wp.josh.com/2014/05/13/ws2812-neopixels-are-not-so-finicky-once-you-get-to-know-them/ ), is that there are only 2 timings which matter for ws2812, even though 5 are defined in the spec.

The rationale for why the other timings in the spec are irrelevant is discussed in detail in the blog post, and if you study it for a while it becomes increasingly hard to imagine that there would be any pixel models which would obey any other timing directly; it would seem more costly and unnecessarily complicated.

The two timings in question constrain the maximum width of a HIGH control pulse that will still be interpreted as a `0` bit, and the maximum width of a LOW control pulse before the pixel strip will latch. These two constraints imply:

* if an interrupt fires within the first "i'm still 0" part of the control pulse, it all but guarantees it will flip to a 1 bit. This happens often enough in practise that it results in "speckles" appearing constantly along the strip in random places with random colours, due to the timer0 interrupt used by Arduino for the clock. To fix this, the first part of the pulse (which guarantees the 0-ness or 1-ness of a data bit) must be in a critical section - and this is the main contribution of this PR.

* if your interrupt handler takes longer than the maximum width of LOW before latching, you will see corruptions on the strip, although they're quite different to the speckles from the bit-flipping.
 
Which sort of corruptions you may see depends mainly on the frequency of your long interrupts, and your target frame rate, although there are some interesting variations.

If your long interrupts are infrequent and your frame rate high, you may see the occasional bright flash across the picture, starting at 0 but going up some random amount. 

If the frame rate is too low, those flickers may stay displayed for a long time. The "flicker" is actually a partially corrupted frame with random colours on some proportion of the lowest pixels. This is because if a latch event happens partway through a byte, the remaining bits from the frame will effectively start a new frame on the strip at pixel 0, and 7/8ths of the time the color bits will be offset with bits from adjacent colors leading to corruption that's very bright and obnoxious unless all your LEDS are set to black.

Alternatively, if the interrupt rate is so high that it interrupts every frame, perhaps more than once, then you may end up with a "cutoff" pixel beyon which your picture can never seem to reach. The higher the interrupt frequency, the closer to 0 that cutoff will get - and so all the pixels will be colour gibberish all the time.

Nevertheless, there are some things you can control.

One thing to consider is that your interrupts may be short enough if it fires alone, but may be too long if it fires simultaneously with another interrupt. This is especially relevant for timer interrupts, remembering that the the Arduino Core clock timer is a thing. 

It's often possible to refactor ISRs to just buffer work for when back in the main loop (e.g. "`button_was_pushed = 1; return` etc). Timers at related frequencies can also be initialised at an offset from each other so they don't coincide with timer0 or each other. 

I have no idea if other popular Arduino libraries have ISRs which are too long and run at a high frequency. I have not tried the "servo" library that seems to come up in support requests a lot.

ben-xo:
Any feedback which would get this merged?

PR #1270, opened 2021-08-01
Feature/avr clockless trinket negative advance by

ben-xo:
This PR mainly allows Clockless Trinket to deal with negative advanceBy values, which are allowed.

A negative advanceBy implies that you want to render an array of LEDS backward, starting at the end. 

Example usage:

Show forward (as usual, but spelled out to illustrate the subsequent example):
```C++   
FastLED[0].setLeds(&leds[0], STRIP_LENGTH);
FastLED.show();
```

Show in reverse:
```C++      
FastLED[0].setLeds(&leds[STRIP_LENGTH-1], -STRIP_LENGTH);
FastLED.show();   
```

I don't think the change to Controller::show() is ideal; I think this needs explicit support in the API. (it's currently only half there). It appears to be very difficult to access PixelController objects from the main scope, as they're created by show() right before showing.

But, regardless, the changes to clockless_trinket are independent and could be taken on their own.

### DIFF
--- a/src/controller.h
+++ b/src/controller.h
@@ -407,7 +407,7 @@ protected:
         PixelController<RGB_ORDER, LANES, MASK> pixels(data, nLeds < 0 ? -nLeds : nLeds, scale, getDither());
         if(nLeds < 0) {
             // nLeds < 0 implies that we want to show them in reverse
-            pixels.mAdvance = -3;
+            pixels.mAdvance = -pixels.mAdvance;
         }
         showPixels(pixels);
     }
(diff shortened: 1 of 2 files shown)


ben-xo:
Part four of the changes from my previous PR (#1265) which I'm separating out into separate pull requests, following from #1268

cc @kriegsman

kriegsman:
Thanks for this contribution, @ben-xo.  Nice way to allow mirroring of LED patterns on multiple strips without requiring extra memory.

Issue #1263, opened 2021-07-25
NO_CORRECTION define is used for two different purposes


ben-xo:
I noticed that NO_CORRECTION is used both [here](https://github.com/FastLED/FastLED/blob/5cc17b2be88982eb34b1998de22b83fee56d4f07/src/platforms/avr/clockless_trinket.h#L90) and [here](https://github.com/FastLED/FastLED/blob/5cc17b2be88982eb34b1998de22b83fee56d4f07/src/controller.h#L151) for different purposes.

The first one is to turn off clock-correction in AVR (which you might want to do if you have got interrupts working) and the second is for color-correction.

Given that this is a compatibility break I'd suggest renaming both defines (NO_COLOR_CORRECTION / NO_CLOCK_CORRECTION) and checking both the new name and the old name NO_CORRECTION (or maybe print out a warning)

kriegsman:
Finding this is a good catch; thank you. 
I like the idea of checking both the new and old name.  I don't think a warning would be net-helpful; the people who need to separately control the two settings will figure it out, and the people who don't need to do that probably shouldn't be bothered with a new warning on their existing code.  But: I want to think about that a little bit, because in general I do like giving informative warnings to people who might will benefit from seeing them.

ben-xo:
@kriegsman is there a project chat / IRC / Slack / whatever? I know the reddit, but that doesn't seem like a chat venue for project discussion

ben-xo:
Raised a PR for this. #1271

ben-xo:
Closed as my PR was merged

PR #1271, opened 2021-08-01
Bugfix/duplicate define no correction

ben-xo:
An implementation of the solution suggested on #1263 

The NO_CORRECTION define accidentally has two meanings. On clockless trinket, when combined with FASTLED_ALLOW_INTERRUPTS, it disables the clock skew correction. (with my AVR interrupts branch clock skew correction becomes unnecessary, but that's another story).

This change renames the more specific define to NO_CLOCK_CORRECTION. In configurations where it would have made a difference, NO_CORRECTION will still work the same way, but will print a helpful warning of how to fix the problem to be unambiguous.

**N.B.** the warning is only true if my PR #1268 is merged, otherwise it might need rewording.

### DIFF
--- a/src/platforms/avr/clockless_trinket.h
+++ b/src/platforms/avr/clockless_trinket.h
@@ -87,9 +87,9 @@ template<> __attribute__((always_inline)) inline void _dc<20>(register uint8_t &
 //
 //////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////////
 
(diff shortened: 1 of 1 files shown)


ben-xo:
Part five of the changes from my previous PR (#1265) which I'm separating out into separate pull requests, following from #1270

ben-xo:
Part six of the changes from my previous PR (#1265) which I'm separating out into separate pull requests, following from #1271

kriegsman:
Thanks for the contributions, @ben-xo !

PR #1272, opened 2021-08-01
Make the minimum wait optional by defining NO_MINIMUM_WAIT

ben-xo:
You may want this if you are doing your own frame-rate accounting and don't want to waste cycles waiting for FastLED between frames.

You may also want this if you are intentionally calling show() multiple times for the same strip.

### DIFF
--- a/src/fastled_delay.h
+++ b/src/fastled_delay.h
@@ -8,6 +8,9 @@
 
 FASTLED_NAMESPACE_BEGIN
 
+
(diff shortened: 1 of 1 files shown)


ben-xo:
Part six of the changes from my previous PR (#1265) which I'm separating out into separate pull requests, following from #1271

kriegsman:
Thanks for this one, too, @ben-xo !

PR #1274, opened 2021-08-04
fixed os_intr_lock without unlock

sascha432:
At this point interrupts are unlocked, calling os_intr_lock() without unlock leaves them locked.
This state fixes itself once the main loop() is left, but might cause issues with other code that runs later in the loop() function.

If interrupts are locked, can_yield() returns false even outside any ISR. It seems this does not cause any issues with the WDT since the NMI cannot be locked, but once the loop() function is left, all the code that was paused runs at once.

### DIFF
--- a/src/platforms/esp/8266/clockless_esp8266.h
+++ b/src/platforms/esp/8266/clockless_esp8266.h
@@ -42,9 +42,7 @@ protected:
       #ifdef FASTLED_DEBUG_COUNT_FRAME_RETRIES
       ++_retry_cnt;
       #endif
-      os_intr_unlock();
       delayMicroseconds(WAIT_TIME);
-      os_intr_lock();
     }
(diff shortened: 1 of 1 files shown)


DmytroKorniienko:
@sascha432 

Seems you moving in a right direction but still yelld() exceprion with you fix + a lot of frame drops. It's a pity but don't useble at this point.

Test environment:  ESP-8266 NodeMCU module, SDK3.0.2, WS8212 matrix 16*16 LEDS, FastLED.show() called each 20ms (50fps), Fastled - latest from your fork.

samguyer:
@sascha432 Thank you for this PR! We included these changes recently, but we did it as a separate commit in order to address a similar problem inside the showRGBinternal function. Take a look at the current head. If it looks good to you, we can close this PR.

Issue #1259, opened 2021-07-19
Teensy4 WS2812B not working


imwhocodes:
Hi

I'm trying to use Fastled with a Teensy4 but the resulting colors on the connected ledstrip are wrong
It seem like the color output is shifted or Xored is some way

Here  a sample code for reproducing the issue, all the led strip stay at half power white
(On a Teensy 3.2 the code work without any problem)

```
#include <Arduino.h>

#include "FastLED.h"

#define LED_PIN_1 9
#define STRIP_LENGTH_1 35
CRGB data_1 [STRIP_LENGTH_1];


#define LED_PIN_2 10
#define STRIP_LENGTH_2 35
CRGB data_2 [STRIP_LENGTH_2];


#define LED_PIN_3 11
#define STRIP_LENGTH_3 25
CRGB data_3 [STRIP_LENGTH_3];


#define LED_PIN_4 12
#define STRIP_LENGTH_4 19
CRGB data_4 [STRIP_LENGTH_4];


void setup() {

  FastLED.addLeds<NEOPIXEL, LED_PIN_1>(data_1, STRIP_LENGTH_1);

  FastLED.addLeds<NEOPIXEL, LED_PIN_2>(data_2, STRIP_LENGTH_2);

  FastLED.addLeds<NEOPIXEL, LED_PIN_3>(data_3, STRIP_LENGTH_3);

  FastLED.addLeds<NEOPIXEL, LED_PIN_4>(data_4, STRIP_LENGTH_4);

  FastLED.clear();

}

void loop() {

  FastLED.show();

}
```

Also attached a photo of the signal of one of the pins if it can be of any help
![IMG_4982](https://user-images.githubusercontent.com/14922653/126173559-c0f2e5d8-04a7-4f74-bc8c-6a85763ca55e.jpg)

Tested on:
- Platformio 5.1.1
- FastLed 3.3.0 / 3.4.0 / git-master
- TennsyCore 1.53 / 1.54

Thanks,
Luca

marmilicious:
Hello @imwhocodes   You need to specify NUM_STRIPS when using Teensy 4  

https://github.com/FastLED/FastLED/wiki/Parallel-Output#parallel-output-on-the-teensy-4

FastLED.addLeds<NUM_STRIPS, WS2812, PIN, GRB>(leds, NUM_LEDS);

imwhocodes:
Thanks @marmilicious 

[...]

PR #1277, opened 2021-08-10
Fix flipped bits in Teensy 4 clockless WS2812 driver

DavidMenting:
Solves an issue where bits are flipped when using the WS2812 on Teensy 4. 
Thanks to @somewhatlurker for unraveling the bug.
Relevant issue: https://github.com/FastLED/FastLED/issues/1259


v-lopez:
It seems this was applied here https://github.com/FastLED/FastLED/commit/8a82b609864146c25693da6738564e1aa4febf91

PR #1280, opened 2021-08-17
Create fill_rainbow_circular and fill_palette_circular functions

dmadison:
This adds a utility function (`fill_rainbow_endless`) for filling a complete rainbow pattern onto a strip so that it's continuous in hue between the end of the strip and the beginning. Useful for out-of-the-box rainbow animations on looped strips and LED rings.

#### Basic Demo:
```cpp
void loop() {
    static uint8_t hue = 0;
    EVERY_N_MILLIS(2) { hue++; }
    fill_rainbow_endless(leds, NUM_LEDS, hue);
    FastLED.show();
}
```

#### Result (12 LED ring):

https://user-images.githubusercontent.com/24282108/129810065-a7c7325e-2ba4-42af-b8a5-01132f9e012f.mp4

### DIFF
--- a/keywords.txt
+++ b/keywords.txt
@@ -130,8 +130,9 @@ fade_video	KEYWORD2
 fill_gradient	KEYWORD2
 fill_gradient_RGB	KEYWORD2
 fill_palette	KEYWORD2
+fill_palette_circular	KEYWORD2
 fill_rainbow	KEYWORD2
-fill_rainbow_endless	KEYWORD2
+fill_rainbow_circular	KEYWORD2
 fill_solid	KEYWORD2
 map_data_into_colors_through_palette	KEYWORD2
 nblend	KEYWORD2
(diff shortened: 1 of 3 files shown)


kriegsman:
Hi @dmadison, and Happy New Year.  Thank you for this contribution to FastLED, and I really like the idea.

What do you think of the idea of adding a similar function for `fill_palette` as well, perhaps `fill_palette_eternal`?  Also I was thinking about the name; what do you think of `fill_rainbow_circular` or `fill_rainbow_to_fit`?

marmilicious:
`fill_rainbow_circular` would get my vote! Nice addition @dmadison.

dmadison:
Thanks for the feedback! I think a similar function for the palettes could be useful as well. I'll see if I can implement that this week.

I'll agree with @marmilicious - `fill_rainbow_circular()` has a nice ring to it. I'll push another commit to change the name.

My initial thought was to use `endless()` because `circular()` seems to imply to me that the function is only for LEDs in a circular layout. Though now that I think about it "endless" has its own confusing connotations...

kriegsman:
Thank you for the contribution, and for your help with FastLED!

PR #1283, opened 2021-08-21
automatically build and deploy docs upon release

5chmidti:
This github action builds and deploys the docs on each release.
Current behavior of this pr is to overwrite existing docs.
~Current behavior of this pr is to create a new directory for the new version of docs.~
~As an example gh-pages/docs/3.1 currently exists and is pointed to by fastled.io/docs.~
When a new release is created, this action would overwrite the docs in the gh-pages/docs/ directory.
The existing docs/3.1 folder should get removed after merging this pr (or even after generating the new docs).
The library version of the library is updated in the Doxyfile and commited to master.

Also fixes the Doxyfile not pointing into the new src/ directory.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,0 +1,37 @@
+name: Build docs via Doxygen and publish to gh-pages
+on:
+  release:
+    types: [published]
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Generate Docs
+        run: |
+          git fetch
+          git checkout origin/master
+
+          cd docs
+          sed -i -E 's/(PROJECT_NUMBER\s*=\s*).*/\1${{ github.event.release.tag_name }}/g' Doxyfile
+
+          git config --global user.name "github-actions[bot]"
+          git config --global user.email "github-actions[bot]@users.noreply.github.com"
+
+          git commit Doxyfile -m "Update release version in Doxyfile"
+          git push origin HEAD:master
+
+          sudo apt install doxygen
+          doxygen Doxyfile
+      - name: Deploy Docs
+        uses: peaceiris/actions-gh-pages@v3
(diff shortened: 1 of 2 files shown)


timothypidashev (approved):

dmadison:
It would be helpful if the current documentation was at a stable location rather than being sectioned by version. That way the published documentation is always up to date and external links don't need to be updated when new versions are published.

5chmidti:
With every release `fastled.io/docs` would point to the newest release docs. I'm not sure if older docs would be useful, so just having a single version at this path could be done as well.

dmadison:
Unfortunately the redirect at the `/docs/` endpoint only works if you're going to the index page. If you're trying to link to a specific page or a specific item the redirect is not helpful. Doxygen's external documentation implementation also requires linking to the exact HTML folder which makes the version-specific docs a pain.

I'd agree that version-specific docs would not be that useful, with a possible exception for major releases.

5chmidti:
You're right, it's just a redirect...
I agree that a single version at `/docs/` or `/docs/latest/` might be best then.

5chmidti:
The docs are now placed directly into `docs/`, overwriting the previous versions docs. Thanks @dmadison .

I'd suggest that the user can locally checkout their required older release and generate docs for that if needed. There shouldn't be too many big changes that would warrant having older versions, at least right now.
But we could have versions just 'live' in the docs folder (like `docs/3.1` or in some special extra folder). We could link to these versions in the wiki, if that is what the maintainers would like.

For deployment: Right now the docs folder has an index.html, that would get overwritten, but the folder 3.1 would not. Before (or affter) generating new docs via a release the 3.1 folder should get removed.

dmadison:
Is there anything preventing this from being merged? I'd like to spend some time improving the code documentation and it would be great to have the online references up to date.

dmadison:
I believe this can be closed. With #1377 and #1506 I think this is now redundant.

PR #1284, opened 2021-08-21
run build action in parallel per board

5chmidti:
Reduce the time the build action takes by running the compile job for each board in parallel.
I.e. for each board have one job/worker compile all examples. This does not create multiple action runs, each build action just has multiple (right now 6) jobs.
Testing resulted in a time of 4m 40s - 6m 30s and a billable time of around 15m - 19m (in line with current time) (my newest test run doesn't show billable time when I click on the action, did they change something?).

Edit: Cool, you can actually see it in effect.


kriegsman:
It looks like I'm seeing some build checks fail on "yun", but that may be unrelated.  I'm taking a look.

zackees:
I've done this change and much more for the build system. So this is not very relevant anymore.

PR #1297, opened 2021-09-18
encapsulate CLEDControllers' list

5chmidti:
Currently `CFastLED` (and other users) manages iterating through the list of `CLEDControllers` itself. This pr encaspulates the list part of `CLEDController` and adds support for range-for (see changes).

I'm using an `uint8_t` to save the list size instead of calculating it each time. This limits the size to 255 controllers.

- Should the size by calculated each time and thus be only limited by the memory size (i.e. how many controllers can fit in memory)?
- Should the size use a larger type?
- Yes, this is an indirection, but the list class being defined in the header makes the code ready for inlining. Even if it were in a .cpp file, the overhead would be minimal.
- Made the `virtual int size()` function of `CLEDController` `const` (and all of their overridden implementations in controllers; also marked `override`)
- `CLEDController::list()` can be used to access the list

### DIFF
--- a/src/FastLED.cpp
+++ b/src/FastLED.cpp
@@ -12,9 +12,9 @@ void *pSmartMatrix = NULL;
 
 CFastLED FastLED;
 
-CLEDController *CLEDController::m_pHead = NULL;
-CLEDController *CLEDController::m_pTail = NULL;
 static uint32_t lastshow = 0;
+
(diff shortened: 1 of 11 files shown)


zackees:
There's no benefit to this cl other than personal preference not to used linked lists.

PR #1305, opened 2021-10-07
Reduce build warnings [please]

henrygab:
Version information is being output as a build warning (or message).  Either way, this is not desired.  Defining `FASTLED_INTERNAL` is not good, as it disables useful warnings (such as lack of hardware SPI pins):

https://github.com/FastLED/FastLED/blob/5cc17b2be88982eb34b1998de22b83fee56d4f07/src/fastspi.h#L143-L149

Therefore, define new symbol `FASTLED_NO_VERSION_MESSAGE` whose sole purpose is to disable this spurious warning, without disabling true warnings.

### DIFF
--- a/src/FastLED.h
+++ b/src/FastLED.h
@@ -9,7 +9,7 @@
 #endif
 
 #define FASTLED_VERSION 3004000
-#if !defined(FASTLED_INTERNAL) || defined(FASTLED_NO_VERSION_MESSAGE)
+#if defined(FASTLED_SHOW_VERSION_MESSAGE)
 #  ifdef FASTLED_HAS_PRAGMA_MESSAGE
 #    pragma message "FastLED version 3.004.000"
 #  else
--- a/src/FastLED.h
+++ b/src/FastLED.h
@@ -9,7 +9,7 @@
 #endif
 
 #define FASTLED_VERSION 3004000
-#ifndef FASTLED_INTERNAL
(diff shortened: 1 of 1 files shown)


kriegsman:
Hey @henrygab, Happy New Year, and thank you for this contribution.  

I think that FastLED doesn't need the version message by default any more.  What do you think of the idea of changing this to a symbol called something like "FASTLED_SHOW_VERSION_MESSAGE" which is undefined by default, but people can #define it before they include FastLED.h if they need to print the version message for some reason?

Basically, I think we're done needing the version message for most users most of the time, and I'd love to make it go away, but still be available when needed for diagnostics.

If you'd change this PR to do that, I'll merge it right away and release it.

Thanks again for all your help with FastLED!

henrygab:
Hi Mark @kriegsman, 
Happy New Year to you as well.

Yes, I think it's better to avoid the output by default also.  I've updated the PR with the new symbol you recommended.

Thanks!

henrygab:
> If you'd change this PR to do that, I'll merge it right away and release it.

Hi Mark (@kriegsman), I hope real life is swell for you.  I'd updated the PR as per your request.  Is this something still of interest to you?  It reduces build noise.

Thanks!

henrygab:
This was apparently closed in 4d73cdd, by requiring that `FASTLED_SHOW_VERSION` be defined to output the old message.  I will now close this PR.

PR #1306, opened 2021-10-12
Fix WS2812B timing on Teensy 4.x

PaulStoffregen:
Fixes multiple timing issues with Teensy 4.0 & 4.1.

Issue first reported on this forum thread:
https://forum.pjrc.com/threads/68437-FastLED-SK6805-Teensy-4-1-and-a-Teensy-rookie-full-white-only

### DIFF
--- a/src/platforms/arm/mxrt1062/clockless_arm_mxrt1062.h
+++ b/src/platforms/arm/mxrt1062/clockless_arm_mxrt1062.h
@@ -56,24 +56,24 @@ protected:
 			next_mark = ARM_DWT_CYCCNT + off[0];
 			FastPin<DATA_PIN>::hi();
 			if(b&0x80) {
-				while((next_mark - ARM_DWT_CYCCNT) > off[1]);
+				while((next_mark - ARM_DWT_CYCCNT) > off[2]);
 				FastPin<DATA_PIN>::lo();
 			} else {
-				while((next_mark - ARM_DWT_CYCCNT) > off[2]);
+				while((next_mark - ARM_DWT_CYCCNT) > off[1]);
 				FastPin<DATA_PIN>::lo();
 			}
 			b <<= 1;
 		}
 
 		while(ARM_DWT_CYCCNT < next_mark);
-		next_mark = ARM_DWT_CYCCNT + off[1];
+		next_mark = ARM_DWT_CYCCNT + off[0];
 		FastPin<DATA_PIN>::hi();
 
 		if(b&0x80) {
 			while((next_mark - ARM_DWT_CYCCNT) > off[2]);
 			FastPin<DATA_PIN>::lo();
 		} else {
-			while((next_mark - ARM_DWT_CYCCNT) > off[2]);
+			while((next_mark - ARM_DWT_CYCCNT) > off[1]);
 			FastPin<DATA_PIN>::lo();
 		}
 	}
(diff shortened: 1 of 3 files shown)


PaulStoffregen:
@samguyer - Any chance I could talk you into merging these fixes?  They're bracketed by ifdefs which should prevent any change to non-Teensy boards.  If it would help, I could send you test hardware.

samguyer:
@PaulStoffregen I'm not super-familiar with the Teensy devices, but I'd like to get these improvements merged in. Can you give me a quick rundown of what the changes do?

PaulStoffregen:
It's 2 commits.

First commit https://github.com/FastLED/FastLED/pull/1306/commits/8a82b609864146c25693da6738564e1aa4febf91 fixes the clockless driver.  It was just using the wrong timing parameters in several places.  No actual change is made to the algorithm, just which of the array of timing parameters are used.  The last fix of that commit fixes the off time wait, where a macro which computes based on nanoseconds was used with a define that is in units of microseconds.  Easy fix, just multiply by 1000 to convert microseconds to nanoseconds.  Like all the clockless drivers, this code is only compiled if the board uses that platform, so it should not be able to affect any other boards.

The second commit https://github.com/FastLED/FastLED/pull/1306/commits/0ea8c7c2ff805a914ff5dbc25f7d12183e690c3c is to the driver for use of OctoWS2811.  It's basically just a copy of the same K20 driver, but copies the pixel data differently because a simpler format is used on this platform.  This stuff is only compiled if the Arduino sketch defines USE_OCTOWS2811 before including FastLED.h.

If you'd like some free hardware for future testing, I'd be happy to send you a Teensy 4.1 on a little carrier board with some RGB and RGBW WS2812 LEDs, so you have a way to run this code.

samguyer:
@PaulStoffregen This looks great. I'm going to merge it!

I really need to get on board with the Teensy platform -- it sounds amazing. I've been using the ESP boards for a while because they are so cheap, but that there's a cost to cheap!

PaulStoffregen:
Email me directly (paul at pjrc dot dom) or DM on Twitter and I'll get the process started to send you hardware next week.

marmilicious:
@samguyer I don't think you'll be disappointed. I have only used Teensy 3.0 and 3.2 boards but they are excellent.

@PaulStoffregen Thank you for your always awesome support Paul.

ssilverman:
@samguyer @PaulStoffregen Not cool that you took the octows2811_controller.h fix without giving me credit. See: https://github.com/FastLED/FastLED/pull/1309

ssilverman (commented):
@samguyer @PaulStoffregen Not cool that this commit was taken without giving me credit. See: https://github.com/FastLED/FastLED/pull/1309

ssilverman:
@samguyer any way you can revert this PR, split it into two commits and take my fix instead so that I'm in the commit history? The OctoWs2811 fix was mine in the first place.

ssilverman:
On record, this was originally a timing fix without my OctoWS2811 fix for Teensy 4. My fix was created On Oct 16, 2021 and finalized on Oct 19.

ssilverman:
@samguyer I have a solution. I'll update my PR here: https://github.com/FastLED/FastLED/pull/1309

PaulStoffregen:
Opps, sorry @ssilverman, my mistake.  Hopefully we can find a way to properly attribute your work.

ssilverman:
Thanks, @PaulStoffregen, for the acknowledgement. I think if my PR at https://github.com/FastLED/FastLED/pull/1309 is taken in then I’ll be happy. It adds my commit with the proper date and also properly splits out the “timing fix” from the OctoWS2811 fix (and keeping the tree intact, of course; no force pushing involved, so no mess).

Issue #1300, opened 2021-09-26
FastLED Not working with New Board Manager 2.0.0


Jpipe001:
I am working on an esp32 Dev Module (Wroom-32) with Arduino IDE 1.8.16 with the latest version of Board Manger 2.0.0.
FastLED 3.4.0 does not compile correctly and LEDs don't work.
My solution is to go back to previous version of Board Manger 1.0.6.
This now works.

![Board_Manager](https://user-images.githubusercontent.com/24758833/134822897-178f5cfc-f8a6-4341-9469-d029533ba5f0.jpg)

podaen:
Are You also using neopixel and fastled together? If yes which versions do have installed? I have rolleback my IDF like you did to 1.0.6... but it doesn't work for me.

Jpipe001:
Actually, although I rolled back the board manager and got the Fastled working with WS2812 leds. I still seem to be having issues.  I am just using Fastled and not Adafruit Neopixels.  These are my libraries:

#include <Arduino.h>
#include <FastLED.h>
#include <WiFi.h>
#include <WiFiMulti.h>
WiFiMulti wifiMulti;
#include <HTTPClient.h>
#include <WiFiClientSecure.h>
#include "time.h"
#include "ESPmDNS.h"

I am pretty sure that the problem is in the Fastled Library and this issue needs to be looked into.

ATLTVHEAD:
I'm having similar problems. When you say the board manager are you referring to the Esp32 package within the Arduino Boards Manager, or the Arduino Boards Manager itself. If that is the case, how does one do that? 

Sorry for all the questions!

Jpipe001:
A picture speaks 1000 words:
![BoardManager](https://user-images.githubusercontent.com/24758833/135098657-47c398e4-59e9-4864-b10f-4b842257da84.jpg)

ATLTVHEAD:
Thank you so much! @Jpipe001

podaen:
It is a bit strang for me, because I had it working on IDF 2.0.0 but I had neopixel also running and I had to downgrade it, Soo I looked to use the newest versions of the librarys. Now I got IDF 1.0.6 and not the lastest neopixel. (see my topic) Fastled is the newest.

podaen:
Like I said, I used the all latest... except the Neopixel library, and it worked for me. Note, that I am not yet implemented the Wi-Fi for now. You need a task manager to accomplish that.

podaen:
[...]

PR #1308, opened 2021-10-15
esp32: Fix unitialized flags field in init of the RMT driver

alorbach:
The flags field was not initialized to 0 (No flags) which caused
random failures of the esp32 RMT driver (Depending on what
was in the unitialized memory).

Also added support for debugging output in ESP32 RMT Driver
(Disabled by default).

related: https://github.com/FastLED/FastLED/issues/1300

### DIFF
--- a/src/platforms/esp/32/clockless_rmt_esp32.cpp
+++ b/src/platforms/esp/32/clockless_rmt_esp32.cpp
@@ -108,12 +108,14 @@ uint8_t * ESP32RMTController::getPixelBuffer(int size_in_bytes)
 void ESP32RMTController::init(gpio_num_t pin)
 {
     if (gInitialized) return;
+    esp_err_t espErr = ESP_OK;
 
     for (int i = 0; i < gMaxChannel; i += gMemBlocks) {
         gOnChannel[i] = NULL;
 
         // -- RMT configuration for transmission
         rmt_config_t rmt_tx;
+        memset(&rmt_tx, 0, sizeof(rmt_config_t));
         rmt_tx.channel = rmt_channel_t(i);
         rmt_tx.rmt_mode = RMT_MODE_TX;
         rmt_tx.gpio_num = pin;
@@ -126,7 +128,8 @@ void ESP32RMTController::init(gpio_num_t pin)
         rmt_tx.tx_config.idle_output_en = true;
 
         // -- Apply the configuration
-        rmt_config(&rmt_tx);
+        espErr = rmt_config(&rmt_tx);
+        FASTLED_DEBUG("rmt_config result: %d", espErr);
 
         if (FASTLED_RMT_BUILTIN_DRIVER) {
             rmt_driver_install(rmt_channel_t(i), 0, 0);
@@ -134,7 +137,8 @@ void ESP32RMTController::init(gpio_num_t pin)
             // -- Set up the RMT to send 32 bits of the pulse buffer and then
             //    generate an interrupt. When we get this interrupt we
             //    fill the other part in preparation (like double-buffering)
-            rmt_set_tx_thr_intr_en(rmt_channel_t(i), true, PULSES_PER_FILL);
+            espErr = rmt_set_tx_thr_intr_en(rmt_channel_t(i), true, PULSES_PER_FILL);
+            FASTLED_DEBUG("rmt_set_tx_thr_intr_en result: %d", espErr);
         }
     }
 
@@ -232,6 +236,7 @@ void IRAM_ATTR ESP32RMTController::startNext(int channel)
 //    for it to finish.
 void IRAM_ATTR ESP32RMTController::startOnChannel(int channel)
 {
+    esp_err_t espErr = ESP_OK;
     // -- Assign this channel and configure the RMT
     mRMT_channel = rmt_channel_t(channel);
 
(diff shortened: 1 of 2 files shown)


samguyer:
@alorbach Thank you so much for identifying this problem. One issue: the rmt_tx.flags part doesn't seem to compile in our commit checks. Does it compile under the old ESP core?

urish:
@samguyer note that there's an alternative way to achieve a similar effect, without depending on the existence of the `flags` field:

```cpp
rmt_config_t rmt_tx = {}:
```

As far as I can tell, this initializes all the fields to 0 (including flags).

samguyer:
@urish OK, great, thanks! I figured there was some way to do this in a newer version of C++. Do you know if this syntax works on older versions of the Arduino IDE?

henrygab:
> Do you know if this syntax works on older versions of the Arduino IDE?

Both [`C++`](https://en.cppreference.com/w/cpp/language/initialization)  and [`C`](https://en.cppreference.com/w/c/language/struct_initialization) requires this to be supported. 

In `C`, non-specified values are zero-initialized.  In `C++`, it's a bit more complex, but presuming `rmt_config_t` is an aggregate type, and the field is an integral type, it's pretty safe to just consider it as being zero.

Or, if you don't trust the compilers conform to the standards, then you'd need to manually review the output of the initialization....

urish:
The definition of [rmt_config_t](https://github.com/espressif/esp-idf/blob/7bd93e3362e212b1e27fbd19a7a45cbe8e0b46ff/components/driver/include/driver/rmt.h#L67-L81):

```cpp
/**
* @brief Data struct of RMT configure parameters
*/
typedef struct {
    rmt_mode_t rmt_mode;   /*!< RMT mode: transmitter or receiver */
    rmt_channel_t channel; /*!< RMT channel */
    gpio_num_t gpio_num;   /*!< RMT GPIO number */
    uint8_t clk_div;       /*!< RMT channel counter divider */
    uint8_t mem_block_num; /*!< RMT memory block number */
    uint32_t flags;        /*!< RMT channel extra configurations, OR'd with RMT_CHANNEL_FLAGS_[*] */
    union {
        rmt_tx_config_t tx_config; /*!< RMT TX parameter */
        rmt_rx_config_t rx_config; /*!< RMT RX parameter */
    };
} rmt_config_t;
```

And with older SDK, 3.0 ([link](https://github.com/espressif/esp-idf/blob/v3.0/components/driver/include/driver/rmt.h#L105-L118)):

```cpp
/**
 * @brief Data struct of RMT configure parameters
 */
typedef struct {
    rmt_mode_t rmt_mode;               /*!< RMT mode: transmitter or receiver */
    rmt_channel_t channel;             /*!< RMT channel */
    uint8_t clk_div;                   /*!< RMT channel counter divider */
    gpio_num_t gpio_num;               /*!< RMT GPIO number */
    uint8_t mem_block_num;             /*!< RMT memory block number */
    union{
        rmt_tx_config_t tx_config;     /*!< RMT TX parameter */
        rmt_rx_config_t rx_config;     /*!< RMT RX parameter */
    };
} rmt_config_t;
```

samguyer:
@henrygab Ah ok, thank you. It's been a while since I thought about initializers. I would probably have gone old-school -- with memset. 

@urish Yeah, I saw that they added that flags field. I haven't lookup much (yet) at what those flags can do.

@alorbach Would you mind modifying your code to use the = {} initialization? Then we can run the commit checks and get this code in the repo.

henrygab:
That's definitely aggregate initialization.  For the union, only the first listed item (`tx_config`) gets initialized.

Thus, if `sizeof(rmt_tx_config_t) >= sizeof(rmt_rx_config_t)`, then all fields will be initialized.  However, depending on what is `#defined`, that may not be true.  

Thus, I'd recommend proper initialization to zero via `memset(&rmt_tx, 0, sizeof(rmt_config_t))`.  It's not as pretty, and might be redundant, but if the compiler optimizes properly, it should not be twice-initialized.

alorbach:
> @henrygab Ah ok, thank you. It's been a while since I thought about initializers. I would probably have gone old-school -- with memset.
> 
> @urish Yeah, I saw that they added that flags field. I haven't lookup much (yet) at what those flags can do.
> 
> @alorbach Would you mind modifying your code to use the = {} initialization? Then we can run the commit checks and get this code in the repo.

Sure no problem, I will amend my commit

urish:
@alorbach these are the flags that are currently available:

```
#define RMT_CHANNEL_FLAGS_AWARE_DFS (1 << 0) /*!< Channel can work during APB clock scaling */
#define RMT_CHANNEL_FLAGS_INVERT_SIG (1 << 1) /*!< Invert RMT signal */
```

Quoting the docs, which explain why setting the `RMT_CHANNEL_FLAGS_AWARE_DFS` can mess up with the timing of the WS2812 signals:

> * Extra miscellaneous parameters for the channel can be set in the **flags**.
> 
>     * When **RMT_CHANNEL_FLAGS_AWARE_DFS** is set, RMT channel will take REF_TICK or XTAL as source clock. The benefit is, RMT channel can continue work even when APB clock is changing. See :doc:`power_management <../system/power_management>` for more information.
    * When **RMT_CHANNEL_FLAGS_INVERT_SIG** is set, the driver will invert the RMT signal sending to or receiving from the channel. It just works like an external inverter connected to the GPIO of certain RMT channel.
> 
> * A **clock divider**, that will determine the range of pulse length generated by the RMT transmitter or discriminated by the receiver. Selected by setting **clk_div** to a value within [1 .. 255] range. The RMT source clock is typically APB CLK, 80Mhz by default. But when **RMT_CHANNEL_FLAGS_AWARE_DFS** is set in **flags**, RMT source clock is changed to REF_TICK or XTAL.

henrygab:
> @alorbach Would you mind modifying your code to use the = {} initialization? Then we can run the commit checks and get this code in the repo.

Actually, using brace-init is not guaranteed to initialize the entirety of the structure.

Depending on which symbols are defined, `sizeof(rmt_tx_config_t)` might be larger or smaller than `sizeof(rmt_rx_config_t)`.  Only when `sizeof(rmt_tx_config_t)`  is **_larger_** will brace-initialization ensure the full structure is set to defaults / zero.  (This is because both those fields are in a union, and only the first member of the union gets initialized.)

Thus, I actually recommended the use of `memset()` in this case.  If you are not concerned about the initialization of the `rmt_rx_config_t`, then the brace-init will work fine.

alorbach:
@henrygab: I have changed the initialization tp 0 with memset as recommended.

alorbach:
I have fixed the merge conflicts, can this now be merged? 
Without the FIX I doubt that anybody can use FastLED successfully with newer ESP32 Arduino Core (2.0.0).

henrygab (approved):

ckohrt:
When will this be delivered?
Thanks

samguyer:
@alorbach and @henrygab I am running the merge test. As soon as that comes back I'll merge it in!
Thank you for your work on this problem.

PR #1309, opened 2021-10-16
Fix OctoWS2811 controller for Teensy 4

ssilverman:
The pixel layout is different.


ssilverman:
I need to do more testing.

PaulStoffregen:
I tested it here with a Teensy 4.0, can confirm it this works.

ssilverman:
@samguyer @PaulStoffregen Not cool that you took this PR without giving me credit.

ssilverman:
I just force pushed changes that reference this PR in the source code.

ssilverman:
@samguyer I just added two commits that properly separate out the non-timing related changes from the commit that says "Fix WS2812B timing on Teensy 4.x". It cleans up the commits and adds my original fix. My Teensy 4 OctoWS2811 was separate and erroneously included there. I've corrected this. It also ensures my commit is in the history.

ssilverman:
I’ll add: this doesn’t muck with the tree in any way. No force pushing or anything when the merge is done. This should apply cleanly.

samguyer:
@ssilverman Apologies for taking these fixes without properly crediting you! I will merge in this PR, which will hopefully make everyone happy. We appreciate all of the great contributions from our users, but sometimes we are overloaded with PRs and some of them fall through the cracks.

ssilverman:
Thank you. I appreciate the resolution.

PR #1318, opened 2021-11-05
RGBW - Update pixel types

jheanley:
Initial commit - leaves original rgb functionality, implementing rgbw leds still to follow

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@
 /.vscode
 /docs/html
 /docs/latex
+
+*.pri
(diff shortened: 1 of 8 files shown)


jheanley:
I created a fully working version a while ago, but a disagreement between windows and linux meant my changes were obliterated.
Recreating from scratch, so may take a while between work and kids.

semka95:
@jheanley, as far as I understand, this PR works only with avr?

zackees:
I've thought a lot about this and here are my thoughts.

RGB8 is good enough for game developers and this desire to add an extra byte for the W component doesn't sit well with me. Additionally, all our blending algorithms are designed for CRGB data. And there's just not a lot of blending algorithms for RGBW anyway.

I have a conversion function that will convert RGB -> RGBW using a light subtraction algorithm (equal amounts of color are subtracted from the color output and given to the W component at 3x units).

I think this is the way to go. All our blending algorithms will work with it too.

zackees:
RGBW feature is now live using a the approach i listed above. So closing this PR.

PR #1321, opened 2021-11-18
Declare header files for PlatformIO

ivankravets:
Hi,

Please accept this PR that adds header files according to https://docs.platformio.org/en/latest/librarymanager/config.html#headers

Developers will see them soon in the new version of PlatformIO Package Registry.

Thanks in advance!


kriegsman:
Thank you very much for the contribution, and the help with FastLED!

PR #1324, opened 2021-12-07
Add lightness LUT support for AP102 and SK9822

maruel:
This changes the color intensity to use perceptual lightness when
LED_USE_GLOBAL_BRIGHTNESS=2 is defined.

This increases the dynamic range by orders of magnitude and makes it
much more pleasant at low intensity, even compared to
LED_USE_GLOBAL_BRIGHTNESS=1. The dynamic range is calculated by taking
the per channel intensity into account. This leads to a 256\*3\*2 bytes
RAM overhead so it cannot be used by default but it's fine for more
powerful devices like a esp8266 or ESP32. I personally tested on a
esp8266 running esphome 2021.11.4 with 150 LEDs. Here's the stats:

  LED_USE_GLOBAL_BRIGHTNESS=0 (default):
    RAM:   [====      ]  38.2% (used 31300 bytes from 81920 bytes)
    Flash: [====      ]  40.6% (used 423688 bytes from 1044464 bytes)

  LED_USE_GLOBAL_BRIGHTNESS=2:
    RAM:   [====      ]  40.1% (used 32836 bytes from 81920 bytes)
    Flash: [====      ]  40.6% (used 424168 bytes from 1044464 bytes)

  RAM:   +1536
  Flash:  +480

Since with this the scaling is done inside the chipset controller, add
loadByte0/1/2() methods to PixelController for gcc.

This is based on the algorithm I wrote for periph.io/x/devices/apa102 a
few years ago. My algorithm there includes temperature in addition to
intensity correction in the ramp. It does provides a slightly better
ramp but doing this here would require a lot more surgery so starting
smaller.

The transitions aren't as nice as they should where the library seems to
be doing a lot of calculation in PixelController that doesn't match to
the resulting lightness generated by the LEDs. Still, the end result
makes a Christmas tree with such LEDs *much more* enjoyable at low
intensity than even LED_USE_GLOBAL_BRIGHTNESS=1.

Copy pasted to SK9822 but only physically tested on APA102. A follow up
should consider regrouping the code.

Another possible follow up is to use static members for LUT, assuming a
use case of single threaded and multiple APA102Controller instances. I
think it's a niche case so not doing this here.

The LightnessLUT struct assumes memory is zero-initialized.

### DIFF
--- a/src/chipsets.h
+++ b/src/chipsets.h
@@ -2,6 +2,7 @@
 #define __INC_CHIPSETS_H
 
 #include "FastLED.h"
+#include "lightness_lut.h"
 #include "pixeltypes.h"
 
 ///@file chipsets.h
@@ -242,13 +243,74 @@ protected:
 		s0 = (maxBrightness * s0 + (brightness >> 1)) / brightness;
 		s1 = (maxBrightness * s1 + (brightness >> 1)) / brightness;
 		s2 = (maxBrightness * s2 + (brightness >> 1)) / brightness;
+#elif FASTLED_USE_GLOBAL_BRIGHTNESS == 2
+		// Calculate the lightness lookup table for each channel.
+		lutr.init(s0);
+		if (s0 == s1) {
+			lutg.init(lutr);
+		} else {
+			lutg.init(s1);
+		}
+		if (s0 == s2) {
+			lutb.init(lutr);
+		} else if (s1 == s2) {
+			lutb.init(lutg);
+		} else {
+			lutb.init(s2);
+		}
 #else
 		const uint8_t brightness = 0x1F;
 #endif
 
 		startBoundary();
 		while (pixels.has(1)) {
+#if FASTLED_USE_GLOBAL_BRIGHTNESS == 2
+			// On this device, we can get 13 bits of range by using both global 5 bits
+			// PWM and the per channel 8 bits PWM.
+			//
+			// Each channel duty cycle ramps from 100% to 1/(31*255) == 1/7905.
+			//
+			// The goal is to use intensity != 0x1f as much as possible since the
+			// global PWM has a lower frequency than the per channel one.
+			//
+			// To use the full range, we need to discard PixelController's scaling
+			// support since we do not need it.
+			//
+			// Convert each channel to a range between 0 and 7905 using a non-linear
+			// look up table (LUT). This is because the human eye perception is
+			// non-linear.
+			const uint16_t lr = lutr.data[pixels.loadByte0()];
+			const uint16_t lg = lutg.data[pixels.loadByte1()];
+			const uint16_t lb = lutb.data[pixels.loadByte2()];
+			const uint16_t m = lr | lg | lb;
+			// Maps the computed colors on a range [0, 255*31] back to global
+			// intensity + per channel intensity.
+			uint8_t brightness, pr, pb, pg;
+			if (m <= 255) {
+				brightness = 1;
+				pr = uint8_t(lr);
+				pg = uint8_t(lg);
+				pb = uint8_t(lb);
+			} else if (m <= 511) {
+				brightness = 2;
+				pr = uint8_t(lr>>1);
+				pg = uint8_t(lg>>1);
+				pb = uint8_t(lb>>1);
+			} else if (m <= 1023) {
+				brightness = 4;
+				pr = uint8_t((lr+2)>>2);
+				pg = uint8_t((lg+2)>>2);
+				pb = uint8_t((lb+2)>>2);
+			} else {
+				brightness = 0x1f;
+				pr = uint8_t((lr+15)/31);
+				pg = uint8_t((lg+15)/31);
+				pb = uint8_t((lb+15)/31);
+			}
+			writeLed(brightness, pr, pg, pb);
+#else
 			writeLed(brightness, pixels.loadAndScale0(0, s0), pixels.loadAndScale1(0, s1), pixels.loadAndScale2(0, s2));
+#endif
 			pixels.stepDithering();
 			pixels.advanceData();
 		}
@@ -258,6 +320,11 @@ protected:
 		mSPI.release();
 	}
 
+#if FASTLED_USE_GLOBAL_BRIGHTNESS == 2
+	LightnessLUT<255*31> lutr;
+	LightnessLUT<255*31> lutg;
+	LightnessLUT<255*31> lutb;
+#endif
 };
 
 /// SK9822 controller class.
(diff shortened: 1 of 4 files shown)


zackees:
Wow, came to FastLED seeing if this was implemented and here is a pull request. I will be using this immediately when it get's merged and I hope this can be rolled into a release pretty fast so that I can use it on platformio by changing the platform.ini to something like:

lib_deps = FastLED@3.4.1

zackees:
FYI warning: comparison between signed and unsigned integer 

My gcc settings are:
  -Wall
  -Werror=return-type

######

In file included from .pio\libdeps\teensy36\FastLED\src/chipsets.h:5:0,
                 from .pio\libdeps\teensy36\FastLED\src/FastLED.h:66,
                 from src\src\gfx\FastLED.h:4,
                 from src\src\gfx\gfx.h:9,
                 from src\src\gfx\gfx.cpp:5:
.pio\libdeps\teensy36\FastLED\src/lightness_lut.h: In instantiation of 'void LightnessLUT<maxOut>::init(uint8_t) [with short unsigned int maxOut = 7905u; uint8_t = unsigned char]':
.pio\libdeps\teensy36\FastLED\src/chipsets.h:378:12:   required from 'void SK9822Controller<DATA_PIN, CLOCK_PIN, RGB_ORDER, SPI_SPEED>::showPixels(PixelController<RGB_ORDER>&) [with unsigned char DATA_PIN = 1u; unsigned char CLOCK_PIN = 0u; EOrder RGB_ORDER = (EOrder)136u; long unsigned int SPI_SPEED = 24ul]'
src\src\gfx\gfx.cpp:159:1:   required from here
.pio\libdeps\teensy36\FastLED\src/lightness_lut.h:31:10: warning: comparison between signed and unsigned integer expressions [-Wsign-compare]
    if (l < linearCutOff) {
          ^

zackees:
Major problems with this code right now. While it does have some hdr effects which are pretty cool, the color balance is way off on my SK9822 chipset test

Test code:
void loop() {
  int i = random(NUM_LEDS);
  leds[i]  = CRGB::White;
  FastLED.show();
  leds[i]  = CRGB::Black;
}

...
With FastLED.setBrightness(255); the white flashes are green. With FastLED.setBrightness(128); the white flashes are all blue.

maruel:
Thanks @zackees for testing! I pushed a new commit d4ca518dc1ce3fad3906ddaadc57489c92349cc4 to try to fix the unsigned/signed comparison.

I don't have SK9822 to test locally. I'm a bit baffled about what is happening. I'm currently testing through esphome which makes direct testing like this a bit harder, I would have to make a one off device..

zackees:
SK9822 and APA102 should produce nearly identical color mappings. Why don't you try the code snippet i've provided with APA102, it should produce the same color results.

maruel:
I can't readily test because of the way the device is currently deployed. It's hard for me to flash over serial, so I will have to create a second temporary setup. May have to wait to this weekend.

uberjay:
> I can't readily test because of the way the device is currently deployed. It's hard for me to flash over serial, so I will have to create a second temporary setup. May have to wait to this weekend.

NB: I have not actually tested this, but it does build. I've been meaning to see if esphome was a good way to deploy/test lower-level FastLED programs, just not got around to it yet. 😉

I had to `#undef delay`, otherwise the esphome `delay` macro was colliding with the FastLED `CFastLED::delay` declaration. I failed to figure out how that's avoided with the bundled FastLED, so, ymmv. Unclear what, if anything, that might affect.

Drop a `custom_component.h` file in your esphome root directory (e.g., `/config/esphome`, where `/config` is my Home Assistant config root directory):
```c++
#define FASTLED_ESP32_RAW_PIN_ORDER
#define FASTLED_RMT_BUILTIN_DRIVER true
#define FASTLED_INTERNAL

#undef delay
#include "FastLED.h"

const uint8_t NUM_LEDS = 10;
const uint8_t DATA_PIN = 12;
const uint8_t CLOCK_PIN = 13;

class CustomComponent : public Component {
    CRGBArray<NUM_LEDS> leds;

public:
    void setup() override {
        FastLED.addLeds<APA102, DATA_PIN, CLOCK_PIN>(leds, NUM_LEDS);
        FastLED.setBrightness(255);
    }
    void loop() override {
        int i = random(NUM_LEDS);
        leds[i] = CRGB::White;
        FastLED.show();
        leds[i] = CRGB::Black;
    }
};
```

And then, configure an esphome node to use the pr-associated-branch of FastLED and the custom component:

```yaml
esphome:
  name: ...
  platform: ESP32
  board: esp32dev
  libraries:
   - fastled=https://github.com/maruel/FastLED#lut
  includes:
    - custom_component.h

custom_component:
- lambda: |-
    auto my_custom = new CustomComponent();
    return {my_custom};
```

uberjay:
Sigh, sorry, did not mean to close this. 😭

maruel:
Thanks @uberjay ! Note that the following doesn't work anymore:
```
  libraries:
   - fastled=https://github.com/maruel/FastLED#lut
 ```
because of https://github.com/esphome/esphome/blob/dev/esphome/components/fastled_base/__init__.py#L47

I tested it locally on 150 apa102 and the LEDs are flashing white as expected.

**Comment by maruel** [comment]
@zackees I'm sorry it doesn't work well with the sk9822. What about I remove the code for this driver and make LED_USE_GLOBAL_BRIGHTNESS=2 use the same code as LED_USE_GLOBAL_BRIGHTNESS=1?

**Comment by zackees** [comment]
Okay, we can always use the APA102 fastled driver to drive the SK9822, and this can be used as playground.

**Comment by maruel** [comment]
Done.

File: src/chipsets.h [hunk 1]
```diff
@@ -242,13 +243,74 @@ class APA102Controller : public CPixelLEDController<RGB_ORDER> {
...
+			// To use the full range, we need to discard PixelController's scaling
+			// support since we do not need it.
+			//
+			// Convert each channel to a range between 0 and 7905 using a non-linear
+			// look up table (LUT). This is because the human eye perception is
+			// non-linear.
+			const uint16_t lr = lutr.data[pixels.loadByte0()];
+			const uint16_t lg = lutg.data[pixels.loadByte1()];
+			const uint16_t lb = lutb.data[pixels.loadByte2()];
+			const uint16_t m = lr | lg | lg;
```
chadrockey, line 285: Typo here: lr | lg | lb, green used twice

maruel: (!!) thanks

maruel: I rebased on 3.5.0 and force pushed

maruel:
This PR is now stale so I'll close it. I still want to note that running APA102 at 13 bits makes the dynamic range significantly better.

Issue #1329, opened 2021-12-14
ESP32: 200ms flicker every ~17.9 seconds with RMT driver


Jueff:
I discovered an issue that LEDs have short ~200ms flicker every ~17.9 seconds.
I investigated the problem and it turned out, the it has to do with __clock_cycles(). __clock_cycles() returns a 32bit value, each time the 32bit overrun occurs you see a short LED flicker at all LEDs which values change more then 10 times per second.

The sample project below blinks one LED very fast, if you watch the LED you will see the flickering every 17,9 seconds.

Could the problem be caused in the ESP32RMTController::fillNext method?

`if (mLastFill != 0 and now > mLastFill) {`

doesn't handle overrun correctly 

**Code sample for problem reproduction**

```
#include <Arduino.h>

#define FASTLED_INTERNAL       // Disable version number message in FastLED library (looks like an error)
#include <FastLED.h>

const unsigned NUM_OF_LEDS = 256;
CRGB leds[NUM_OF_LEDS];

TaskHandle_t  FASTLedTaskHnd;

void setup() {
    Serial.begin(921600);
    //FastLED.addLeds<WS2812B, 27, EOrder::RBG>(leds, 0, NUM_OF_LEDS);
    CLEDController& controller0 = FastLED.addLeds<NEOPIXEL, 27>(leds+  0, NUM_OF_LEDS);
    controller0.clearLeds(256);
    FastLED.setDither(DISABLE_DITHER);
    
    xTaskCreatePinnedToCore(FASTLedTask,"FASTLed",4096,NULL,1,&FASTLedTaskHnd,0);
}

void FASTLedTask( void * parameter ) {
//--------------------------------
  while(1) {
    yield();
    loopFASTLed();
    delay (1);
  }
}

void loop() 
{
}

auto lastSecond = millis()/1000;

void loopFASTLed()
{
    auto timeBefore = millis();
    leds[0].r = ((timeBefore/30)&0x01)*128;
    FastLED.show();
    auto passedTime = millis() - timeBefore;

    if ((passedTime!=8&&passedTime!=7) || timeBefore/1000!=lastSecond)
    {
      lastSecond = timeBefore/1000;
      Serial.print("millis=");
      Serial.print(timeBefore);
      Serial.print(" cycles=");
      Serial.print(__clock_cycles());
      Serial.print(" show() took ");
      Serial.print(passedTime);
      Serial.println(" ms");
    }
}

```

Jueff:
[...]

PR #1330, opened 2021-12-15
avoid 200ms flicker every ~17.9 seconds with RMT driver

Jueff:
see also issue #1329

### DIFF
--- a/src/platforms/esp/32/clockless_rmt_esp32.cpp
+++ b/src/platforms/esp/32/clockless_rmt_esp32.cpp
@@ -455,9 +455,9 @@ void IRAM_ATTR ESP32RMTController::fillNext(bool check_time)
 {
     uint32_t now = __clock_cycles();
     if (check_time) {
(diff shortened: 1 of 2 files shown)


ssilverman:
Why is the `led_sysdefs_esp32.h` change included?

Jueff:
The  `led_sysdefs_esp32.h` change should not be part of the PR. I unfortunately did another commit to my repo after creating the PR, because I didn't know that this influences the PR.  So I reverted the `led_sysdefs_esp32.h` commit.

---> Jueff **force-pushed** the master branch from 514bdbc to 54c409e 7 days ago 

I thought, that the second commit will then not be part of PR anymore.

only the change in `clockless_rmt_esp32.cpp` is relevant.

Jueff:
closed this PR because it contains extra files not relevant for the merge
replaced with PR #1336

ssilverman:
For your reference, and to make it easier in the future, there’s no need to make a new PR. You could change things locally and then force push. PR’s aren’t set in stone.

PR #1320, opened 2021-11-17
Support ESP32 C3, S3 and possibly H2 variants

maddogjt:
Adding support required for ESP32 C3 & S3 variants, and possibly H2 variant (not able to compile for H2 yet)

- Updated pin mapping mechanism to support multiple variants based on valid output pin defines provide by the ESP-IDF, should automatically support any new chip variants with up to 64 pins.  This also fixes incorrectly masked out pins on ESP32-S2 
- Updated FastPin to support chip variants that only have a single output retister
- Fix compile error on RISC-V based variants
- Updated clockless_rmt_esp32 to support all upcoming variants.  Tested on ESP32 & ESP32-C3, I'm waiting for a ESP32-S2 board to test on.   ESP32-S3 compiles (by hacking up a custom arduino core with support), but is not yet testable.

PR #1332, opened 2021-12-15
Add some defaults to avoid build failure on old platform versions e.g. 1.0.4

Jueff:
see also issue https://github.com/FastLED/FastLED/pull/1320#issuecomment-994761190
and https://github.com/FastLED/FastLED/pull/1320#issuecomment-995114712


Jueff:
closed because duplicate with #1333

PR #1333, opened 2021-12-16
Improve compatibility with older ESP32 Arduino core versions

maddogjt:
Improve compatibily with older versions of ESP32 Arduino core.  Tested with versions 1.0 & higher.


Jueff (commented):
works fine!

samguyer:
@maddogjt Thank you! I'm merging it in now...

PR #1336, opened 2021-12-22
avoid 200ms flicker every ~17.9 seconds with RMT driver

Jueff:
see also issue #1329
replace PR #1330 because #1330 contains extra files not relevant for the merge


Panos-Jr (approved):
Exceptional!

PR #1338, opened 2021-12-28
optional increase of DMABuffer count by using a define

Jueff:
In my environment I had frequent flickering issues when using the I2S driver, most likely caused by interrupts of other code parts, e.g. WIFI stack. As the I2S driver is much faster when using multiple Controllers I tried to find the cause of the problem. It turned out the increasing the DMA Buffer count is a good solution.

I changed implementation to optionally set the number of DMA Buffers to a value (2-16) defined with FASTLED_ESP32_I2S_NUM_DMA_BUFFERS preprocessor variable. 

Situation with default DMA buffer count: [Video DefaultDMABuffers](https://www.dropbox.com/s/qxg6a3tw4npnieb/DefaultDMABuffers.mp4?dl=1)

using `#define FASTLED_ESP32_I2S_NUM_DMA_BUFFERS 4 ` 

solved all my flickering issues: [Video FourDMABuffers](https://www.dropbox.com/s/dewpdcy3hxqgtjo/FourDMABuffers.mp4?dl=1)

### DIFF
--- a/src/platforms/esp/32/clockless_i2s_esp32.h
+++ b/src/platforms/esp/32/clockless_i2s_esp32.h
@@ -65,6 +65,19 @@
  * can automatically start on the next buffer.
  */
 /*
+ * The implementation uses two DMA buffers by default. To increase the 
+ * number of DMA buffers set the preprocessor definition 
+ * 
+ * FASTLED_ESP32_I2S_NUM_DMA_BUFFERS 
+ *
+ * to a value between 2 and 16. Increasing the buffer to 4 by adding
+ *
+ * #define FASTLED_ESP32_I2S_NUM_DMA_BUFFERS 4 
+ *
+ * solves flicker issues in combination with interrupts triggered by 
+ * other code parts.
+ */ 
+/*
  * Permission is hereby granted, free of charge, to any person obtaining a copy
  * of this software and associated documentation files (the "Software"), to deal
  * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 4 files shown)


samguyer:
@Jueff This looks like a nice feature for the I2S driver. I don't know this code as well as I know the RMT driver, but would it make sense to make the buffers bigger rather than add more buffers?

Jueff:
I'm not an expert in the I2S code. It looks like one buffer has exactly the size to hold data for one RGB Led for up to 32 channels. Increasing the buffer size would have impact on the way DMA buffers are filled. Especially ending the i2s sending currently is fully coupled to DMA interrupts, changing the the buffer size would need are redesign of "end of stream".

My PR focused on keeping existing code unchanged in case the FASTLED_ESP32_I2S_NUM_DMA_BUFFERS is undefined and also keep existing buffer filling and I2S sending logic completely unchanged. I didn't want break compatibility. And also I don't want to become an I2S expert :-)

Only benefit of changing the size of the single DMA buffers would be a reduction of DMA interrupts, the buffer memory consumption is nearly the same.

samguyer:
Hello! We are looking to merge this PR today, but I just want to check with Yves Bazin, who wrote the I2S driver.
Thanks for your patience!

samguyer:
Hi @Jueff ! Sorry for the delay in reviewing this PR. It looks good, from a code standpoint, and if it helps with flickering issues then I'm all for it! Thanks for your contribution.

PR #1342, opened 2022-01-03
New overloaded function ColorFromPalette what take color from GradientPalette

Palpalych13:
ColorFromPalette from TProgmemRGBGradientPalette
This is a new implementation of a popular ColorFromPalette function. 
It became necessary for me when I began to use gradient palettes from progmem, and
also use a structure of variables with implementations of CRGBPalette16 palettes for different effects.
That all takeout an extra 64 bytes of memory for each effect. (((
Now we can use a simply pointer and... It's works! )))

For declaration of the array of palettes:
	...
	const TProgmemRGBGradientPalettePtr gradientPalettes[] = {
		&Sunset_Real_gp[0],	&es_rivendell_15_gp[0], &es_ocean_breeze_036_gp[0],	&rgi_15_gp[0], ...};
	...

To call someone:
	...
	CRGB color = ColorFromPalette(*gradientPalettes[1], index, brightness, blendType);
	...
	fill_palette(leds, NUM_LEDS, startIndex, incIndex, *gradientPalettes[0], brightness, blendType);
	...

### DIFF
--- a/colorutils.cpp
+++ b/colorutils.cpp
@@ -1047,6 +1047,86 @@ CHSV ColorFromPalette( const struct CHSVPalette256& pal, uint8_t index, uint8_t 
     return hsv;
 }
 
+CRGB ColorFromPalette(const TProgmemRGBGradientPalette_byte &progpal, uint8_t index, uint8_t brightness, TBlendType blendType)
+{
+	TRGBGradientPaletteEntryUnion u, u1; // initiate 2 vars of needed colors
+	uint8_t count = 0;
+	do
+	{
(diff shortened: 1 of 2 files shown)


kriegsman:
Thanks for this contribution - this looks very very useful.  I'm a little concerned about the performance, given that there is -- as you point out -- a cpp division ("/") operation in the main path of the function.  Do you think there's another way that we could achieve the same thing, but without the very heavy delay of long division?

zackees:
64 bytes per effect seems small. Change seems risky.

PR #1350, opened 2022-01-12
Update chipsets.h

Sab78:
Fixed a bug in the LPD6803Controller class. Multiple diodes at the end of the LEDS array did not work correctly, no data was sent from the array to the LPD6803.


kriegsman:
Great find, and great fix.  Thank you for contributing this back, and thanks for your help with FastLED!

JoshNoise:
I had an issue with LPD6803's that i think is being caused by the call to endBoundary(pixels.size()); added on this commit.
When controlling >=32 lpd6083's the first one would just go on static magenta. controlling less than 32 and all works fine.
Commenting out the call to endBoundary(pixels.size()); and all works fine again for me.

PR #1353, opened 2022-01-17
Enable arbitrary pixel remappings

henrygab:
#### THIS IS CURRENTLY A DRAFT PR

However, constructive review and comments / recommendations will be warmly received.

#### Problem -- assumption of memory layout

The problem that was faced was that some of the existing functions in FastLED required a pointer to a `CRGB` or `CHSV` struct.  Unfortunately, this hard-coded the requirement that those structures were contiguously laid out in memory, which means any effect that uses those functions has to double-buffer when the logical layout of the pixels does not match the physical layout (connection order).

#### Fix -- templatize array-based functions

<details><summary>Example (click to expand)</summary><P/>

```C++
void fill_solid( struct CRGB * leds, int numToFill, const struct CRGB& color) {
    for( int i = 0; i < numToFill; ++i) {
        leds[i] = color;
    }
}
```

Becomes:
```C++
template <typename T>
void fill_solid( T& leds, int numToFill, const struct CRGB& color) {
    for( int i = 0; i < numToFill; ++i) {
        leds[i] = color;
    }
}
```

By changing the current function definitions (which take `CRGB*` as first parameter, for example), to instead take a template parameter type for that first parameter, it should both:
A. Work with existing code that passed `CRGB*` ... as 
By adding versions that take a reference to a templated type, it allows any/all of the following to work:

1. Passing `struct CRGB*` -- just works, because `operator[]` is the standard array calculation (backwards compatible)
2. Passing `struct CHSV*` -- just works, because `operator[]` is the standard array calculation (backwards compatible)
3. Passing reference to any class with `PIXEL_TYPE & operator[](size_t idx) const` defined -- works with additional class overhead to access the specified index (e.g., remapping table or remapping function) ... where `PIXEL_TYPE` can be either `struct 
4. CRGB` or `struct CHSV`. 

</details><hr/>

Critically, if the templated function is never called, there is zero impact to final binary.  Win/Win.


henrygab:
This is just a set of notes for which functions I am looking to templatize, in order to remove the implicit assumption that the CRGB objects are contiguously allocated in memory....

#### Working list functions to be considered for an update

For these, the underlying class needs only provide support for `PIXEL_TYPE "& operator[](size_t idx)`.  The functions are generally changed such that a parameter previously explicitly taking `CRGB*` (with implied contiguous memory layout) instead takes a reference to a template parameter (e.g., `T&`).

NOTE: This does mean that `nullptr` cannot be passed as an argument.  I argue that this is a "Good Thing".

##### `colorutils.h`

* [x] `fill_solid` -- add `fill_solid2()` to allow non-zero start index
* [x] `fill_rainbow` -- add `fill_rainbow2()` to allow non-zero start index
* [x] `fill_gradient` -- four template functions (aka `fill_gradient_HSV`)
* [x] `fill_gradient_RGB` -- four template functions

* [] `fadeUsingColor` -- 
* [] `blend` -- two functions
* [] `nblend` -- two functions
* [] `blur1d` -- 
* [] `blur2d` -- 
* [] `blurRows` -- 
* [] `blurColumns` -- 
* [] `fill_palette` -- 
* [] `map_data_into_colors_through_palette` -- 

* [] `fadeLightBy` -- alias of `nscale8_video`
* [] `fade_video` -- alias of `nscale8_video`
* [] `nscale8_video` -- simple per-pixel iteration to `nscale8_video`
* [] `fadeToBlackBy` -- alias of `nscale8`
* [] `fade_raw` -- alias of `nscale8`
* [] `nscale8` -- simple per-pixel iteration to `nscale8`
* [] `napplyGamma_video` -- simple per-pixel iteration to `applyGammaVideo`

5chmidti:
Interesting changes, you do need a reference for the arguments though. Template arguments will not deduce a reference ([example](https://godbolt.org/z/cMP5Yjdqv)).

henrygab:
> Interesting changes, you do need a reference for the arguments though. Template arguments will not deduce a reference ([example](https://godbolt.org/z/cMP5Yjdqv)).

@5chmidti -- **_Thank you_** for the early review!

Yes, you're absolutely right ... I will need to explicitly mark the parameter as a reference, to meet my goal.  Sigh.  The difficulty lies in the wide range of compiler support.   I believe the header `type_traits` is not available for AVR... which prevents the use of `std:enable_if<>` for proper template metaprogramming / specialization.  ☹️ 

Using `fill_solid()` as the simplest example, do you have a recommendation on how to enable 100% backwards-compatibility, while also allowing the function to work with any class that implements `CRGB& operator[](int index)`?

5chmidti:
You can use a [forwarding-reference](https://en.cppreference.com/w/cpp/language/reference#:~:text=Forwarding%20references,a%20function%20argument%2C) to bind both l- and r-values (didn't see there was a failure before because of r-values needing to be accepted as arguments).

```c++
template<typename T>
void fill_solid2( T&& leds, int startPos, int numToFill, const struct CRGB& color)
{
    int end = startPos + numToFill;
    for( int i = startPos; i < end; ++i) {
        leds[i] = color;
    }
}
```

henrygab:
#### hic sunt dracones

The template argument must be a reference, else it receives a **_copy_** of the parameter.  Getting a copy is fine, when the parameter is a pointer to an array or other contiguous memory.  When it's a class that's wrapping an arbitrary remapping, it's at best poor performance, and at worst can cause unexpected behavior.

Previously, `fill_solid()`'s first parameter was `CRGB* leds`, and presumed contiguous layout (e.g., an array).

Thus, the following would be valid:
```C++
CRGB leds[100];
fill_solid(  &(leds[0]), 100, CRGB::White );
```

That first parameter would take the address of the first element of the array.  Pretty standard C++.

However, when using `template <typename T> fill_solid(T& leds, const CRGB& color)`, this would cause an error, because now that parameter must be a reference.  References to non-const values can only be references to `lvalues`.  However, the expression `&(leds[0])` is an `rvalue`.

This is why the compiler throws the following fun error:

```
In constructor 'CHSVPalette16::CHSVPalette16(const CHSV&)':
lib/src/colorutils.h:673:21: error:
    cannot bind non-const lvalue reference of type 'CHSV*&'
    to an rvalue of type 'CHSV*'
        fill_solid( &(entries[0]), 16, c1);
                    ^~~~~~~~~~~~~
```

Oh, the joys of `lvalue` vs. `rvalue`.  Now to look into the concept of "perfect forwarding", or whether we can use an "rvalue reference to type" (`T&&`).  Fun?

henrygab:
> You can use a [forwarding-reference](https://en.cppreference.com/w/cpp/language/reference#:~:text=Forwarding%20references,a%20function%20argument%2C) ...

@5chmidti ... Yes, I am looking into this.  However, AVR fails to include any part of `std::` namespace, even the parts that are just compilation / type_traits / metaprogramming related.  Thus, unless you have alternatives, I think I may need to replicate some bare minimum functionality of `<type_traits>` to ensure perfect forwarding can work:

* [x] remove_reference
* [x] integral_constant
* [x] is_lvalue_reference
* [x] forward
* [ ] others?  TBD

henrygab:
I think the next step is to depend upon [Embedded Template Library](https://github.com/ETLCPP/etl-arduino/), which has done the hard work of enabling STL-like capabilities (including type_traits), but with target of embedded systems (C++98-compatible, no run-time allocation, no heap usage, etc.).

Issue #373, opened 2016-11-19
SK6812 support


eltang:
Would it be possible for this chipset to be supported by FastLED?

kriegsman:
Yep - it already works.   It's a clone of the WS2812, so you can just tell FastLED that you're using WS2812, and the SK6812's will "just work".

eltang:
Is the information on [this page](https://learn.adafruit.com/adafruit-neopixel-uberguide/advanced-coding) inaccurate then?

focalintent:
There are two flavors ofnsk6812 - RGB And RGBW - fastled doesn't supportnRGBW yet (and there's already an open issue for tracking that)

eltang:
Oops, I hadn't noticed that issue. Has any progress been made since June?

eltang:
For your information, I'm going to be getting some SK6812RGBW LEDs along with a Teensy 2.0 soon.  If RGBW LED support is something you're actively working on, I'd be happy to test it for you.

kriegsman:
Thank you.  Sadly, there's a been a death in our family, and we're attending to that first and foremost.  But thank you for the offer of help; we'll come back around to that.

eltang:
Sorry for your loss. I may need to have the bare AVR support we discussed in the other issue before I start testing, though, as I do not want my project to be dependent on the Arduino IDE.

eltang:
I have my hardware now. Just let me know when you'd like me to test something.

Ethanmose:
sk6812 have sk6812 rgbw,sk6812 rgb,sk6812 wwa,sk6812 3535 ,sk6812 white,ws2812b also have ws2812b rgb,ws2812b 3535. and sk6812 is better than ws2812b,tansmission speed more faster.

Blink515:
No it will not work. The colors will not be what you expect.

DarkiStar:
Here 
[Link to FastLED with RGBW](http://www.partsnotincluded.com/neopixels/fastled-rgbw-neopixels-sk6812/)
someone found a workaround for using rgbw with FastLed.

TheCellMC:
any news on this?

coryking:
Just a note that I've added RGBW support to [my fork of FastLED](/coryking/FastLED) -  Works only on ESP8266.  So far, pretty sweet!

PR #1357, opened 2022-01-24
SK6812 is a WS2812 clone.

zark0-UwU:
Apparently the SK6812 is a clone of WS2812 as said in [#373](https://github.com/FastLED/FastLED/issues/373#issuecomment-261682267) by @kriegsman 

I thought about adding a binding for WS2812 as SK6812, but i have no idea where to start, since this is my first time doing anything with this project.

Maybe the idea of a binding for SK6812 is a bad idea if no direct support for this chipset is given, since it may cause some trouble. 

But what is not a bad idea is documenting the fact that SK6812 is a clone of a supported chipset (WS2812), indicating that it may not work perfectly, but if the clone is good enouth it should work flawlessly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ How quickly can you get up and running with the library?  Here's a simple blink 
 Here's a list of all the LED chipsets are supported.  More details on the led chipsets are included [on our wiki page](https://github.com/FastLED/FastLED/wiki/Chipset-reference)
 
 * Adafruit's DotStars - AKA the APA102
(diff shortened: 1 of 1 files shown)


zark0-UwU:
I just realized a binding is not even required, since it autodetects the protocol to use, I was amazed to realize that, but the documentation sitill should mention the chip as a clone of WS2812 as modified in the pull request.

zark0-UwU:
as described [here](https://www.partsnotincluded.com/fastled-rgbw-neopixels-sk6812/) the sk812 does NOT work straight forward, but can work, since it is very similar to the WS2812, I think a profile for sk6812 can be made, I personally do not have the knowledge nor the time to do it ****yet** **, but i hope someone can help

PR #1361, opened 2022-01-31
Clarifies the need for SmartMatrix library and scope limited to Teensy boards

tochinet:
My very first pull request, I hope this is the way to do.


kriegsman:
Hi @tochinet !  Thanks for the contribution and this looks very helpful.  Thanks for the PR and the help with FastLED!

PR #1369, opened 2022-02-27
Prevent -Wunused-but-set-variable in clockless_rmt_esp32.h

sethtroisi:
Otherwise `errcode` is considered set but unused when not in DEBUG.


samguyer:
@sethtroisi Thank you!

PR #1391, opened 2022-05-20
Adding D21/M0 DMA

sheaivey:
I have no expectations this will ever get merged into the mainline as is. However I wound up having a need for DMA rendering to free up the CPU and allow for longer interrupts without causing flickering. I couldn't find any existing FastLED DMA solutions for the SAMD21 based boards other than the Adafruit_NeoPixel_ZeroDMA library.

This is a quick port of [Adafruit_NeoPixel_ZeroDMA](https://github.com/adafruit/Adafruit_NeoPixel_ZeroDMA) but to work with FastLED.

You will need to have the [Adafruit_ZeroDMA](https://github.com/adafruit/Adafruit_ZeroDMA) library.

To enable DMA add `#define FASTLED_ARM_M0_DMA` to the top of your project before the `#include <FastLED.h>` line.

See [src/platforms/arm/d21/dma_pins.h](src/platforms/arm/d21/dma_pins.h) for supported boards and SERCOM pins.

Typically the SPI `MOSI` pin is available on every supported board.


kriegsman:
Thanks for the contribution, and this is a particularly interesting one.  I'm not sure that we want to create the dependency on the ZeroDMA library, but we'll discuss it -- and thank you again for helping experiment and advance things in interesting ways

sheaivey:
> Does it work with all chipsets (including the 400 KHz ones) or is it NeoPixel only?

Yes, you can adjust the DMA speed via the SERCOM configuration if you know which one your board is using.

zackees:
We can't have a dependency on another library that the user is required to install. If you need a library it needs to be ported in and optionally put under it's own namespace so it doesn't have symbol collision in the linker.

sheaivey:
I'll just leave this as is for anyone needing an unofficial solution for the SAMD21 based chips. I have since moved on from using them myself.

Closing this PR. Thanks

PR #1400, opened 2022-06-26
Non-wrapping blend mode for ColorFromPalette

smurphyUSI:
ColorFromPalette using either the 16 or 32 color options (i.e *Palette16 or *Palette32) have implicit "wrapping" of colors near the end of the range, explicitly when using LINEARBLEND.

This allows for colors to be looped over without any noticeable "cuts" in the palette blending.

However, not all users are aware this "wrapping" will occur, and the index position the warp will occur is different depending on which of the 16 or 32 palette options are used.  This effect is a consequence of the bit shifting of the values, which is necessarily different for 16 and 32 color palettes.

Experienced users will manually avoid the end index positions in their code to avoid unintended "wrapping", but most users aren't aware of this feature or the technicalities.

I've added a TBlendType named  LINEARBLEND_NOWRAP, which is the same as LINEARBLEND, but pre-maps the requested indexed color in the palette to avoid the "wrapping" effect.
Since the change is implemented directly into the ColorFromPalette methods, nobody needs be aware of the bit-shifting differences. or manually avoid any index values.

Thank you


marmilicious:
I like this idea. Numerous times over the years people have been confused by this blending feature. (Asking why is their end color isn't working, why it's changing color, etc).  Not needing to stop indexing at 240 (with 16 bit) when one doesn't want wrapping seems like it would make things easier for new users to understand, and also give back having the full range of 255 steps of transition.

kriegsman:
I'm trying to get some CI build issues worked out so that I can merge this PR.  I'm going to Close -- and immediately Reopen this PR, which I hope should clear it for merging.

kriegsman:
Thank you again for this contribution to FastLED!

marmilicious:
Great, thank you @smurphyUSI and Mark.

PR #1403, opened 2022-07-04
Override Default use of HSV2RGB_Rainbow with HSV2RGB_Spectrum

smurphyUSI:
I know that 'Rainbow' is the preferred color model used in FastLED, but it seems that preference was hard code into the RGB operators which take HSV values as inputs.

In the cases where 'Spectrum' would be instead preferred, it would be nice to override the default use of 'Rainbow' with 'Spectrum'.

I tried several methods, including variable assignment to FastLED, and another for sub-classing the HSV class, but they either had downside of having HSV class be 'aware' of the CFastLED class (and complicating complication a bit). or inuring virtual call costs on functions I didn't intend to affect (did not try partial virtual).

I settled on using a #define, which would need to be before the inclusion of either FastLED.h or just anywhere before pixeltypes.h.
It isn't the most convenient, but it has zero run-time cost for anybody who wants to make the switch.

If anybody has a better method for doing this, without killing performance, I'd love to hear it.


kriegsman:
Thanks for working on this and thinking about how we could let people change the default HSV2RGB mapping mechanism.  

Before we commit to this particular implementation, I'd like to propose a further idea:
I have (elsewhere) two other HSV2RGB mappings that are also useful in different contexts.  One is hsv2rgb_maxbright, which maximizes the light output of each hue, even though some hues will be brighter than others, e.g. "purple" is normally something like (128,0,128) so that it has a total brightness that is about the same as other colors, but in the maxbright mapping, purple is (255,0,255) which has the same hue, but is as bright as possible.  This, in my head, is the "Burning Man" mapping, where you're going to maximum brightness at all times.  The other hsv2rgb mapping is a "blue-corrected" one, which reduces the brightness of the blue component to more accurately represent the lower inherent brightness of blue colors.

Anyway, I think you can see where I'm going with this: what do you think of the idea of having something like a #define FASTLED_HSV2RGB_DEFAULT_MAPPING which could be FASTLED_HSV2RGB_RAINBOW or FASTLED_HSV2RGB_SPECTRUM or FASTLED_HSV2RGB_MAXBRIGHT or FASTLED_HSV2RGB_RAINBOW_LOWBLUE ?

smurphyUSI:
I didn't consider more then two choices when updating the code to swap to the other mapping.

Give me a few days and I'll explore changing the code to use a template to to the swap.
It should be easier for users to see that the can use different mappings, with the bonus that the unused mappings will complied out of the binary.
If that turns out to be a pain, then I can instead put in your list of defines.

'MAXBRIGHT' could be helpful in many situations, but I have a concern that your 'LOWBLUE' options maybe stepping into the responsibility of the color correction functions.

smurphyUSI:
I finished testing alternatives using templates, and in all variations, it's a disaster.
It doesn't seem possible to override all instances on a global level using this method.

I recommend using your suggested FASTLED_HSV2RBG_DEFAULT_MAPPING define for this.

kriegsman:
Thanks for thinking it through here with me - how to offer different options for HSV to RGB conversion.  And you're right that you can definitely 'simulate' the RAINBOW_LOWBLUE by adjusting the color correction values (I've done this a bunch myself) -- but that affects _all_ your pixels' output values, not just pixels that converted from HSV values.  So I think there's probably still some use for RAINBOW_LOWBLUE separate from color correction.  

Anyway, now I'm thinking about how we can get this in; I'd really like to have it in there.  And, as you said, if we can avoid more templates, I think we'll all be happier.

zackees:
I'll give this attention if you are willing to make some changes.

First, this cl needs to be squashed into one commit. Usually I do git commit --amend and a force push when i'ml doing branches, or I do a git rebase --interactive. You can also just copy the code and diff it back into a new branch.

Otherwise I have to do this. And eventually I'll try it out because this CL looks interesting but I can't evaluate it when it's spread out in 6 different commit layers.

The second this is that this is a lot more code I expected for a simple feature like this. I don't like the large diffs in the files. If you are going to add a new something then put it as it's own file and include it in somehow.

PR #1405, opened 2022-07-07
Fix issue #414 "FastLED doesnt work with Arduino MKRZero" by adding A…

sonatique:
…RDUINO_SAMD_MKRZERO to the list of define leading to correct pin definition for MKR ZERO arduino board


kriegsman:
Thank you for this contribution to FastLED, and for your help with the library!

PR #1411, opened 2022-07-17
Add esp32s2 spi (FSPI) support, some code usability improvements

m-g-m:
Add FSPI bus to support ESP32S2.
Add spi.h include for easier compilation.
Make ledSPI static to avoid multiple instantiations.


samguyer:
@m-g-m Thank you for your contribution! A few people have been asking for this update, so it will be much appreciated.

Issue #1363, opened 2022-02-02
FastLED not functional on Trinket M0 with WS2812B


Spegs21:
I've been trying to get the blink sketch running on a Trinket M0 but have so far been unsuccessful. Output from pin 1 is buffered to through a 1g126 to 5V. The same setup works just fine with the Neopixel library.

I have a picoprobe SWD hooked up to it and it's not stuck or in hardfault. I'm not sure what to look at to debug this issue.

Spegs21:
Forcing software pins in the config allows it to function. I checked the pin with a scope in hardware mode and there is no activity.

JesseMFrey:
@Spegs21 which config are you referring to? I'm having a similar problem with on the QtPy SAMD21. Works with Neopixel library, but not FastLED. weird part is it worked before, can't figure out what changed

Spegs21:
@JesseMFrey In the fastled_config.h file:
https://github.com/FastLED/FastLED/blob/4d73cddfe4bd2b370ee882b6f68769bf7f8309f4/src/fastled_config.h#L12

Spegs21:
It's probably just all SAMD21 based boards. @samguyer can you comment on why this might be happening?

samguyer:
@Spegs21  I'm not an expert in the ARM support for FastLED. I wonder if it's just a pin naming problem? But @JesseMFrey , you are saying that it *used* to work and now it doesn't?

JesseMFrey:
@samguyer yes, it used to work. Sadly I was not paying enough attention to know which versions of things I was using before that worked.

samguyer:
@JesseMFrey That's strange. Looking at the history, I don't see anything substantial committed in some time. Do you know which specific ARM platform was affected? The D21?

JesseMFrey:
@samguyer presumably, I've only used it with the QtPy SAMD21

Spegs21:
Seeing as these are both Adafruit boards I wonder if it's related to a Adafruit SAMD Core change.

Sheepybloke2-0:
[...]

PR #1424, opened 2022-08-12
Fix issue #1363 for adafruit d21 based boards

shlomozippel:
Moving the generic version to the end of the pin mapping file fixes the issue where generic SAMD ZERO pin mapping were being used for specific adafruit boards like the Trinket M0 or Gemma M0. 

Tested with a Gemma M0

See #1363


kriegsman:
I'm going to Close and Re-Open this PR to kick off a fresh round of CI tests

kriegsman:
Thank you for this PR, and for your continuing help with FastLED!

PR #1425, opened 2022-08-17
Fix keyword `register` warnings

winking324:
`register` is deprecated in C++11. It is unused and reserved in C++17.

[C++ keywords: register](https://en.cppreference.com/w/cpp/keyword/register)


smurphyUSI:
This was reported previously with issue 1397.
@JakeSays found a temporary work around by changing fastpin.h
`#pragma GCC diagnostic ignored "-Wregister"`

I think your pull request is quite good, but the maintainers may have to update your define in cpp_compat.h
There maybe instances where the user may want to manualy override the compiler option.
Otherwise, I think this is a great patch.

JakeSays:
Ah this is a much better solution than mine.

kriegsman:
This is a much-appreciated change, and I appreciate the way you implemented it, @winking324.  Thank you, also, @JakeSays, for exploring the pragma approach.

PR #1427, opened 2022-09-01
Fix fill_noise functions with lengths > 255

dmadison:
These two functions (`fill_noise8` and `fill_noise16`) take an `int` for the length parameter but use the `fill_noise_raw` functions internally which use a `uint8_t` for the same parameter. This causes an issue where CRGB arrays longer than 255 won't be filled.

To fix, this changes these functions so that they loop through the fills in increments of 255. This preserves the existing behavior and function signature while allowing fills longer than 255.


kriegsman:
Great change - thank you for the improvement to the noise functions, and for your help improving the whole FastLED library.  Very much appreciated, and thank you again for your help.

Issue #1419, opened 2022-08-06
Mystery PR checks fail: "Error: Unknown development platform 'atmelavr'"


ssilverman:
My last ~two~ three PRs can't pass checks because of an "Error: Unknown development platform 'atmelavr'" error.

https://github.com/FastLED/FastLED/pull/1413
https://github.com/FastLED/FastLED/pull/1418
https://github.com/FastLED/FastLED/pull/1420

kriegsman:
In some places online I'm seeing 'atmelmegaavr' used instead of 'atmelavr'; looking into it, but any additional thoughts or input would be quite welcome, as I'm definitely not a CI expert here!

kriegsman:
OK, this looks fixed by https://github.com/FastLED/FastLED/pull/1428, with thanks to @dmadison.

In order to process the PRs that were submitted in the interim, I'll probably have to Close and Re-open them to make the updated tests re-run.

PR #1428, opened 2022-09-02
Fix CI Build Issues

dmadison:
The build script used for checking all examples with PlatformIO is currently broken. It appears that this is due to the older version of PlatformIO being incompatible with the current platform packages.

This PR updates the packages used in the build script to their latest versions:

| Dependency           | From  | To     |
|----------------------|-------|--------|
| PlatformIO           | 5.03  | 6.1.4  |
| Python               | 3.8   | 3.10   |
| actions/checkout     | v2    | v3     |
| actions/setup-python | v2    | v4     |
| Ubuntu               | 20.04 | latest |

I ran the latest commit (637fac50a28d597e122678340a5cb07a1c766fdc) on my fork [here](https://github.com/dmadison/FastLED/runs/8145649975) and everything appears to be working swimmingly.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,4 +1,5 @@
 on:
+  workflow_dispatch:
   push:
     branches:
       - master
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -22,7 +22,7 @@ jobs:
 
       - name: install platformio
         run: |
-          pip install platformio==5.0.3
+          pip install platformio==6.1.4
 
       - name: build FastLED examples
         run: ./ci/ci-compile
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -11,15 +11,15 @@ name: build
 jobs:
 
   build:
-    runs-on: ubuntu-20.04
+    runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@v2
+        uses: actions/checkout@v3
 
       - name: install python
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v4
         with:
-          python-version: '3.8'
+          python-version: '3.10'
 
       - name: install platformio
         run: |


kriegsman:
Thank you very much for this contribution, and for fixing the CI issues!  Much appreciated!

kriegsman:
OK, so now I'm seeing different build issues and would like your thoughts on it if you have any, @dmadison 

E.g. in https://github.com/FastLED/FastLED/actions/runs/2975023467/jobs/4766378325,

```
Processing uno (platform: atmelavr; board: uno; framework: arduino)
--------------------------------------------------------------------------------
Platform Manager: Installing atmelavr
Downloading...
Unpacking...
Platform Manager: atmelavr @ 4.0.0 has been installed!
Error: Unknown development platform 'atmelavr'
Error: Process completed with exit code 1.
```

kriegsman:
Looks related to https://github.com/FastLED/FastLED/issues/1419

dmadison:
That's the issue this PR is supposed to fix - the AVR platform updated and was no longer compatible with the older version of PlatformIO. Try running those jobs again.

kriegsman:
Please forgive my newbie status on CI issues, and thank you for the help.  When I _literally_ just re-run the jobs, I get the same errors; it is as if the 're-run' isn't re-running with the latest updates that you contributed here.  Do you know if there's something I need to do to get one of those, e.g. this (very nice) one, to pick up the new CI config? https://github.com/FastLED/FastLED/pull/1400

dmadison:
I think you're right, and re-running the exact same job will use the older build config. Hmm.

Searching around it seems that there isn't a good turn-key solution to re-run the older jobs with updates to the script. I *think* the solution is to create new jobs altogether based off of the same commit. In other words, trigger the `pull_request` action hook in some way.

By default the `on: pull_request` hook triggers on `opened`, `synchronize` (commits), and `reopened` types.  The easy (but messy) way to trigger a new build without any changes would be to close and re-open the PR in question. I just tried that on #1427 and it seemed to do the trick.

Alternately, the build script could be modified to trigger on a different `pull_request` type, such as `labeled` or `assigned`. Then marking the PR with those attributes should trigger another build.

I'm surprised there doesn't seem to be a simpler solution. I'm also not an expert with GitHub Actions, so there may be something both you and I are missing.

kriegsman:
Great, thank you for your insights and the suggested approach -- I just tried it on https://github.com/FastLED/FastLED/pull/1400 and it seems to have worked as intended and desired!  Very much appreciated, @dmadison!

PR #1434, opened 2022-09-20
Adafruit Grand Central M4 support

jgildred:
Submitting as I tried to confirm it works but can’t see why it is not recognizing defined ADAFRUIT_GRAND_CENTRAL_M4 which is define on the platform. The data pin mapping is complete.


kriegsman:
Thank you so much for this contribution, and for your help with FastLED!

PR #1450, opened 2022-11-26
Make CRGB::operator bool() explicit

### LINKED ISSUE #819: Making CRGB::operator bool() explicit would be useful
It's easy to assign CRGB to an integral value for a number of reasons, however, if you aren't careful to use your own, for example, `uint32 toColor` function, the `bool()` operator is used.

My suggestion is to make `CRGB::operator bool()` `explicit`.

Ref: See mention of the Safe Bool idiom mentioned in the "Explicit Conversion Operators (C++11)" section at: https://stackoverflow.com/a/16615725.

I find myself doing this:

```c++
// Extends FastLED's CRGB to add a uint32_t conversion and make
// `operator bool()` explicit.
class MyCRGB : public CRGB {
 public:
  MyCRGB() : CRGB(){};
  explicit MyCRGB(const CRGB &rhs) : CRGB(rhs){};
  explicit MyCRGB(const CHSV &rhs) : CRGB(rhs) {}
  explicit MyCRGB(uint32_t c) : ::CRGB(c) {}
  ~MyCRGB() = default;

  MyCRGB &operator=(const CRGB &rhs) {
    CRGB::operator=(rhs);
    return *this;
  }

  MyCRGB &operator=(const CHSV &rhs) {
    CRGB::operator=(rhs);
    return *this;
  }

  explicit operator bool() const {
    return CRGB::operator bool();
  }

  // Converts a CRGB to a 32-bit colour having an alpha of 255.
  explicit operator uint32_t() const {
    return uint32_t{0xff000000} |
           (uint32_t{r} << 16) |
           (uint32_t{g} << 8) |
           uint32_t{b};
  }
};
```

ssilverman:
Resolves #819

### DIFF
--- a/src/pixeltypes.h
+++ b/src/pixeltypes.h
@@ -477,7 +477,7 @@ struct CRGB {
     }
 
     /// This allows testing a CRGB for zero-ness
-    inline operator bool() const __attribute__((always_inline))
+    inline explicit operator bool() const __attribute__((always_inline))
     {
         return r || g || b;
     }


ssilverman:
Converting to a draft until I verify that this works on compilers before C++ 11.

ssilverman:
@kriegsman what is the minimum acceptable compiler version for FastLED? For example, as-is, this change works for C++11. Should I target C++03?

kriegsman:
As I think noted elsewhere, let's aim for C++11 compliance; no need to go all the way back to C++03 since the Arduino IDE is providing C++11.

ssilverman:
One other note is that implicit conversions will need to be explicitly cast or used as a Boolean expression. See here: https://github.com/ssilverman/QNEthernet/blob/master/README.md#operator-bool-and-explicit

If that caveat is fine, then let’s get this in.

kriegsman:
Thank you for the contributions and your help with FastLED!
I'm sure that this will break _someone's_ code somewhere, possibly even mine ... I can't say for sure that I haven't accidentally done this, but this will be easy to find and fix.  Thanks again for all the contributions of time and thinking and code!

PR #1451, opened 2022-11-26
Add CRGB::operator uint32_t()

### LINKED ISSUE #820: Add a CRGB::operator uint32_t()
I'd find it useful to have an `explicit operator uint32_t()` function in CRGB because I'm frequently caching this value. (My style would be to use `explicit` to avoid unintended integral conversions, like what happens with `operator bool()` from #819.)

ssilverman:
Resolves #820

### DIFF
--- a/src/pixeltypes.h
+++ b/src/pixeltypes.h
@@ -482,6 +482,15 @@ struct CRGB {
         return r || g || b;
     }
 
+    /// Converts a CRGB to a 32-bit color having an alpha of 255.
+    inline explicit operator uint32_t() const
+    {
+        return uint32_t{0xff000000} |
+               (uint32_t{r} << 16) |
+               (uint32_t{g} << 8) |
+               uint32_t{b};
+    }
+
     /// Invert each channel
     inline CRGB operator- () const
     {


kriegsman:
Thanks for thinking of this and adding it.

I was a little surprised by the type-constructor (I think?) syntax using curly braces in 

```
        return uint32_t{0xff000000} |
               (uint32_t{r} << 16) |
               (uint32_t{g} << 8) |
               uint32_t{b};
```

I've not seen that syntax before -- I've always seen it more as a cast, something ilke:
```
        return uint32_t(0xff000000) |
               (uint32_t(r) << 16) |
               (uint32_t(g) << 8) |
               uint32_t(b);
```
But I presume this is something in newer versions of C++ than the fairly old one that I have in my head.  So apparently I'm learning new things today.

Do you know (1) which version of C++ they introduced this in -- or has it always been there as an option, and (2) if there's an expressive or functional difference between the {} style and the () style, in this case?

zackees:
According to this, it was introduced in C++11

https://en.cppreference.com/w/cpp/language/list_initialization

ssilverman:
Thanks for the notes. I should probably make sure this works for compilers before C++ 11. Converting to a draft until this is so. Oh, by the stairs look at the stairs.

kriegsman:
Huh.  Well, now I know.  The 'list initialization' -- even of a scalar -- can flag 'narrowing' errors (at compile time).  And it's old: C++11, I believe.  Well, ok!

ssilverman:
@kriegsman what is the minimum acceptable compiler version for FastLED? For example, as-is, this change works for C++11. Should I target C++03?

Yes, that’s why I use those with scalars for some types of conversions. It’s a handy compile-time check. :)

kriegsman:
Well, the current Arduino IDEs are C++11.  We talked about it a bit here, and decided that C++11 seems like a good target for the library overall.

ssilverman:
Thanks, @kriegsman! I'll un-draft this now, then.

ssilverman:
Note: I made the style match, but I think all the `inline` keywords are redundant since these functions are defined inside the class declaration.

kriegsman:
Thank you for this contribution -- and all of the others!  The help and input is greatly appreciated, and it really helps improve the library.  Thanks again!

Issue #1457, opened 2022-12-29
ISO C++17 does not allow 'register' storage class specifier anymore


nielsnl68:
Hi, with the latest versions of espressif8266 using the most recent version of PlatformIO i get the warning that ISO C++17 does not allow 'register' storage class specifier anymore

here is my compiler header message:
```
processing lum-iot25 (board: esp01_1m; framework: arduino; platform: platformio/espressif8266 @ 3.2.0)
--------------------------------------------------------------------------------
Library Manager: Installing ottowinter/ESPAsyncTCP-esphome @ 1.2.3
INFO Installing ottowinter/ESPAsyncTCP-esphome @ 1.2.3
Unpacking  [####################################]  100%
Library Manager: ESPAsyncTCP-esphome@1.2.3 has been installed!
INFO ESPAsyncTCP-esphome@1.2.3 has been installed!
Library Manager: Installing esphome/ESPAsyncWebServer-esphome @ 2.1.0
INFO Installing esphome/ESPAsyncWebServer-esphome @ 2.1.0
Unpacking  [####################################]  100%
Library Manager: ESPAsyncWebServer-esphome@2.1.0 has been installed!
INFO ESPAsyncWebServer-esphome@2.1.0 has been installed!
Library Manager: Resolving dependencies...
INFO Resolving dependencies...
Library Manager: Installing fastled/FastLED @ 3.5.0
INFO Installing fastled/FastLED @ 3.5.0
Unpacking  [####################################]  100%
Library Manager: FastLED@3.5.0 has been installed!
INFO FastLED@3.5.0 has been installed!
HARDWARE: ESP8266 80MHz, 80KB RAM, 1MB Flash
LDF: Library Dependency Finder -> https://bit.ly/configure-pio-ldf
Dependency Graph
|-- ESPAsyncTCP-esphome @ 1.2.3
|-- ESPAsyncWebServer-esphome @ 2.1.0
|   |-- ESPAsyncTCP-esphome @ 1.2.3
|   |-- Hash @ 1.0
|   |-- ESP8266WiFi @ 1.0
|-- DNSServer @ 1.1.1
|-- ESP8266WiFi @ 1.0
|-- ESP8266mDNS @ 1.2
|-- FastLED @ 3.5.0 
```

and this part of the compiler log:
```
[...]

PR #1460, opened 2023-01-02
fixes https://github.com/FastLED/FastLED/issues/1457

zackees:
C++17 doesn't support the register keyword. This was already accounted for with the new `#define REGISTER`, however it wasn't applied everywhere. This pull request completes the migration in `fastpin.h` in order to address issue https://github.com/FastLED/FastLED/issues/1457

### DIFF
--- a/src/fastpin.h
+++ b/src/fastpin.h
@@ -88,7 +88,7 @@ public:
 	/// Set the state of a port
 	/// @param port the port to modify
 	/// @param val the state to set the port to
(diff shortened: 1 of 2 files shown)


zackees:
[@kriegsman](https://github.com/kriegsman)

kriegsman:
Thank you for adding these fixes, very much appreciated.  And thanks for your help with the library!

PR #1461, opened 2023-01-03
Update all REGISTER to FASTLED_REGISTER

nielsnl68:
In the latest compilers we get an error that "register" is no longer supported there was already done some work but not all register where update to REGISTER.

After the last place were fixed i found out that the use of REGISTER was also used in other libraries as well and the behavier of that library broke because of the REGISTER was being used in the FastLED library. Changing REGISTER to FASTLED_REGISTER will make user that no other library will be affected anymore.

See also #1457


zackees:
Figure out how to squash all your commits into one, or otherwise recreate this pull request

nielsnl68:
> Figure out how to squash all your commits into one, or otherwise recreate this pull request

Not sure what you mean?

zackees:
> Not sure what you mean?

You have 10 commits, a pull request should have exactly one or else we can't see what changed because it's spread out through multiple commits.


https://stackoverflow.com/questions/5189560/how-do-i-squash-my-last-n-commits-together

zackees:
Looking through this code [here](https://github.com/FastLED/FastLED/pull/1461/files/55f00bde45e910ea672838bc25754299d8df3634..622ef15cd2d26490b0eb0e60ae9ec14f3d540c4e) I see that you are replacing `REGISTER` with `FASTLED_REGISTER`. And the way you are doing this is the exact same thing as exists right now:

https://github.com/FastLED/FastLED/blob/master/src/cpp_compat.h#L24

```C++
#if __cplusplus < 201703L
#define REGISTER register
#else
#define REGISTER
#endif
```

See how REGISTER for C++17 and above (>=201703L) evaluates to nothing? That's exactly what you are doing using FASTLED_REGISTER. You should revert your FASTLED_REGISTER changes. Additionally is tough to be able to evaluate your changes because you've got so many commits, and I see that the diff between (above) is diffing against one of your revisions instead of the head revision.

Since I went in and fixed the register -> REGISTER issue, it's entirely possible that this already fixes your issue.

Why don't you try changing your src of FastLED from `https://github.com/FastLED/FastLED` to `https://github.com/zackees/FastLED`. You can do this pretty painlessly in platformio and likely Arduino IDE. Or you can just clone my github and in your Arduino project in a directory at the root called `src`.

nielsnl68:
> Looking through your code diff [here](https://github.com/FastLED/FastLED/pull/1461/files/55f00bde45e910ea672838bc25754299d8df3634..622ef15cd2d26490b0eb0e60ae9ec14f3d540c4e) I see that you are replacing `REGISTER` with `FASTLED_REGISTER`. And the way you are doing this is the exact same thing as exists right now:
> 
> https://github.com/FastLED/FastLED/blob/master/src/cpp_compat.h#L24
> 
> ```c++
> #if __cplusplus < 201703L
> #define REGISTER register
> #else
> #define REGISTER
> #endif
> ```
> 
That is correct the reason that i changed REGISTER to FASTLED_REGISTER is because the define REGISTER could be used in other libraries as well. And that could harm in the long run; 

> See how REGISTER for C++17 and above (>=201703L) evaluates to nothing? That's exactly what you are doing using FASTLED_REGISTER. You should revert your FASTLED_REGISTER changes. Additionally its tough to be able to evaluate your changes because you've got so many commits, and I see that the diff between (above) is diffing against one of your revisions instead of the head revision.
> 
As i wrote above i did it with an purpose so i do not like to revert this.
 
> Since I went in and fixed the register -> REGISTER issue, it's entirely possible that this fixes your issue.
> 
Sadly NO. the PR also fixes the ICACHE_RAM_ATTR .

> Why don't you try changing your src of FastLED from `https://github.com/FastLED/FastLED` to `https://github.com/zackees/FastLED`. You can do this pretty painlessly in platformio and likely Arduino IDE. Or you can just clone my github and in your Arduino project in a directory at the root called `src`. If my fork doesn't fix your issue then post it here and I'll take a look at it. I'm an experienced C++ developer and familiar with FastLED so I can help you get through this.

Why would i do that i already cloned and made the necessary changes to get it compiling without any warnings and harming other libraries in the main time.

samguyer:
@nielsnl68 Thanks for your work on FastLED! I'd like to separate the whole register thing from the IRAM_ATTR change (which I'm planning to merge in right away). Would you mind making a separate PR just for that change?

nielsnl68:
> @nielsnl68 Thanks for your work on FastLED! I'd like to separate the whole register thing from the IRAM_ATTR change (which I'm planning to merge in right away). Would you mind making a separate PR just for that change?

okay done. i also updated the first post explaining why i suggest this PR is imho important to be added as well.

nielsnl68:
> @nielsnl68 Thanks for your work on FastLED! I'd like to separate the whole register thing from the IRAM_ATTR change (which I'm planning to merge in right away). Would you mind making a separate PR just for that change?

Hi @samguyer, I'm still waiting on those PR to being merge.

samguyer:
Merged! Thank you for this contribution.

samguyer:
Closing and reopening to force a resync for CICD pipeline.

PR #1466, opened 2023-01-22
fix missing ICACHE_RAM_ATTR change to IRAM_ATTR

nielsnl68:
As requested by @samguyer her is the missing change from ICACHE_RAM_ATTR  to IRAM_ATTR.


samguyer:
Thank you for your contribution to FastLED!

PR #1467, opened 2023-01-22
remove a warning that still does not make sense.

nielsnl68:
This fix was also part of my earlier PR.

I added this fix after an idea from @zackees 

but the strangest thing is that some rows below the same sort of while seams totally fine.


zackees:
To clarify, C++ compiler does not like

```
while (condition);
```

But it does like

```
while (condition) {
;
}
```

nielsnl68:
@zackees , when you look at the same file/function than there a couple of other while loops that is written the same way and does not give the same warning.

zackees:
weird

smurphyUSI:
I now suspect that it's a compiler inline error.

Certain compilers will make mistakes on code elimination if it can't clearly figure out if a function is static or volatile.
You could check the compiled output, but I suspect that it was incorrectly trying to inline the __clock_cycles() code.
(You could alternatively check by removing the inline attribute and keyword from that function)

kriegsman:
That is an odd warning - but thank you for digging in and working on it.  Thanks for the help with the FastLED library!

PR #1478, opened 2023-02-11
Add functions to get the number of lanes from controllers.

kcouck:
This can be used by external code to get this information without needing access to the 'input' of the template.


kcouck:
Should be squashed to remove the .gitignore update. I need to change the layout of my repo to just have this applied locally

kriegsman:
Thank you for this addition, and thanks for helping out with FastLED!

PR #1479, opened 2023-02-18
added support for DP1903

KhanKudo:
It's a variant of WS2811 with significantly different timings


kriegsman:
Thank you very much for this addition to the library, and thanks for helping out with FastLED!

kriegsman:
We're having some build problems now that this is merged.  I'm going to back it out, and see if we still have the build problems -- either way I'd like to get this merged in, but I need to puzzle out the build problems with the 'yun' platform.

KhanKudo:
Interesting. I can't see anything that could cause any issues, but definitely let me know if you find anything.

kriegsman:
Thank you for the contribution and the help!  After reverting this change, the builds succeeded.  @samguyer and I looked at it a little bit, and are going to look in a bit further.  We're doing a (loooong-overdue) new release of the library today, and if we can't get it figured out before  today's release, I'll do a second quick release to incorporate this support for the DP1903 that you put together.

The problem that we were having looked like it had something to do with the template expansion for the new controller, which is odd because the new code looks pretty much exactly like the code for some other chipsets that are compiling just fine.

samguyer:
Hi @KhanKudo I think we found the issue. In chipsets.h there are two entries for each of the clockless chips: one for AVR and one for everything else. The definitions you added are only in the "for everything else" part of the header file. The two sections are divided by the big #if statement on line 484 (here: https://github.com/FastLED/FastLED/blob/15d619db5ac0fc559694b57857a13031a85e01ac/src/chipsets.h#L482).

What you'll need to do is add definitions for the DP1903 controllers in that first section of the conditional compilation. Note that the timing is expressed in terms of FMUL, which is the cycle time for 8MHz (125 nanoseconds). You can compare the examples for WS2812 to see what to do.

If you need help, please don't hesitate to contact us!

samguyer:
Looking at this definition:

`// DP1903 - 250ns, 1000ns, 250ns                                                                                           
template <uint8_t DATA_PIN, EOrder RGB_ORDER = RGB>
class DP1903Controller800Khz : public ClocklessController<DATA_PIN, C_NS(250), C_NS(1000), C_NS(250), RGB_ORDER> {};`

250ns should be 1*FMUL, and 1000ns should be 4*FMUL. So, the final AVR definition should look like this:

`// DP1903 - 250ns, 1000ns, 250ns                                                                                           
template <uint8_t DATA_PIN, EOrder RGB_ORDER = RGB>
class DP1903Controller800Khz : public ClocklessController<DATA_PIN, FMUL, 4*FMUL, FMUL, RGB_ORDER> {};`

KhanKudo:
Hmm, well based on the WS2812 and WS2811, shouldn't 250ns be 2 * FMUL? With 1000 ns being 8*FMUL and 2000 16*FMUL?

samguyer:
@KhanKudo Yes, you are absolutely correct. Sorry for the brain malfunction!

KhanKudo:
No worries mate. I'm sure pushing a new release after more than a year is plenty stressful enough. ;)

Issue #1485, opened 2023-03-14
The declaration of the template is not as per the standards.


BabaYaga1221:
Recently, I am working on a robotics project where I found that the iterator name is redeclared. As a result, I found that the name of the template class in ``FastLED/src/lib8tion.h 827:39`` is declared as ``q`` which is not as per the standards initialed. I thought that people are facing the same errors as I do. If you would like to adjust the standard, assign this issue to me. Thanks

~Regards

<img width="279" alt="image" src="https://user-images.githubusercontent.com/86480450/224982895-8d3aa0fb-1856-407e-97bd-f8cac09ea2bb.png">

zackees:
What's the error message and what are repro steps including your operating system and other setup parameters?

zackees:
This symbols are missing on some toolchain setup. That's why they are defined here. There should be a pre processor define to conditionally compile these symbols

vortigont:
actually there are "[Naming Conventions in C++](https://www.geeksforgeeks.org/naming-convention-in-c/)" where classes are declared with first uppercase letter and vars are lowercase'd. Not mandatory for sure, but a respectful style de facto. So would be better to use `Q` or `Qfx`.

BabaYaga1221:
@robertlipe is right. While it does shift the issue, my approach was to replace "q" with "qfx" as a temporary fix. I believe it's important to create documentation for the repository and provide clear explanations about these changes.

zackees:
I see that q was actually replaces with qfx so this should be resolved and thus closing the issue.

PR #1486, opened 2023-03-16
update the class q name as class qfx in lib8tion.h

BabaYaga1221:
I update the template class named as q to qfx which might resolve the problem as re-declaration of q of different kind of type.
This PR will might resolve issue #1485.


kriegsman:
Great contribution- thank you very much for this, and for your help with FastLED!

PR #1493, opened 2023-04-07
add support for ATtiny48/88

VAleSh-Soft:
add support for ATtiny48/88


kriegsman:
Thank you for this addition, and thanks for helping out with FastLED!

PR #1495, opened 2023-04-16
Add operator==() and operator!=() for CHSV

ssilverman:


kriegsman:
Thank you for this addition to the library, and thanks for helping out with FastLED!

PR #1514, opened 2023-06-21
Include OctoWS2811.h if USE_OCTOWS2811 is defined

ssilverman:
This is a convenience in the style of the library already including SmartMatrix, DmxSimple, and DMXSerial. It avoids having to include the header every time FastLED.h is included; this can get tedious for a larger project.

### DIFF
--- a/src/FastLED.h
+++ b/src/FastLED.h
@@ -43,6 +43,10 @@
 #include <DMXSerial.h>
 #endif
 
+#ifdef USE_OCTOWS2811
+#include <OctoWS2811.h>
+#endif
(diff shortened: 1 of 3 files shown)


ssilverman:
Can someone comment on why this wasn't done before? The includes were actually commented out in the _octows2811_controller.h_ files, which leads me to think there might have been a good reason for not doing this before.

kriegsman:
Looks good.  Thank you for making this a little bit easier for folks using OctoWS, and thanks for your help with the FastLED library!

PR #1525, opened 2023-07-16
Improve OctoWS2811 driver

ssilverman:
These improvements include:
1. Some minor improvements for C++-isms,
2. Better handling for when there's low memory conditions so the system doesn't crash,
3. ~~Added a destructor for freeing the allocated memory, and~~
4. A way to use external buffers for the controller.

When there's low memory, the behaviour degrades gracefully. Firstly, the frame buffer is allocated first and the draw buffer is allocated next. Because a NULL draw buffer is allowed by the underlying OctoWS2811 driver, a failure to allocate it is okay. Second, if there's no frame buffer, then nothing else is allocated and `showPixels()` doesn't do anything. Same for an error allocating the OctoWS2811 driver.

Solving the potential system crash was necessary for a project of mine that let the user dynamically change a configuration. If the pixel count was too large, the memory allocation done in the original _Octo_ controller failed leading to a crash when `showPixels()` was eventually called. The system rebooted and didn't allow the user adjust the settings. Point number two solves this problem.

For the fourth point, the external buffers, that's very useful for a larger project of mine where I need to manage the memory externally. I use the `COctoWS2811Controller` controller directly. Note that the new parameters have defaults, meaning it’s optional to include them. This means that existing code won’t need to change.

### DIFF
--- a/src/platforms/arm/k20/octows2811_controller.h
+++ b/src/platforms/arm/k20/octows2811_controller.h
@@ -13,18 +13,35 @@ class COctoWS2811Controller : public CPixelLEDController<RGB_ORDER, 8, 0xFF> {
   uint8_t *drawbuffer = nullptr;
   uint8_t *framebuffer = nullptr;
 
+  bool tryToAllocate = true;
+
   void _init(int nLeds) {
-    if (pocto == nullptr) {
-      drawbuffer = static_cast<uint8_t *>(malloc(nLeds * 8 * 3));
+    if (pocto == nullptr && tryToAllocate) {
+      // Allocate the draw buffer second in case the malloc fails,
+      // because it can technically be NULL. In other words,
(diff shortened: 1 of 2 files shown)


ssilverman:
Also, some kind of logging mechanism would be a useful thing to add sometime in the future.

ssilverman:
I just removed the destructor commit. Probably not needed at the moment.

ssilverman:
I should note that these changes shouldn’t affect existing code.

ssilverman:
I'm not certain this will be looked at, and I've got more advanced stuff that's an improvement on this driver that obsoletes what I have here. I'm also afraid someone else will submit this work as their own due to the time it takes for PRs to be looked at and incorporated. Closing because I don't think this project is on track anymore.

Issue #920, opened 2019-11-19
FastLED and Adafruit Grand Central


winyguy:
Is there an update for FastLED that works with the Adafruit Grand Central?

Thanks.

spaceriqui:
You can use Circuit Python for the same thing instead of FastLed.  That's what I ended up doing.    https://learn.adafruit.com/adafruit-grand-central/what-is-circuitpython

bwmaas:
Should have a fix in PR #1539 to get this platform working.

zackees:
Closing — **the Adafruit Grand Central M4 is supported**, and is in CI.

@bwmaas was right that #1539 would sort it. Current state:

- Pin definitions: `src/platforms/arm/d51/fastpin_arm_d51.h:210`, under `#elif defined(ADAFRUIT_GRAND_CENTRAL_M4)`
- CI board: `adafruit_grand_central_m4` in `ci/boards.py:1518`

So it's not just present but built on every push, alongside the other SAMD51 boards (Feather M4, Matrix Portal M4).

```cpp
FastLED.addLeds<WS2812B, DATA_PIN, GRB>(leds, NUM_LEDS);
```

@spaceriqui — CircuitPython was a reasonable escape hatch in 2021, and worth leaving on the record for anyone who prefers it. But you don't have to give up FastLED for this board any more.

The Grand Central is a good target for it, incidentally: SAMD51 at 120 MHz with 256 KB RAM handles large strips comfortably, and the board has plenty of free pins for parallel output.

Sorry the original question went four years without a direct answer.

Issue #1531, opened 2023-08-25
Sketch fails to compile on Adafruit Grand Central M4 Express: "fastpin.h:261:24: error: static assertion failed: Invalid pin specified"


JNordgren42:
I'm unable to compile even the simplest sketch using FastLED on an Adafruit Grand Central M4 Express with DotStar LEDs. 

12 pixel DotStar strip connected to the Hardware SPI pins (Pin 51 Data, Pin 52 Clock). Power from board 5V, Ground to board GND. 

Arduino IDE version 2.1.1, 06/30/2023
Adafruit SAMD Boards package version 1.7.13
FastLED version 3.6.0

This configuration functions perfectly using the Adafruit DotStar library. I have confirmed the board definition is in **\FastLED\src\platforms\arm\d51\fastpin_arm_d51.h** and the pins appear to be correctly mapped, specifically:

_#define SPI_DATA 51_
_#define SPI_CLOCK 52_

_#define HAS_HARDWARE_PIN_SUPPORT 1_

I have also tried several other pin combinations that are included in the mapping and receive the same error.

Sketch:
```
#include <FastLED.h>

#define NUM_LEDS 12

#define DATA_PIN 3
#define CLOCK_PIN 2

CRGB leds[NUM_LEDS];

void setup() { 
  FastLED.addLeds<DOTSTAR, DATA_PIN, CLOCK_PIN>(leds, NUM_LEDS);
}

void loop() { 
  leds[0] = CRGB::Red; 
  FastLED.show(); 
  delay(30); 
}
```
Serial Output:

In file included from C:\Users\jason\OneDrive\Documents\USS Enterprise\Grand Central M4\FastLED\Blink\Blink.ino:1:
c:\Users\jason\OneDrive - Associated Building Supply\Documents\Arduino\libraries\FastLED\src/FastLED.h:665:2: warning: #warning "No pin/port mappings found, pin access will be slightly slower. See fastpin.h for info." [-Wcpp]
  665 | #warning "No pin/port mappings found, pin access will be slightly slower. See fastpin.h for info."
      |  ^~~~~~~
In file included from c:\Users\jason\OneDrive - Associated Building Supply\Documents\Arduino\libraries\FastLED\src/FastLED.h:75,
                 from C:\Users\jason\OneDrive\Documents\USS Enterprise\Grand Central M4\FastLED\Blink\Blink.ino:1:
[...]

Issue #1071, opened 2020-08-19
Add support for Adafruit Grand Central M4 (SAMD51)


spaceriqui:
Add support for Adafruit Grand Central M4 (SAMD51).  Currently using FastLED 3.3.3.

[Adafruit Grand Central M4 product page](https://www.adafruit.com/product/4064)

[Pinouts](https://learn.adafruit.com/adafruit-grand-central/pinouts)

`In file included from C:\../FastLED.h:41,`
`                 from C:\......ino:29:`
`C:\....\FastLED/led_sysdefs.h:53:2: error: #error "This platform isn't recognized by FastLED... yet.  See comments in FastLED/led_sysdefs.h for options."`
`   53 | #error "This platform isn't recognized by FastLED... yet.  See comments in FastLED/led_sysdefs.h for options."`
`      |  ^~~~~`
`exit status 1`
`Error compiling for board Adafruit Grand Central M4 (SAMD51).`

hkolbeck-streem:
If anyone has time to tackle this it would be much appreciated. I'm very new to the world of microcontrollers or I'd take a stab at it myself.

Also, there's a duplicate of this issue here: https://github.com/FastLED/FastLED/issues/920

spaceriqui:
You can use Circuit Python for the same thing instead of FastLed.  That's what I ended up doing.    https://learn.adafruit.com/adafruit-grand-central/what-is-circuitpython

bwmaas:
I'm willing to start working on this as I need support for this board...

Forking now for development. I'll try to report back in a day or two.

bwmaas:
Should have a fix in PR #1539 to get this platform working.

zackees:
Tracked in FastLED/fbuild#386 for SAMD21 / SAMD51 fbuild platform/build support validation.

That fbuild issue is scoped to board metadata, toolchain/backend support, and FastLED smoke-build evidence. Any remaining FastLED source/platform-code blocker should stay linked from both sides.

zackees:
**fbuild tracker: [FastLED/fbuild#386](https://github.com/FastLED/fbuild/issues/386)**

Validated in the [fbuild](https://github.com/FastLED/fbuild) build tool. The Grand Central M4 (SAMD51P20A) builds cleanly:

- Fixture: [`tests/platform/samd51p`](https://github.com/FastLED/fbuild/tree/main/tests/platform/samd51p) (`board = adafruit_grand_central_m4`)
[...]

PR #1539, opened 2023-09-09
fix missing check for Adafruit Grand Central Metro M4 in platform.h

bwmaas:
The Adafruit Grand Central Metro M4 (`__SAMD51P20A__`) was in [led_sysdefs.h](https://github.com/FastLED/FastLED/blob/2874c7c8eb13d99ec4ad9a3201054c387d2d8aaa/src/led_sysdefs.h#L34) but missing from [platforms.h](https://github.com/FastLED/FastLED/blob/2874c7c8eb13d99ec4ad9a3201054c387d2d8aaa/src/platforms.h#L34).

This should fix Issues #920 and #1071 and potentially #1531.


JNordgren42:
Grabbed the update platforms.h file and that resolved the issue. Thanks!

kriegsman:
Thank you, great catch, and much appreciated patch.  Thanks for helping with the FastLED library!

Issue #1557, opened 2023-11-03
No information on how to port FastLED to a new platform


rah2501:
I've searched the website, the wiki, the FAQ and found no information on how to port FastLED to a new platform.  I would expect at least a paragraph on what one needs to do in the codebase in order to add a new platform.

zackees:
If you attempt to compile it you'll get a compiler error pointing you to the correct files.

The platforms can be found in the platform directory.

samguyer:
@rah2501 I can probably help put this documentation together. I did the ESP32 support, so I had to go through the process recently (relatively recently, anyway). There's also a considerable difference between adding support for a variant of an existing platform (e.g., a different ARM board) vs a totally new platform. What are you hoping to do?

rah2501:
> What are you hoping to do?

To be clear, I created this issue to track the lack of documentation rather than seek help for my own particular porting efforts.  The appropriate place for those discussions would be a separate issue I think.

>  There's also a considerable difference between adding support for a variant of an existing platform (e.g., a different ARM board) vs a totally new platform. 

Can I suggest for consideration, that you could write two responses, one each for both possible answers that I could give, and those responses could be included in the README.md? :-)  That way, you're helping both me and others.

samguyer:
Do you need this documentation right now, or not? I think it's a great idea to get this information together, but I'm unclear on what the impetus is at the moment.

rah2501:
> Do you need this documentation right now, or not?

That depends on what's in the documentation and the documentation doesn't exist :-)  So difficult to answer..

> I'm unclear on what the impetus is at the moment

The impetus for me to create this issue was the knowledge that in future, people will need to port the project and will find an absence of information on porting, as I did.

samguyer:
What I mean, specifically, is: are you currently working on a port or planning to do so in the near future? Do you have a specific platform target in mind?

rah2501:
> What I mean, specifically, is: are you currently working on a port or planning to do so in the near future? Do you have a specific platform target in mind?

I don't think my personal activities are relevant to this issue.

samguyer:
It is relevant, in the following sense: this is an entirely volunteer-run project, so many of the new features and content come from user contributions, or some collaboration between us and the users who need those features. It is very much a community effort. So, if you want new documentation, you are probably going to need to engage with us a bit more.

samguyer:
Actually, it looks like there is an existing (brief) porting document that covers the basics:

[...]

PR #1559, opened 2023-11-04
 [README.md] Add a pointer to PORTING.md

rah2501:
Make a start (however miniscule) on providing information about porting FastLED to new platforms.

Fixes: #1557


rah2501:
Information on PORTING.md apparently already exists so I've updated the pull request to add a pointer to that file instead.

kriegsman:
Thanks for adding this doc pointer; the docs are only useful if people can find them!

Issue #1561, opened 2023-11-08
Should we change CRGB constructors to constexpr?


vovagorodok:
Should we change CRGB constructors to constexpr?:
```
constexpr CRGB(uint32_t colorcode)  __attribute__((always_inline))
```

For example:
```
constexpr uint32_t MY_COLOR = CRGB::Green;
...
CRGB color = CRGB::Black;
...
if (color == MY_COLOR) // HERE
    ...
...
color = MY_COLOR; // AND HERE

```
In the marked lines each time will be created rvalue of CRGB and each time inline construction will be called.
After change constexpr constructor will be called only once in compile time, or there is some additional optimization aspect that I haven't noticed?

vovagorodok:
> ...but CRGB:foo is an enum (maybe a class enum). The compiler will extract the numeric value of that value and use it. There's no real code gen, so I'd be surprised if any self-respecting emitter (i.e. not gcc -O0 which is known to be dain bramaged) would emit anything different in this case.

Sure compiler will extract the numeric value. I mean more about the constructor:
```
    inline CRGB( uint32_t colorcode)  __attribute__((always_inline))
    : r((colorcode >> 16) & 0xFF), g((colorcode >> 8) & 0xFF), b((colorcode >> 0) & 0xFF)
    {
    }
```
where are additional calculations of `r/g/b` fields from `colorcode`

vovagorodok:
Your right, compiler optimizes it as well. For this code: 
```
#include <iostream>

struct Color
{
    constexpr/inline Color(int colorcode) :
        r((colorcode >> 16) & 0xFF), g((colorcode >> 8) & 0xFF), b((colorcode >> 0) & 0xFF)
    {}
    char r;
    char g;
    char b;
};

int main()
{
    Color c{555};
    std::cout << c.b;
}
```
I have identical assembly for `constexpr` and `inline`.

The only one reason to change constructor to `constexpr` that I see, is possibility to create `constexpr CRGB` structures:
```
constexpr CRGB MY_CONSTEXPR_CRGB = CRGB::Green;
```
Currently it will generate compilation error.

Note: `constexpr` includes `inline` and it shouldn't impact to code behavior

PR #1564, opened 2023-11-14
Change CRGB constructors from inline to constexpr

vovagorodok:
Fixes: https://github.com/FastLED/FastLED/issues/1561

### DIFF
--- a/src/pixeltypes.h
+++ b/src/pixeltypes.h
@@ -28,28 +28,28 @@ struct CHSV {
     union {
         struct {
             union {
-                /// Color hue.
+                /// Color hue. 
                 /// This is an 8-bit value representing an angle around
                 /// the color wheel. Where 0 is 0°, and 255 is 358°.
                 uint8_t hue;
                 uint8_t h;  ///< @copydoc hue
             };
             union {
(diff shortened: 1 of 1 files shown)


kriegsman:
Thanks for helping out with FastLED, and thanks for helping nudge the code into more modern directions.   I like the constexpr.  Should some of these be constexpr _and_ inline, or just constexpr?  If the latter, I'll just accept the PR, but I wanted to ask explicitly first.

vovagorodok:
There is a need for additional alignment on my side? Or will it be done in the next iteration?

zackees:
What happens with older versions of Arduino IDE?

Is everyone using C++17 already and something like this would be safe?!?!

My gut feeling is that this would have to be put behind a #define so we don't break old compilers.

kriegsman:
I think if this is C++11, we should go ahead and merge it, and say thanks for helping modernize the code base a little bit.  If it turns out that there's some unintended consequence down the road, we can come back and address that then.

kriegsman:
Thank you very much for helping modernize the code base a bit, and thanks for the help with the FastLED library!

Issue #1541, opened 2023-09-18
FastLED with earlephilhower - redefined errors for sei and cli


bradanlane:
For RP2040, the only Arduino framework to support both cores for use with PlatformIO is [earlephilhower](https://github.com/maxgerhardt/platform-raspberrypi).

This implementation of the Arduino framework recognizes there are two cores and allows `sei()` and `cli()` to work on each core.
It does this by having its own defines for `sei` and `cli`.

The FastLED library also attempts to define `sei` and `cli`. 

Ideally, the FastLED code would check if the definitions already exist. Currently, it generates compiler warnings.

```
In file .../FastLED/src/led_sysdefs.h:38,
                 from .../FastLED/src/FastLED.h:55,
                 from src/leds.h:13,
                 from src/main.cpp:30:
.../FastLED/src/platforms/arm/rp2040/led_sysdefs_arm_rp2040.h:87: warning: "cli" redefined
    #define cli() (saved_interrupt_status = save_and_disable_interrupts())

In file .../.platformio/packages/framework-arduinopico/cores/rp2040/Arduino.h:74:
note: this is the location of the previous definition
    #define cli() noInterrupts()

In file .../FastLED/src/led_sysdefs.h:38,
                 from .../FastLED/src/FastLED.h:55,
                 from src/leds.h:13,
                 from src/main.cpp:30:
.../FastLED/src/platforms/arm/rp2040/led_sysdefs_arm_rp2040.h:88: warning: "sei" redefined
    #define sei() (restore_interrupts(saved_interrupt_status))

In file .../.platformio/packages/framework-arduinopico/cores/rp2040/Arduino.h:73: 
note: this is the location of the previous definition
    #define sei() interrupts()
```

JonnyHaystack:
Wrapping in `#ifndef cli` and `#ifndef sei` or checking for any other defines doesn't seem to work, because `Arduino.h` is included *after* `led_sysdefs_arm_rp2040.h` in `led_sysdefs.h`. I think the `Arduino.h` include needs to be moved above the block of platform-specific includes in `led_sysdefs.h`, *or* if this is a problem, it could be included explicitly at the start of `led_sysdefs_arm_rp2040.h`.

PR #1575, opened 2023-12-16
Don't define cli/sei if defined by Arduino core

JonnyHaystack:
Fixes: #1541

### DIFF
--- a/src/platforms/arm/rp2040/led_sysdefs_arm_rp2040.h
+++ b/src/platforms/arm/rp2040/led_sysdefs_arm_rp2040.h
@@ -3,6 +3,12 @@
 
 #include "hardware/sync.h"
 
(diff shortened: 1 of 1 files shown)


JonnyHaystack:
Note: It may be better to move the `Arduino.h` include in `led_sysdefs.h` above all the platform-specific includes, and add this same check to the other platforms' led sysdef headers as well, but because I'm not sure if that could have any negative impact, I've kept the changes localised to `led_sysdefs_arm_rp2040.h` as that's the platform I'm actively developing for and am most familiar with.

kriegsman:
Thank you very much for this PR, and for helping get the SEI/CLI situation a little more compatible -- and thanks for helping with the FastLED library!

Issue #1481, opened 2023-02-20
"Error: invalid offset, value too big" when building for Raspberry Pi Pico (RP2040)


cinderblock:
I've been trying to get FastLED compiling with the PicoSDK for a custom RP2040 based board.

I've got it mostly working but I'm getting what I think is a bug in the m0 clockless assembly implementation.

I was originally working on this build chain in a [pico-sdk issue](https://github.com/raspberrypi/pico-sdk/issues/1234), if you care to look at that.
To summarize:

The main `CMakeLists.txt`, setup mostly as the pico-sdk prescribes:

```cmake
cmake_minimum_required(VERSION 3.13)

include(lib/SDK/pico_sdk_init.cmake)

project(my_project)

# initialize the Raspberry Pi Pico SDK
pico_sdk_init()

# TODO: Use `add_subdirectory(src)` instead of the following
add_executable(my_executable
    src/main.cpp
)

target_link_libraries(my_executable FastLED)

target_link_libraries(my_executable pico_stdlib)

target_link_libraries(my_executable
    pico_stdio
    hardware_spi
    # hardware_uart
    hardware_i2c
    hardware_gpio
)
 
add_subdirectory(lib/FastLED)

# create map/bin/hex/uf2 file in addition to ELF.
pico_add_extra_outputs(my_executable)
```

A `lib/FastLED/CMakeLists.txt` for the FastLED "`INTERFACE`" library:

```cmake
add_library(FastLED INTERFACE)

target_sources(FastLED INTERFACE
    compat/FastLED.cpp
)

# Required pico SDK libraries
target_link_libraries(FastLED INTERFACE
    hardware_sync
    hardware_gpio
    hardware_clocks
    hardware_dma
    hardware_pio
    cmsis_core
    pico_malloc
    pico_time
)

# Ensure our compatiblity header is in the include search path before FastLED's sources
target_include_directories(FastLED INTERFACE
    compat/include
)

target_compile_definitions(FastLED INTERFACE
    # FASTLED_RP2040_CLOCKLESS_PIO=1
)
```

Notably, since FastLED expects to be built in the Arduino environment, some functions need to be declared (and defined inline).
We can declare and implement what's needed with `compat/Adapter.hpp`:

```hpp
#include <stdlib.h>
#include <pico/time.h>
#include <RP2040.h>

// FastLED tests for this to enable RP2040 platform
#define ARDUINO_ARCH_RP2040

[...]

Issue #1571, opened 2023-12-07
On RP2040, multiple calls to addLeds() results in "Error: invalid offset, value too big"


rah2501:
Compiling any of the examples in the "Multiple" subdirectory results in an error similar to this:

```
{standard input}: Assembler messages:
{standard input}:1100: Error: invalid offset, value too big (0x00000494)
```

Commenting out all but one of the calls to addLeds() results in a successful compilation in each case.

samguyer:
Not a lot of information! Could it be a compiler bug? Do we even have any assembly code in the ARM platform implementation?

rah2501:
> Not a lot of information!

Enough to reproduce the issue.

> Could it be a compiler bug?

Could be.

> Do we even have any assembly code in the ARM platform implementation?

FastLED contains Arm assembly code, yes.

rah2501:
> with basically zero information ... it's hard to know where to look

If you want to work on this issue, I would suggest that reproducing the error for yourself would be the first step.

samguyer:
Remind me again: what is the exact configuration you're using? Are you using the Arduino IDE? What version and what board?

rah2501:
> Are you using the Arduino IDE?

The error occurs compiling with both the Arduino IDE and PlatformIO.

> What version and what board?

Arduino 1.8.19, Raspberry Pi Pico/RP2040 3.6.2, FastLED 3.6.0; board Raspberry Pi Pico

PlatformIO 6.1.1, framework-arduinopico 1.30600.0+sha.41b0686, FastLED 3.6.0; board pico

rah2501:
> It looks like we've got multiple issues for this problem already.

Bit of a waste of my time getting those version numbers.

samguyer:
Well, it was a bit of a waste of my time to find the links for the existing issues for you. So, I think we're square on that account.

zackees:
This typically happens when asm code is compiled with debug symbols on. Arduino was surprisingly compiling in debug by default in a lot of cases. I'm not sure if this asm generation bug is still lurking. Recently I've moved the RPXXXX builds to an open source version and it seems to have gone away in our tests.

I might suggest moving to platformio and using the [@max](https://github.com/maxgerhardt/maxgerhardt) platform.

Here are some of our platformio.ini build changes from our config file:

    "rpipico2": [
        "platform=https://github.com/maxgerhardt/platform-raspberrypi.git",
[...]

PR #1576, opened 2023-12-17
feat: option to disable M0 clockless fallback on RP2040

JonnyHaystack:
Fixes: #1481, #1571

When building for RP2040, at least with -O3 optimizations and the arduino-pico core, the M0 clockless assembly code fails to be assembled for RP2040. This code is only included to be used as a fallback if it fails to claim a DMA channel, which is an undesirable situation to be in anyway.

To keep everyone happy, I've added an option `FASTLED_RP2040_CLOCKLESS_M0_FALLBACK` which is 1 by default but can be set to 0 to disable this fallback and skip compiling `m0clockless.h`. This means the default behaviour remains the same (not breaking backwards compatibility), but gives users the ability to disable the fallback if it's causing them issues.

### DIFF
--- a/src/platforms/arm/rp2040/clockless_arm_rp2040.h
+++ b/src/platforms/arm/rp2040/clockless_arm_rp2040.h
@@ -2,7 +2,10 @@
 #define __INC_CLOCKLESS_ARM_RP2040
 
 #include "hardware/structs/sio.h"
+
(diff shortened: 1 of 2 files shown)


kriegsman:
Thank you for this PR, and for figuring out what the issue was here - and making a good work-around, including the #define option in case anyone wants that.  Much appreciated, and thanks for the help with the FastLED library!

PR #1577, opened 2023-12-17
Silence -Wstack-usage for noise.cpp

JonnyHaystack:
The `fill_...()` functions in `noise.cpp` trigger "warning: stack usage 
might be unbounded [-Wstack-usage=]" if you have -Wstack-usage enabled. There clearly are bound checks in the code, so this warning should be suppressed.

### DIFF
--- a/src/noise.cpp
+++ b/src/noise.cpp
@@ -6,6 +6,8 @@
 #include "FastLED.h"
 #include <string.h>
 
+// Compiler throws a warning about stack usage possibly being unbounded even
+// though bounds are checked, silence that so users don't see it
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wstack-usage="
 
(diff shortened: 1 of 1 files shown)


kriegsman:
Thanks for working on this issue about the stack space warning.  
It looks to me like there might be a typo: on line 9 in the PR it says 
`#pragma GCC diagnostic push`
and then on line 862 it says
`#pragma GCC diagnostic push`

Should the second one be `pop` instead?

JonnyHaystack:
Oops, yes it should. Definitely had it as pop before so must've accidentally pressed undo or something... Can fix in the morning.

JonnyHaystack:
Managed to fix it from mobile :raised_hands:

kriegsman:
Thanks for this contribution, and thanks for helping out with FastLED!

PR #1581, opened 2023-12-27
Using correct data types when starting SPI bus for ESP32.

pauldotknopf:
I need to override these static/global variables to have a "-1", but the types don't allow it.

Like this:

spiMosi = LED_DATA_PIN;
spiClk = LED_CLOCK_PIN;
spiMiso = -1;
spiCs = -1;
static APA102Controller<LED_DATA_PIN, LED_CLOCK_PIN, BGR,  DATA_RATE_MHZ(10)> c; FastLED.addLeds(&c, _leds, NUM_OF_LEDS);

### DIFF
--- a/src/platforms/esp/32/fastspi_esp32.h
+++ b/src/platforms/esp/32/fastspi_esp32.h
@@ -55,15 +55,15 @@ static SPIClass ledSPI(FASTLED_ESP32_SPI_BUS);
 
 
 #if FASTLED_ESP32_SPI_BUS == VSPI
-    static uint8_t spiClk = 18;
-    static uint8_t spiMiso = 19;
-    static uint8_t spiMosi = 23;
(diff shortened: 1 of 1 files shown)


pauldotknopf:
Also, I think [this](https://github.com/FastLED/FastLED/blob/3adbeb3fed67d4ee1681af546d86090a0cddf7b1/src/platforms/esp/32/fastspi_esp32.h#L90) line should be using the ```DATA_PIN``` and ```CLOCK_PIN```. Is that a correct assumption?

It should be:

```c++
ledSPI.begin(CLOCK_PIN, -1, DATA_PIN, -1);
```

samguyer:
That looks correct to me -- the interface for SPIClass::begin defines those parameters as int8_t.
I'll merge it in. Thank for your contribution!

PR #1583, opened 2024-01-02
Fixes five_bit_gamma - scale16b8 global fix for brightness == 0 and improve user overrides

zackees:
This is a follow up CL to fix five_bit_gamma brightness when brightness == 0 and also to improve user overrides in platformio. This was found while using the Teensy 4.1 platform.

# brightness == 0

It turns out that `uint16_t scale16by8( uint16_t i, fract8 scale )` doesn't work correctly when scale==0 and `FASTLED_SCALE8_FIXED == 1`, which is the case with Teensy 4.1.

This is because there is a +1 that is added on to the scaling factor so that scale16by8(0xffff, 0xff) == 0xffff (well, at least that's the documentation for `scale8(...)`, I'm not sure it applies to `scale16by8(...)`, but I assume it does). However this factor results in a non zero value when using `scale16b8(i, 0)`.

Therefore, an early test is performed where if scale == 0, then zero is immediately returned.

To be honest, I could have fixed this in just my own code but this seems to be a bug that could affect other clients so it's probably better that the fix to `scale16by8` is applied globally.

# improve user override

It turns out that platformio doesn't do static linking in the correct order to allow a strong symbol in client code to override a weak symbol in the fastled library. Therefore I've added two `#define` statements that turn on user override by conditionally declaring `five_bit_hd_gamma_function()` and `five_bit_hd_gamma_bitshift()` as extern functions. Now if the user doesn't define these replacement functions then this will result in a linker error, but only if the proper defines are entered into the build system. The defines are:

  * `FASTLED_FIVE_BIT_HD_BITSHIFT_FUNCTION_OVERRIDE`
  * `FASTLED_FIVE_BIT_HD_GAMMA_FUNCTION_OVERRIDE`

### DIFF
--- a/src/five_bit_hd_gamma.cpp
+++ b/src/five_bit_hd_gamma.cpp
@@ -11,10 +11,9 @@
 
 FASTLED_NAMESPACE_BEGIN
 
-
+#ifndef FASTLED_FIVE_BIT_HD_GAMMA_BITSHIFT_FUNCTION_OVERRIDE
 #ifndef FASTLED_FIVE_BIT_HD_GAMMA_FUNCTION_2_8
 // Fast a memory efficient gamma=2 function.
-__attribute__((weak)) 
 void five_bit_hd_gamma_function(
   uint8_t r8, uint8_t g8, uint8_t b8,
   uint16_t* r16, uint16_t* g16, uint16_t* b16) {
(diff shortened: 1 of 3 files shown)


zackees:
This PR was broken up into two and merged independently. Therefore this is being closed.

PR #1608, opened 2024-03-10
Issue #1552: Fix to enhance ESP32-S3 SPI compatibility by enabling flexible pin assignment

coryking:
This pull request aims to resolve issue #1552 by adapting FastLED's SPI handling for the ESP32-S3. The ESP32-S3 diverges from previous ESP32 models by not using named SPI buses (VSPI/HSPI). Instead, it supports flexible pin assignments for SPI communication. In response, this update defaults to "FSPI" for the ESP32-S3, allowing dynamic pin assignment rather than relying on hardcoded pin mappings.

I want to note that while this solution aligns FastLED with the ESP32-S3's SPI flexibility, I'm not entirely certain this is the absolute "best practice" within the broader context of esp-idf and the arduino-espressif library ecosystem. If you know a better way, I'm all ears!

### DIFF
--- a/src/platforms/esp/32/fastspi_esp32.h
+++ b/src/platforms/esp/32/fastspi_esp32.h
@@ -53,24 +53,33 @@ FASTLED_NAMESPACE_BEGIN
 
 static SPIClass ledSPI(FASTLED_ESP32_SPI_BUS);
 
-
+#if CONFIG_IDF_TARGET_ESP32S3
+#pragma message "S3 Mode, you tell me the pins, I use them!"
+#define spiMosi DATA_PIN
+#define spiClk CLOCK_PIN
+#define spiMiso -1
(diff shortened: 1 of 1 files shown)


samguyer:
Hello! Thanks for this contribution. A couple of things before I merge it ... I had to pull in an earlier PR that might invalidate yours (or create a bunch of conflicts). It's in the same file, but not directly conflicting with yours. I'm also wondering if you can limit the changes a bit more -- right now, your PR appears to change almost every line in the file.

coryking:
I see those line changes... VSCode probably did something stupid when I saved and i didn't catch it before making this pull request.

coryking:
Going to close this pull request and roll with PR #1610. VSCode decided to reformat the entire file in this commit and I didn't catch it!

PR #1611, opened 2024-03-15
Add a couple more things for git to ignore

coryking:
Ignore some cruft my build environment tried to foist upon this project.


kriegsman:
Thank you for this contribution - should make things cleaner and easier for everyone.  Thanks for helping out with FastLED!

PR #1614, opened 2024-04-10
Add SparkFun Thing Plus - RA6M5 port

santaimpersonator:
Provides pin definitions for our board
However, the `defined(ARDUINO_*` portion, is dependent on [our board being added to the Renesas Arduino core](https://github.com/arduino/ArduinoCore-renesas/pull/290)


kriegsman:
Thank you very much for this contribution - and thanks for helping out with FastLED!

Issue #1553, opened 2023-10-21
Blink Arduino sample fails to compile on ESP32S3


sukesh-ak:
Sample used without any changes:
https://github.com/FastLED/FastLED/tree/master/examples/Blink

My custom sample which was working earlier, is not compiling anymore. So tested with the Blink sample and same issues.
You don't need hardware to test it. Set the MCU to S3 device and try compiling.

Compiler Error output. 
```
Compiling libraries...
In file included from c:\Users\sukes\Documents\Arduino\libraries\FastLED\src/FastLED.h:75,
                 from C:\Users\sukes\AppData\Local\Temp\.arduinoIDE-unsaved2023921-15596-mlo18f.socmm\Blink\Blink.ino:5:
c:\Users\sukes\Documents\Arduino\libraries\FastLED\src/fastspi.h:157:23: note: '#pragma message: No hardware SPI pins defined.  All SPI access will default to bitbanged output'
  157 | #      pragma message "No hardware SPI pins defined.  All SPI access will default to bitbanged output"
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
Compiling library "FastLED"
[...]

Issue #1606, opened 2024-03-07
Compiling FastLED blink sketch using FastLED 3.6.0 for an ESP32-C3 with Arduino IDE 1.8.19 seems to fail.


HenkHoldijk:
Trying to build the FastLED blink example for an ESP32-C3 using the Arduino IDE 1.8.19 results in:

- rmt.h that must be replaced with rmt_tx.h and/or rmt_rx.h
- periph_ctrl.h not being used anymore
- Several warnings regarding volatile operands being deprecated
- cpu_ll_get_cycle_count() being deprecated
- incompatible types

When disabling all warnings, still the following is reported:

...\FastLED\src\platforms\esp\32\clockless_rmt_esp32.cpp: In member function 'void ESP32RMTController::startOnChannel(int)':
...\FastLED\src\platforms\esp\32\clockless_rmt_esp32.cpp:266:29: error: 'RMTMEM' was not declared in this scope
  266 |         mRMT_mem_start = & (RMTMEM.chan[mRMT_channel].data32[0].val);
      |                             ^~~~~~

...\FastLED\src\platforms\esp\32\clockless_rmt_esp32.cpp: In static member function 'static void ESP32RMTController::doneOnChannel(rmt_channel_t, void*)':
...\FastLED\src\platforms\esp\32\clockless_rmt_esp32.cpp:347:5: error: 'gpio_matrix_out' was not declared in this scope; did you mean 'gpio_iomux_out'?
  347 |     gpio_matrix_out(pController->mPin, 0x100, 0, 0);
      |     ^~~~~~~~~~~~~~~
      |     gpio_iomux_out

HenkHoldijk:
Hi Sam (?), Adding ugly #ifdefs is not a long term solution. Sooner or later this kind of "swamp" will catch up with you. The ESP32 series has several variations with different behavior (and definitely more will come) an additional platform layer would make sense. I already can see that IO mapping of a C3 is handled different from some other EPS32's. The ESP8266 ("First Of Class") is EOL,
but I guess the ESP32 series will continue for a while (before the ESP64 series is announced ;) ). Several changes in the architecture will appear while using the same well known ESP32 name (exploit "ESP32"). In the end ESP32 is just a name and the SOC architecture will change while still called ESP32. This really calls for some sort of adaptation layer. Cheers, Henk

samguyer:
[...]

Issue #1605, opened 2024-03-04
Compile Issue Caused By pixeltypes.h:7,controller.h:9,FastLED.h:57, & noise.cpp:6:


ylnest2018:
I get this when compiling my project, I have seen other links with a similar issue on this page spreading over a couple years, everyone seems to link it to another issue but im yet to find one with a real solution.

https://github.com/FastLED/FastLED/issues/1553

`Users/leo/Documents/Arduino/libraries/FastLED/src/noise.cpp: In function 'void fill_raw_2dnoise8(uint8_t*, int, int, uint8_t, q44, fract8, int, uint16_t, int, uint16_t, int, uint16_t)':
/Users/leo/Documents/Arduino/libraries/FastLED/src/noise.cpp:623:100: error: ambiguous overload for 'operator*' (operand types are 'q44' {aka 'q<unsigned char, 4, 4>'} and 'int')
  623 |     fill_raw_2dnoise8(pData, width, height, octaves-1, freq44, amplitude, skip+1, x*freq44, freq44 * scalex, y*freq44, freq44 * scaley, time);
      |                                                                                             ~~~~~~ ^ ~~~~~~
      |                                                                                             |        |
      |                                                                                             |        int
      |                                                                                             q44 {aka q<unsigned char, 4, 4>}
In file included from /Users/leo/Documents/Arduino/libraries/FastLED/src/pixeltypes.h:7,
                 from /Users/leo/Documents/Arduino/libraries/FastLED/src/controller.h:9,
                 from /Users/leo/Documents/Arduino/libraries/FastLED/src/FastLED.h:57,
                 from /Users/leo/Documents/Arduino/libraries/FastLED/src/noise.cpp:6:
/Users/leo/Documents/Arduino/libraries/FastLED/src/lib8tion.h:909:14: note: candidate: 'uint32_t q<T, F, I>::operator*(uint32_t) [with T = unsigned char; int F = 4; int I = 4; uint32_t = long unsigned int]'
  909 |     uint32_t operator*(uint32_t v) { return (v*i) + ((v*f)>>F); }
      |              ^~~~~~~~
/Users/leo/Documents/Arduino/libraries/FastLED/src/lib8tion.h:911:14: note: candidate: 'uint16_t q<T, F, I>::operator*(uint16_t) [with T = unsigned char; int F = 4; int I = 4; uint16_t = short unsigned int]'
[...]

PR #1620, opened 2024-05-02
Fix rmt driver for newer versions of idf / freertos

shlomozippel:
This fixes the RMT output for ESP32 when using newer esp32 arduino core (3.x) which uses esp idf 5.x and newer FreeRTOS.

This can most charitably be described as a surgical hack. The right way would be to rewrite using the new RMT interface instead of patching the old one

Sanity tested on ESP32 with default RMT settings

Partially fixes issues #1553, #1605

### DIFF
--- a/src/platforms/esp/32/clockless_rmt_esp32.cpp
+++ b/src/platforms/esp/32/clockless_rmt_esp32.cpp
@@ -27,7 +27,11 @@ static intr_handle_t gRMT_intr_handle = NULL;
 
 // -- Global semaphore for the whole show process
 //    Semaphore is not given until all data has been sent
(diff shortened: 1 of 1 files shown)


shlomozippel:
This also fixes #1606

samguyer:
@shlomozippel This looks fantastic! Thank you! I've been meaning to get this work done for a while. Looks like a completely reasonable hack to me. :-)

shlomozippel:
Yeah, it's been on my radar for a while. I just got my third project that required idf 5.x so finally bit the bullet :)

zackees:
Reviewed.

LGTM as is.

Issue #1631, opened 2024-06-08
Issues with ESP32-S2 Release v3.0.1(fastled 3.7)


marcobrianza:
Hello, building the example blink sketch on a loin esp32-s2 does not work anymore upgrading to ESP32 core 3.0.1 and fasteld 3.7

the errors are the following:

/Users/marco/Nextcloud-marco/Arduino/libraries/FastLED/src/platforms/esp/32/clockless_rmt_esp32.cpp: In member function 'void ESP32RMTController::tx_start()':
/Users/marco/Nextcloud-marco/Arduino/libraries/FastLED/src/platforms/esp/32/clockless_rmt_esp32.cpp:339:37: error: 'volatile union rmt_chnconf1_reg_t' has no member named 'mem_rd_rst'
  339 |     RMT.conf_ch[mRMT_channel].conf1.mem_rd_rst = 1;
      |                                     ^~~~~~~~~~
/Users/marco/Nextcloud-marco/Arduino/libraries/FastLED/src/platforms/esp/32/clockless_rmt_esp32.cpp:340:37: error: 'volatile union rmt_chnconf1_reg_t' has no member named 'mem_rd_rst'
  340 |     RMT.conf_ch[mRMT_channel].conf1.mem_rd_rst = 0;
      |                                     ^~~~~~~~~~
/Users/marco/Nextcloud-marco/Arduino/libraries/FastLED/src/platforms/esp/32/clockless_rmt_esp32.cpp:347:37: error: 'volatile union rmt_chnconf1_reg_t' has no member named 'tx_start'
  347 |     RMT.conf_ch[mRMT_channel].conf1.tx_start = 1;
      |                                     ^~~~~~~~
/Users/marco/Nextcloud-marco/Arduino/libraries/FastLED/src/platforms/esp/32/clockless_rmt_esp32.cpp: In static member function 'static void ESP32RMTController::doneOnChannel(rmt_channel_t, void*)':
/Users/marco/Nextcloud-marco/Arduino/libraries/FastLED/src/platforms/esp/32/clockless_rmt_esp32.cpp:400:32: error: 'volatile union rmt_chnconf1_reg_t' has no member named 'tx_stop'
  400 |     RMT.conf_ch[channel].conf1.tx_stop = 1;
      |                                ^~~~~~~
[...]

PR #1634, opened 2024-06-13
Fix RMT driver member names for ESP32S3

acalatrava:
This should fix https://github.com/FastLED/FastLED/issues/1631


zackees:
Nope, this code is wrong. You don't take into consideration older toolchains. Anyone with an older ArduinoIDE or platformio settings will fail.

zackees:
Here's the correct implementation:

https://github.com/FastLED/FastLED/pull/1652

PR #1636, opened 2024-06-21
Update README.md

CaHG-eiPalOpto:
Included Würth Elektronik IC LEDs in the Readme. These LEDs have been tested with FastLED library and they can be controlled same as the digital LEDs from other companies.


CaHG-eiPalOpto:
Thanks for your input. 

As the product manager for these products at Würth Elektronik, it has been requested that this visibility of ICLEDs on the FastLED is crucial for their purchasing decisions and ensuring compatibility with their projects.

Could you specify the type of partnership you're looking for so we can push this request into the main branch?

For example my proposal will be to modify the .h and the .cpp files  adding code fields like those for WorldSemi. This will enable users to input Würth part numbers directly into FastLED and control the ICLELDs

PR #1643, opened 2024-07-14
Update URL for cpt-city

uberjay:
> The original URL for cpt-city now redirects to a mirror of the site, but it only redirects to the base URL rather than the full pathname. This behaviour breaks paletteknife.
> 
> The changes here point paletteknife directly to the mirror, which resolves the problem for me.

Thank you to @sutaburosu for these changes, which were originally submitted here: FastLED/fastled-fastled.github.io#1


sutaburosu:
Well I'll be...
Thanks for your attention to detail here, @uberjay. I can confirm that this has now fixed the URLs served on fastled.io. Nice work.

Issue #1438, opened 2022-10-10
ESP32: FastLED.show hangs when core 1 is under load


Ezekiel-DA:
Hi all,

I am working on a project that uses FastLED to drive a significant number of WS2812B LEDs while performing quite a bit of other work, namely audio and BLE.

For a little bit of context (and, I will admit, a shameless attempt at trying to get more help ASAP before Halloween!), the actual "finished" project is a costume for a kiddo in a wheelchair, as part of a charity project I and several coworkers donate time to via the awesome non profit [Magic Wheelchair](https://www.magicwheelchair.org/).

You can see some shots of the costume in action and the eye-searing glory of close to 1500 LEDs, all on battery power, at an early Halloween parade in Salem, MA, in [this news report](https://whdh.com/news/salem-girl-in-wheelchair-receives-custom-costume-joins-in-annual-parade/).

Now for the issue I am attempting to figure out: as the code stands right now, when mp3 playback starts, FastLED essentially "crashes": FastLED.show() no longer returns, which blocks the main task (using the Arduino framework in Platform.io's ESP32 Platform support).

If I move the call to FastLED.show() to a separate task, that task stops running, while the main loop() task continues to run, predictably.

Hunting down a minimal reproduction for this has been a little tricky:
* without mp3 playback, FastLED behaves normally. I attempted to simulate load on Core 1 with something simpler than mp3 playback but was unsuccessful
* with fewer than 3 FastLED controllers (only 1 or 2 calls to addLeds()), the problem also stops occurring
* the number of LEDs added also seems to play a role

This is the most minimal reproduction I was able to create: https://github.com/Ezekiel-DA/FastLED_audio_hang_repro

It still relies on bringing in a whole external [I2S / mp3 decoding library](https://github.com/schreibfaul1/ESP32-audioI2S), but it is otherwise fairly simple, does not require any of the external hardware that the actual costume uses, etc. You do not need to connect any LEDs or connect anything that would allow actual sound output to see the problem occur.

To reproduce:
* clone [this repo](https://github.com/Ezekiel-DA/FastLED_audio_hang_repro) and open the project with Platform.io
* run the Upload Filesystem Image command in Platform.io to upload the sample mp3 file to flash
* upload the actual code
* observe that, once music playback starts (automatically after a set number of loops through loop()), looping through the task that runs FastLED.show() stops

Thanks in advance for any help!

Ezekiel-DA:
More information:

[...]

PR #1651, opened 2024-08-04
clockless_rmt_esp32 FastLED.show() hang fix

davidlmorris:
# clockless_rmt_esp32 FastLED.show() infrequent hang fix

See issue 1438 (https://github.com/FastLED/FastLED/issues/1438) and others.  FastLED.show() does not return (hangs) very infrequently, sometimes many hours after starting usually while under load.

See https://github.com/davidlmorris/FastLED_Hang_Fix_Demo for a demonstration of this problem (noting that the hang will happen extremely infrequently).

The FastLed RMT driver makes extensive use of interrupts. The comments in the FastLED driver hint at least one per 32 bits of data out. The driver does all the sending in one hit so that even with multiple controllers (as we have in this demo) it is only when the last one is called does it finally does the work. I am uncertain what might happen if only one controller that is not the last one is called. (Something for a future experiment?). On a send, it waits 50 microseconds to signal a start, then starts the channels for each controller. It then sets a semaphore which is then released when the final channel on the final controller is complete. It appears that rarely something goes wrong here, and presumably, that interrupt is swallowed by the Esp32 system. The result is that ESP32RMTController::showPixels() waits forever for the semaphore to be released (which it never will be) so we have a hang. Note that this hang will occur minutes or hours into a run, depending on other factors (like load or power), and is essentially out of the control of the FastLED driver.

The quick and dirty fix is simply to assign some timeout value to the semaphore take rather than portMAX_DELAY (line 204 for version 3.6.0 and line 223 for version 3.7.0).

We do this in the modified FastLED lib forked [here](https://github.com/davidlmorris/FastLED) by defining FASTLED_RMT_MAX_TICKS_FOR_GTX_SEM to a value like (2000/portTICK_PERIOD_MS) before we call FastLED.h (see [clockless_rmt_esp32.h](https://github.com/davidlmorris/FastLED/tree/master/src/platforms/esp/32/clockless_rmt_esp32.h)).

Alternatively, we can call GiveGTX_sem(); which has been added to [clockless_rmt_esp32.cpp](https://github.com/davidlmorris/FastLED/tree/master/src/platforms/esp/32/clockless_rmt_esp32.cpp) so we can programmatically decide on the wait time.

### DIFF
--- a/src/platforms/esp/32/clockless_rmt_esp32.cpp
+++ b/src/platforms/esp/32/clockless_rmt_esp32.cpp
@@ -58,6 +58,15 @@ typedef struct {
 extern rmt_block_mem_t RMTMEM;
 #endif
 
+
+void IRAM_ATTR GiveGTX_sem()
+{
+    if (gTX_sem != NULL)
+        {
+        xSemaphoreGive(gTX_sem);
+        }
+}
+
 ESP32RMTController::ESP32RMTController(int DATA_PIN, int T1, int T2, int T3, int maxChannel, int memBlocks)
     : mPixelData(0), 
       mSize(0), 
@@ -198,7 +207,7 @@ void IRAM_ATTR ESP32RMTController::showPixels()
     gNumStarted++;
 
     // -- The last call to showPixels is the one responsible for doing
-    //    all of the actual worl
+    //    all of the actual work
     if (gNumStarted == gNumControllers) {
         gNext = 0;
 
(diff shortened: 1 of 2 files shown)


zackees:
Thanks for this, looks like a good work around a very detailed description (thanks!)

Looking for @samguyer for final approval.

samguyer:
@davidlmorris I am running the CICD tests, but this change looks good to me. I have been intending to look more closely at the whole synchronization strategy for the ESP32 driver, but so far I haven't had time. So, this is a welcome fix!

davidlmorris:
Hi and thank you both for your comments.

I've been playing with the test code this morning trying to trigger hang-ups more frequently.  Doing this seems seems risk 'Guru meditations'.  I have some updated versions here https://github.com/davidlmorris/FastLED_Hang_Fix_Demo.

One other thing worth mentioning is that by default this change does nothing.  That is, the GTX_sem semaphore by default has a timeout of portMAX_DELAY, so a FastLED program needs to set FASTLED_RMT_MAX_TICKS_FOR_GTX_SEM to something less than that to activate this fix.  It might be worth considering reversing this so that we default to say  (2000/portTICK_PERIOD_MS) and have an option for portMAX_DELAY since there will be many application developers who are unaware of this issue.  (I am happy to make this change if you want - but my suggestion is to fold this in first and fix the default later).

Issue #1653, opened 2024-08-09
FastLED does not compile for ESP32S3 under Arduino IDE 2.3.2


DasElvetritsche:
If i compile the sketch without line 1 "#define FASTLED_ALL_PINS_HARDWARE_SPI" i get the message rearding bit banging and all is OK. If i include it i get the following errors:

##################################################################
In file included from c:\TMP\Arduino\libraries\FastLED\src/platforms/esp/32/fastled_esp32.h:6,
                 from c:\TMP\Arduino\libraries\FastLED\src/platforms.h:42,
                 from c:\TMP\Arduino\libraries\FastLED\src/FastLED.h:66,
                 from C:\TMP\Arduino\LED-colors\LED-colors.ino:2:
c:\TMP\Arduino\libraries\FastLED\src/platforms/esp/32/fastspi_esp32.h:2:17: note: #pragma message: ESP32 Hardware SPI support added
 #pragma message "ESP32 Hardware SPI support added"
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from c:\TMP\Arduino\libraries\FastLED\src/platforms/esp/32/fastled_esp32.h:6,
                 from c:\TMP\Arduino\libraries\FastLED\src/platforms.h:42,
                 from c:\TMP\Arduino\libraries\FastLED\src/FastLED.h:66,
                 from C:\TMP\Arduino\LED-colors\LED-colors.ino:2:
c:\TMP\Arduino\libraries\FastLED\src/platforms/esp/32/fastspi_esp32.h:54:18: note: #pragma message: Targeting ESP32S3, which has better SPI support. Configuring for flexible pin assignment.
  #pragma message "Targeting ESP32S3, which has better SPI support. Configuring for flexible pin assignment."
                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
c:\TMP\Arduino\libraries\FastLED\src/platforms/esp/32/fastspi_esp32.h: In instantiation of 'static void ESP32SPIOutput<DATA_PIN, CLOCK_PIN, SPI_SPEED>::writeWord(uint16_t) [with unsigned char DATA_PIN = 40; unsigned char CLOCK_PIN = 39; unsigned int SPI_SPEED = 10000000; uint16_t = short unsigned int]':
[...]

PR #1670, opened 2024-08-17
fixes 1653 compiler error for Apa102 hardware spi pins on ESP32-S3

zackees:
Fixes https://github.com/FastLED/FastLED/issues/1653

The compiler fix can be verified by running

`ci/ci-compile --boards esp32-s3-devkitc-1 --examples _bugs/1653_S3_Apa102_Hardware_Pins`

Note that I don't actually know if this works or not, I just fixed the compiler errors.

### DIFF
--- a/src/chipsets.h
+++ b/src/chipsets.h
@@ -92,8 +92,9 @@ class LPD8806Controller : public CPixelLEDController<RGB_ORDER> {
 	public:
 		// LPD8806 spec wants the high bit of every rgb data byte sent out to be set.
 		__attribute__((always_inline)) inline static uint8_t adjust(FASTLED_REGISTER uint8_t data) { return ((data>>1) | 0x80) + ((data && (data<254)) & 0x01); }
-		__attribute__((always_inline)) inline static void postBlock(int len) {
-			SPI::writeBytesValueRaw(0, ((len*3+63)>>6));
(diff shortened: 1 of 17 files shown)


DasElvetritsche:
Didnt work for me. When i make the changes like in the attache files it works.
[chipsets.h.txt](https://github.com/user-attachments/files/16667924/chipsets.h.txt)
[fastspi.h.txt](https://github.com/user-attachments/files/16667927/fastspi.h.txt)
[fastspi_esp32.h.txt](https://github.com/user-attachments/files/16667928/fastspi_esp32.h.txt)
[ModifiedLibraryFiles.txt](https://github.com/user-attachments/files/16667929/ModifiedLibraryFiles.txt)

But when using in a scetch with TFT_eSPI lib something in the TFT part crasches.

DasElvetritsche:
[fastspi.h.txt](https://github.com/user-attachments/files/16686175/fastspi.h.txt)
[fastspi_esp32.h.txt](https://github.com/user-attachments/files/16686176/fastspi_esp32.h.txt)
Thise 2 files i still have to modify when i use all files new from the repo. With this mods in there it runs. 
But when using TFT_eSPI lib in the same sketch it looks like the having a confilct on the SPI at runtime.
I had to change some of the #if because it works for me only that way. Otherwise the right branches do not get compiled for me.
Hope my info helps.

PR #1687, opened 2024-08-30
Add ColorFromPaletteExtended function for higher precision

zackees:
courtesy of https://github.com/sutaburosu for the code solution.

See also.
https://github.com/FastLED/FastLED/pull/1659

### DIFF
--- a/src/colorutils.cpp
+++ b/src/colorutils.cpp
@@ -564,6 +564,50 @@ inline uint8_t lsrX4( uint8_t dividend)
     return dividend;
 }
 
+CRGB ColorFromPaletteExtended(const CRGBPalette32& pal, uint16_t index, uint8_t brightness, TBlendType blendType) {
+  // Extract the five most significant bits of the index as a palette index.
+  uint8_t index_5bit = (index >> 11);
+  // Calculate the 8-bit offset from the palette index.
+  uint8_t offset = (uint8_t)(index >> 3);
+  // Get the palette entry from the 5-bit index
+  const CRGB* entry = &(pal[0]) + index_5bit;
+  uint8_t red1   = entry->red;
+  uint8_t green1 = entry->green;
+  uint8_t blue1  = entry->blue;
+
+  uint8_t blend = offset && (blendType != NOBLEND);
+  if (blend) {
+    if (index_5bit == 31) {
+      entry = &(pal[0]);
+    } else {
+      entry++;
+    }
+
+    // Calculate the scaling factor and scaled values for the lower palette value.
+    uint8_t f1 = 255 - offset;
+    red1   = scale8_LEAVING_R1_DIRTY(red1,   f1);
+    green1 = scale8_LEAVING_R1_DIRTY(green1, f1);
+    blue1  = scale8_LEAVING_R1_DIRTY(blue1,  f1);
+
+    // Calculate the scaled values for the neighbouring palette value.
+    uint8_t red2   = entry->red;
+    uint8_t green2 = entry->green;
+    uint8_t blue2  = entry->blue;
+    red2   = scale8_LEAVING_R1_DIRTY(red2,   offset);
+    green2 = scale8_LEAVING_R1_DIRTY(green2, offset);
+    blue2  = scale8_LEAVING_R1_DIRTY(blue2,  offset);
+    cleanup_R1();
+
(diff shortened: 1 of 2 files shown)


zackees:
https://github.com/sutaburosu please feel free to pull these changes and re-submit as your own pr

zackees:
Fixes:

https://github.com/FastLED/FastLED/pull/202

sutaburosu:
It's not me who should be credited for this. It was @generalelectrix who originally suggested this improvement and provided the implementation for `CRGBPalette16`. I just added the code for 32-colour palettes. I guess, for completeness, a version should also be added for 256-colour palettes.

I'm no expert in C++, but is it not possible to keep the same function name for the Extended version and have the compiler select which version to use based on the types? The original function takes a uint8_t index into the palette, and the extended version takes uint16_t.

generalelectrix:
@sutaburosu  our 9-year collaboration has come to fruition 🤣

zackees:
I don't like selecting templates that auto parameterize on coercable types.

You don't know what you are going to get.

If anything there should be a define that eliminates the older version for non avr chipsets.

Are you that comfortable with this gradient that I can do this?

sutaburosu:
> I don't like selecting templates that auto parameterize on coercable types.

I don't think templates are being used here. This should be plain old function overloading. My understanding is that if an exact match of types exists, the relevant function will be chosen. If this fails to match any functions, chars and shorts would be promoted to int, which will then fail to match any function signatures. Perhaps I am wrong in this?

> If anything there should be a define that eliminates the older version for non avr chipsets.

I don't understand your reasoning here. If sketches aren't using the 8-bit indexing version, it will be absent from the binary. Why does AVR need different behaviour than other platforms?

Let users choose whether they want 8-bit or 16-bit indexing based on the type of the `index` parameter. This keeps backwards compatibility with existing sketches, and permits the variant to be chosen on a case-by-case basis.

> Are you that comfortable with this gradient that I can do this?

I've used the 16-colour palette variant extensively on AVR for 4 years. I've never noticed a problem with it. I have very rarely used the 32-colour palette version, so I have far less confidence in it.

sutaburosu:
@zackees [this sketch](https://wokwi.com/projects/407831886158110721) has the missing variant for 256-colour RGB palettes. I've tested all three (16, 32 and 256 colour palette variants) in multiple sketches today, and they all seem to perform flawlessly.

See the comments at the top of that sketch for benchmark results. In short, on AVR the 16-bit versions are 35%, 23% and 136% (for good reasons) slower. On platforms that can do arbitrary bit-shifts, rather than only 1-bit at a time like AVR, I would expect the performance penalty to be far lower.

That same sketch uses the function overloading idea that I suggested. Feel free to change it as you see fit. If you feel the need for the FastLED API to be different on AVR, I'll just keep including my preferred version in my sketches as I have for several years now.

As I reminder, I would prefer if you credit @generalelectrix for this work. It was their observation, and their implementation before I became obsessive over the details of it.

I hope this information is useful to you. It would be great if some form of these suggestions could make it into tomorrow's release, rather than the 32-colour palette only version that currently stands in the root branch of the repo.

All the best. :wine_glass:

zackees:
> That same sketch uses the function overloading idea that I suggested. Feel free to change it as you see fit. If you feel the need for the FastLED API to be different on AVR, I'll just keep including my preferred version in my sketches as I have for several years now.

If someone uses a 16 bit index on avr then they automatically get the new Palette function. I'm not a big fan of that as it's non obvious.

Are you really using AVR that much?

sutaburosu:
> Are you really using AVR that much?

My projects are almost exclusively AVR. I use Teensy 4.x when I need something bigger and faster, but easily >95% of my projects use USD$0.80 Nano clones. Apart from the price, I very much enjoy the challenge of getting pleasing results at high frame rates with such limited specifications.

I notice the recent release of FastLED is missing the 16-colour palette variant of ColorFromPaletteExtended. There is well tested code in [that sketch I linked previously](https://wokwi.com/projects/407831886158110721).

zackees:
16-colour palette extended added.

zackees:
Okay I didn't realize you were an AVR expert. I'm willing to revisit this and move *Extended to the normal function calls... IF and only if it doesn't break my compiled size tests for AVR.

Can you confirm the performance is reasonable in comparison to bit banging the let's out?

PR #1690, opened 2024-08-31
improved simplex noise

zackees:
Optimizations lifted from https://github.com/aykevl

### DIFF
--- a/src/simplex.cpp
+++ b/src/simplex.cpp
@@ -114,7 +114,7 @@ uint16_t snoise16(uint32_t x) {
 
     int32_t n = n0 + n1;   // .15
     n += 2503;             // .15: fix offset, adjust to +0.03
-    n = (n << 14) / 40225; // .15: fix scale to fit in [-1,1]
+    n = (n * 26694) >> 16; // .15: fix scale to fit in [-1,1]
     return uint16_t(n) + 0x8000;
 }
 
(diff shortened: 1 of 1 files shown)


src/simplex.cpp
```diff
@@ -181,7 +181,7 @@ uint16_t snoise16(uint32_t x, uint32_t y) {
     // Add contributions from each corner to get the final noise value.
     // The result is scaled to return values in the interval [-1,1].
     int32_t n = n0 + n1 + n2;    // .30
-    n = (n << 6) / 46360; // fix scale to fit exactly in an int16_t
+    n = ((n >> 8) * 23163) >> 16; // fix scale to fit exactly in an int16
```
aykevl, line 184: I guess you meant `int16_t`?

```suggestion
    n = ((n >> 8) * 23163) >> 16; // fix scale to fit exactly in an int16_t
```

Same for 3D noise (which was also wrong in my original FastLED PR).

src/simplex.cpp
```diff
@@ -313,7 +313,7 @@ uint16_t snoise16(uint32_t x, uint32_t y, uint32_t z) {
     // Add contributions from each corner to get the final noise value.
     // The result is scaled to stay just inside [-1,1]
     int32_t n = n0 + n1 + n2 + n3; // .30
-    n = (n << 6) / 64120;          // fix scale to fit exactly in an int16
+    n = ((n >> 8) * 16748) >> 16 ; // fix scale to fit exactly in an int16
```
aykevl, line 316:
```suggestion
    n = ((n >> 8) * 16748) >> 16;  // fix scale to fit exactly in an int16_t
```

aykevl (commented):
Change looks the same. Untested (not using FastLED at the moment) but it looks reasonable.

A few nitpicks below, feel free to ignore.

PR #1699, opened 2024-09-03
Remove unneccessary CI files for RP2040/RP2350

maxgerhardt:
With https://github.com/maxgerhardt/platform-raspberrypi/pull/72 merged, you don't need to locally store the board definitions in the repo anymore. The board definitions of the platform work just fine.


zackees:
This causes our RP2350 builds to fail.

Do I need to update the repo that we are pointing to?

```
    "rpipico2": [
        #"platform=https://github.com/earlephilhower/arduino-pico.git",
        "platform=https://github.com/maxgerhardt/platform-raspberrypi.git",
        "platform_packages=framework-arduinopico@https://github.com/earlephilhower/arduino-pico.git",
        "framework=arduino",
        "board_build.core=earlephilhower",
        "board_build.filesystem_size=0.5m"
    ],
```

maxgerhardt:
..Yeah. It refuses to do a `pio init -b rpipico2` because the custom `maxgerhardt/platform-raspberrypi.git` is not yet installed (`pio pkg install -g -p [..]`) which provides that board. I see why you had the custom board definition to circumvent that. But actually, let me restructure that to just trigger an install.

maxgerhardt:
Interesting. I've also added the "RP2040 but with Arduino-Pico instead of ArduinoCore-mbed" and it's discovered a.. compiler bug? [job logs](https://github.com/maxgerhardt/FastLED/actions/runs/10695755946/job/29649702820).
```
/tmp/cczflU3e.s: Assembler messages:
/tmp/cczflU3e.s:875: Error: invalid offset, value too big (0x00000530)
/tmp/cczflU3e.s:895: Error: invalid offset, value too big (0x000004EC)
*** [.pio/build/rpipico/src/Blink.ino.cpp.o] Error 1
```
I need to look into this some more.

zackees:
I've disabled FASTLED_RP2040_CLOCKLESS_M0_FALLBACK and this might have fixed your issues. We had a lot of assembly code. The comment says it's default disabled but the code was for default enabled. https://github.com/FastLED/FastLED/commit/fe5b54f165cfef63cc12f3631c13ea077361445a

zackees:
So still failing.

Here are some notes on your build:

You are using a platform arg which doesn't exist in the build_template_custom_board. Everything is overridden in the args. However, there is also the custom board options in ci/ci-compile.py file. This is done so that developers can run the ci/ci-compile.py tool and get a pretty close approximation to the github builders.

## Recommendations

* If you want to have custom options my recommendation is that you place the options in the ci/ci-compile.py in the CUSTOM_PROJECT_OPTIONS section.
* revert your changes for the platform arg and instead place it in args section with something like `--project-option=platform=https://github.com/maxgerhardt/platform-raspberrypi.git`

maxgerhardt:
>You are using a platform arg 

I've explicitly added that one. So that you  don't need the entire custom boards folder anymore, because a platform install **first** will enable regular project creation with the board ID.

grafik

I'll still refactor it to actually get *rid* of the platform URLs in the ci_compile.py then.

maxgerhardt:
..yeah, giving up on cleaning up the CI build to remove the `ci/boards` hack. It really shouldn't be done in the first place, you just need to `pio pkg install -g -p <platform>` the custom platform the board ID comes from first. The `ci-compile.py` is uncecessarily complicated, supporting a "comma separated list of boards" when every `.yml` file only ever uses a single board. I'll copy the `rpipico.json` file here and let it be.

zackees:
It's complicated because it's fast and concurrent.

Github runners have massive bandwidth so installing all the platforms at the same time makes a lot of sense.

I've also done a trick to make every example not have to re-compile the entire project. This brings down compilation down to a small fraction of what it would normally take. All these tricks summed together shave off 2/3rds of the build time.

The fact that it is used for only one board at a time was a recent development and it just works.

That's why it's complicated. Is it too complicated for a single board compilation? Yes it is.

I'll investigate using your `pio pkg install -g -p <platform>` and see if that resolves the issue. Not that this is a major issue. I'd like to have the builders support a board even before anyone else can compile for it so that when it arrives on the market in a big way we already support for it. Getting devs to work on FastLED when they are in the middle of a board bringup is critical. Once devs move onto other boards then volunteer efforts dwindle.

zackees:
I've refactored the ci-compile.py program and now it's much easier to make changes.

I'm now ready to do the pio pkg install feature.

Here is my package definition for rpico2:

```python
RPI_PICO2 = Project(
    board_name="rpipico2",
    platform="https://github.com/maxgerhardt/platform-raspberrypi.git",
    platform_packages="framework-arduinopico@https://github.com/earlephilhower/arduino-pico.git",
    framework="arduino",
    board_build_core="earlephilhower",
    board_build_filesystem_size="0.5m",
)
```

Given this set of attributes, can you give me just one example of how to use `pio pkg install` to install the necessary board?

maxgerhardt:
Sure. `pio pkg -g -p "https://github.com/maxgerhardt/platform-raspberrypi.git"`. 

[Docs](https://docs.platformio.org/en/latest/core/userguide/pkg/cmd_install.html).

zackees:
This isn't working for me:

```bash
# First command works
pio pkg install  -g -p https://github.com/maxgerhardt/platform-raspberrypi.git

# Second one does not.
pio project init --project-dir .build/rpipico2 --board rpipico2 --project-option=platform=https://github.com/maxgerhardt/platform-raspberrypi.git --project-option=platform_packages=framework-arduinopico@https://github.com/earlephilhower/arduino-pico.git --project-option=framework=arduino --project-option=board_build.core=earlephilhower --project-ackages=framework-arduinoption=board_build.filesystem_size=0.5m
```

Output:
```
Resolving rpipico2 dependencies...
UnknownBoard: Unknown board ID 'rpipico2'
```

Am I doing this wrong?

zackees:
looks like i have a typo and some build rules not being generated correctly

zackees:
Okay everything is working now and I've removed the custom board injection hack.

PR #1701, opened 2024-09-04
Add Dummy controller.

zackees:
In Gnu C, use weak attribute for XY.

This allows the code to link even if XY isn't filled in by the application, as long as the calling method is not used. The syntax is GNU-specific.

Add Unix platform.

Add PNG creation and WebSocket server code.

Add WEB chipset, based on PNG Websocket server.

Improve instructions for websocket mode


zackees:
This file needs to be re-written. The code locations have shifted.

zackees:
https://github.com/kw217

I tried to do a rebase off of master but it didn't fast forward. It needs manual merging, but it'd like 2 or 3 files.

PR #1702, opened 2024-09-04
Remove unneccessary CI files for RP2040/RP2350

zackees:
Restructure build without custom board files, install platform properly

Explicitly state RP2040 Earle Philhower / Arduino-Pico core


zackees:
I've squashed the commits and rebase https://github.com/maxgerhardt

PR #1703, opened 2024-09-04
Add RP2040 Arduino-Pico build

maxgerhardt:
Instead of just compiling for `nanorp2040connect` which by-default uses the ArduinoCore-mbed, explicitly adds `rpipico` to the build as Rasberry Pi Pico (Arduino-Pico core by Earle Phihower).


maxgerhardt:
Squashed.

PR #1726, opened 2024-09-23
fill_rainbow_circular(CHSV now produces hues from 0x00 to 0xbf rather…

un-clouded:
… than 0x00 to 0xff.

`fill_rainbow_circular(CHSV` seems to produce values for `hue` from `0x00` to `0xff`.  With a non-zero `initialhue` this results in a discontinuity when the HSV colors are converted to RGB with `hsv2rgb_raw_C`.  This patch results in values for `hue` only between `0x00` and `0xbf` (`HSV_SECTION_3` per section of the hue wheel).

### DIFF
--- a/.github/workflows/build_template.yml
+++ b/.github/workflows/build_template.yml
@@ -7,13 +7,6 @@ on:
         type: string
         required: false
         default: ""
-  workflow_dispatch:
-    inputs:
-      args:
-        description: 'Board to build'
-        required: true
-        type: string
-
 
 jobs:
   build:
@@ -56,6 +49,11 @@ jobs:
       - name: CPP Check
         run: |
           uv run ci/ci-cppcheck.py
+
+      - name: Inspect Binary
+        continue-on-error: true
+        run: |
+          uv run ci/inspect_binary.py --first
 
       - name: Generate timestamp and random hex
         id: generate_id
--- a/.github/workflows/build_template_binary_size.yml
+++ b/.github/workflows/build_template_binary_size.yml
@@ -50,6 +50,10 @@ jobs:
               print(f'{board} size {size} is within the limit of ${{ inputs.max_size }}')
           "
 
+      - name: Inspect Binary
+        continue-on-error: true
+        run: |
+          uv run ci/inspect_binary.py --first --cwd ci/tmp/FastLED
 
       - name: Generate timestamp and random hex
         id: generate_id
--- a/.github/workflows/build_template_custom_board.yml
+++ b/.github/workflows/build_template_custom_board.yml
@@ -46,6 +46,10 @@ jobs:
           fi
         if: always()
 
+      - name: CPP Check
+        run: |
+          uv run ci/ci-cppcheck.py
+
       - name: Generate timestamp and random hex
         id: generate_id
         run: |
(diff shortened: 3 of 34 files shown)


zackees:
Can you please elaborate? I am not familiar with the specifics of HUE wheel <-> RGB conversion other than surface level specifics.

un-clouded:
> Can you please elaborate? I am not familiar with the specifics of HUE wheel <-> RGB conversion other than surface level specifics.

Sure.  From the looks of [hsv2rgb_raw_C](https://github.com/FastLED/FastLED/blob/18d38b7f55e341bf791e1e33d5ca15dbb868ff65/src/hsv2rgb.cpp#L41), and assuming that all other hsv2rgb functions behave the same way, `hue` should be between `0x00` and `0xbf`.  I'm not familiar with the FastLED codebase but it seem from the hsv2rgb implementation that each third of the wheel spans `HSV_SECTION_3` (defined as `0x40`), totalling `0xc0` and that `hue` values of `0xc0` or greater are not valid.  This tracks with the discontinuity that I saw when using `fill_rainbow_circular` with a non-zero `initialhue` parameter.  When I animated the `initialhue` parameter, the discontinuity was noticeable.  With the updated code, there is no discontinuity in the colors and the animation looks as it should.

zackees:
Thanks for your comments. It was my understanding that HUE should be in range of 0x0 ... 0xff.

It seems that your change will break that? Sorry for the stupid question. I just don't want to break anything and the author has passed unexpectedly. I don't want to have a regression here so I'm being overly cautious.

@samguyer @kriegsman do you have insght here?

un-clouded:
Cautious is good.  Full disclosure, I'm not FastLED as-is, just lifting those two functions and using them with light_ws2812, but it seems like those functions should work together to make a rainbow that wraps without a join.  Looking forward to feedback.

zackees:
CL approved.

Please rebase to master and update.

zackees:
I'm still interested in this PR but I was doing some refactoring. I'll keep it open for now and merge it in later.

sutaburosu:
> I'm still interested in this PR but I was doing some refactoring. I'll keep it open for now and merge it in later.

I think this change only benefits @un-clouded, who would probably be better served using the same HSV->RGB conversion that FastLED uses internally, i.e. `hsv2rgb_rainbow()` instead of `hsv2rgb_raw_C()`.

[This sketch demonstrates](https://wokwi.com/projects/410005911140776961) that FastLED has no problems here when using implicit HSV->RGB conversions. The problem this PR fixes only happens when using `hsv2rgb_raw_C()`, which can be shown by setting `USE_RAW_CONVERSION` to `true`.

zackees:
> > I'm still interested in this PR but I was doing some refactoring. I'll keep it open for now and merge it in later.
> 
> 
> 
> I think this change only benefits @un-clouded, who would probably be better served using the same HSV->RGB conversion that FastLED uses internally, i.e. `hsv2rgb_rainbow()` instead of `hsv2rgb_raw_C()`.
> 
> 
> 
> [This sketch demonstrates](https://wokwi.com/projects/410005911140776961) that FastLED has no problems here when using implicit HSV->RGB conversions. The problem this PR fixes only happens when using `hsv2rgb_raw_C()`, which can be shown by setting `USE_RAW_CONVERSION` to `true`.

Thanks. Closing PR.

PR #1747, opened 2024-10-04
default order GRB for WS211X controllers in chipset.h

isidromv:
Controllers for WS211X chipsets are defined twice in `chipdset.h`, within a `#ifdef ... #else ...` structure.
The default color order is different for both definitions.

Set to `GRB` for all of them.


zackees:
thank you for your contribution.

PR #1758, opened 2024-10-27
Use #pragma push_macro(ARDUINO) in arduinojson wrapper

### LINKED ISSUE #1757: Including FastLED.h breaks ARDUINO macro
```cpp
static_assert(ARDUINO == 10607, ""); // succeeds
#include <FastLED.h>
static_assert(ARDUINO == 10607, ""); // fails

void setup() {}
void loop() {}
```

```
Arduino/fastled-arduino-bug/fastled-arduino-bug.ino: At global scope:
Arduino/libraries/FastLED/src/third_party/arduinojson/json.h:24:17: error: '__RESTORE_ARDUINO' was not declared in this scope
 #define ARDUINO __RESTORE_ARDUINO
                 ^
Arduino/fastled-arduino-bug/fastled-arduino-bug.ino:3:15: note: in expansion of macro 'ARDUINO'
 static_assert(ARDUINO == 10607, "");
               ^~~~~~~
```
This breaks user code or other libraries that depend on the macro being available (e.g. the MIDIUSB library).

The problem is this line: https://github.com/FastLED/FastLED/blob/8082fdceccc0c2d15ea39262222f4d1dcb28c52b/src/third_party/arduinojson/json.h#L25

Side note: using identifiers that start with an underscore or that contain consecutive underscores is technically not allowed (https://en.cppreference.com/w/cpp/language/identifiers).

tttapa:
Fixes #1757 for GCC >=5, but an alternative solution needs to be found for older GCC (e.g. Arduino Due, which is still stuck on GCC 4.8).

### DIFF
--- a/src/third_party/arduinojson/json.h
+++ b/src/third_party/arduinojson/json.h
@@ -10,8 +10,9 @@
 #define ARDUINOJSON_ENABLE_ARDUINO_PRINT 0
 
 #if defined(ARDUINO)
-#define __RESTORE_ARDUINO ARDUINO
+#pragma push_macro("ARDUINO")
 #undef ARDUINO
+#define FASTLED_RESTORE_ARDUINO
 #endif
 
 #define FASTLED_JSON_GUARD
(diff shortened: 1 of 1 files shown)


zackees:
Oh wow! Thanks!

PR #1765, opened 2024-11-03
fix: don't use DMA with RMT on ESP32-C6

WereCatf:
Using DMA with RMT on the ESP32-C6 causes a crash. This commit adds the -C6 to the list of devices for which FastLED shouldn't be using DMA.

Signed-off-by: Nita Vesa <nita.vesa@outlook.com>


zackees:
Thanks, a fix was already merged in.

PR #1791, opened 2024-12-15
Add workflow dispatch to docs action

dmadison:
This makes it possible to generate the documentation without needing to push a new release, in case documentation is updated between releases, documentation software is updated, or if the documentation action fails for any reason.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -1,6 +1,7 @@
 name: docs
 
 on:
+  workflow_dispatch:
   release:
     types: released
 


zackees:
While I agree that this is useful, there is too much in flux for the 3.9.X for this to make sense. Otherwise users will see master's view of the code instead of what the current release shows. After 4.0 this would be appropriate, or have a separate set of docs that clearly show this is master branch version.

zackees:
Seems like it might be a good idea then

@kriegsman, @samguyer ?

PR #1793, opened 2024-12-15
Fix compile errors in examples

nomakewan:
This fixes several compile errors that have cropped up in example sketches.

The fixes to `FxEngine.ino` and `Overclock.ino` are due to the changes from commit be60207.

Note that the order of operations shift in `Fire2012WithPalette.ino` was made because the WASM compiler does not automatically insert function prototypes like the Arduino IDE does and so it was erroring out. Moving the custom function above the loop resolves this issue (and in fact this keeping of custom functions above the loop can be seen in FastLED's existing example sketches such as `SmartMatrix.ino`).

On the topic of the WASM compile test, `Esp32Rmt51.ino` should be excluded as its platform definition lacks `FASTLED_RMT5` and thus this is an invalid sketch for WASM. Additionally `SmartMatrix.ino` should be excluded as this sketch is only valid on Teensy and uses an external third-party library that the WASM check is not using as a dependency.

### DIFF
--- a/examples/Fire2012WithPalette/Fire2012WithPalette.ino
+++ b/examples/Fire2012WithPalette/Fire2012WithPalette.ino
@@ -16,79 +16,7 @@ bool gReverseDirection = false;
 
 CRGB leds[NUM_LEDS];
 
-// Fire2012 with programmable Color Palette
-//
-// This code is the same fire simulation as the original "Fire2012",
-// but each heat cell's temperature is translated to color through a FastLED
-// programmable color palette, instead of through the "HeatColor(...)" function.
-//
-// Four different static color palettes are provided here, plus one dynamic one.
-// 
-// The three static ones are: 
-//   1. the FastLED built-in HeatColors_p -- this is the default, and it looks
-//      pretty much exactly like the original Fire2012.
-//
-//  To use any of the other palettes below, just "uncomment" the corresponding code.
-//
-//   2. a gradient from black to red to yellow to white, which is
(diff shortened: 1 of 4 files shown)


zackees:
Thanks, much appreciated!

zackees:
> Moving the custom function above the loop resolves this issue (and in fact this keeping of custom functions above the loop can be seen in FastLED's existing example sketches such as `SmartMatrix.ino`).

This is already done and is pending release.

zackees:
> On the topic of the WASM compile test, `Esp32Rmt51.ino` should be excluded as its platform definition lacks `FASTLED_RMT5` and thus this is an invalid sketch for WASM

I agree this is no longer relevant and I am taking it out now. Thanks for the reminder.

Keep in mind that because of the way that RMT5 works, it can't be linked in with RMT4, or else a weak symbol get's defined to non null and the ESP will then force a crash on boot, and I didn't want to put everything in the header at the time because the header list is absolutely gigantic. Therefore the header was a build-define and not an include define. The former isn't easy to put in the ArduinoIDE, despite it being a standard feature of literally every other build system out there, even the new Arduino Web compiler. But not for the downloadable IDE.

The build-define was part of the github builder, which I'm retiring now, see below.


```
name: esp32rmt_51

on:
  push:
    branches:
      - master

jobs:
  build:
    uses: ./.github/workflows/build_template.yml
    with:
      args: esp32rmt_51 --examples Esp32Rmt51 --defines FASTLED_RMT5=1

```

PR #1799, opened 2024-12-17
Update fastpin_arm_d51.h added 8 more usable pins

slyalfa:
Added some missing pins that can be used on some Grandcentral clones/ or raw SAMD51 chips.


zackees:
If only more people were like you...

Approved!

zackees:
Don't worry about the espdev size check, that happened earlier today and is not your fault.

Issue #1821, opened 2024-12-30
scale16by8 silently relies on integer overflows?


Be-ing:
When [porting the pseudo-13-bit gamma correction algorithm to Rust](https://github.com/smart-leds-rs/apa102-spi-rs/pull/15), I noticed that [`scale16by8`](https://github.com/FastLED/FastLED/blob/d5aaf65be19782f3e52b8b0fe38778f14376a293/src/lib8tion/scale8.h#L476) silently relies on integer overflows to work. The values passed to that function in its tests overflow the multiplication of `i * (1 + ((uint16_t)scale))` before the bitshift is applied. A `uint32_t` value is needed to hold the value of that multiplication before the `>> 8`.

I don't understand how this works because unsigned integer overflow in C++ is specified to wrap, but using Rust's `wrapping_mul` function to replicate that behavior fails the tests. Does overflowing an unsigned integer smaller than `size_t` silently just work like a `size_t` or something...?

zackees:
oh geez. It's possible that earlier checks are preventing degenerate values from reaching the scale function. I do have unit tests, but yes, those tests run on a native device. It's totally possible it has degenerate values on 8 bit avrs

sutaburosu:
As I understand it, AVR is the only platform FastLED supports with a word size less than 32-bits and AVR has assembly variants of these functions so there should be no bug evident in FastLED. 

I guess rather than relying on implicit promotion to uint, the C implementation should be made explicitly 32-bits just like the scale16() variant.

PR #1823, opened 2024-12-30
scale16by8: explicitly use uint32_t for intermediate computations

Be-ing:
`i * (1 + scale)` would overflow before the `>> 8` if it was stored in a uint16_t. Some implicit type conversion must be going on to make this work, but it's unclear if that implicit type conversion would reliably work on all platforms. The `1` literal may be casting `1 + scale` to an `unsigned int`, which on most platforms is 32 bits, but the C++ standard allows it to be 16 bits.

Fixes https://github.com/FastLED/FastLED/issues/1821


zackees:
Follow up, can you make this work without upcasting to uint32_t?

zackees:
It would involve a test to see if the number would overflow.

zackees:
Gotcha. Good work!

PR #1824, opened 2024-12-30
scale16by8: cast `scale` to 16 bit rather than 32

Be-ing:
`scale` starts as 8 bit. To add 1 to it without overflowing only requires 16 bits. 32 bits are only needed when multiplying scale by the 16 bit value of `i`.


zackees:
yeah i saw this and fixed this.

PR #1836, opened 2025-01-08
Fix #1830

### LINKED ISSUE #1830: Cannot compile example of I2S driver for the ESP32-S3
I was already running original [LED driver from hpwit](https://github.com/hpwit/I2SClockLessLedDriveresp32s3). 
I tried the insert version directly into FastLED strictly by example, but I can't compile:

```.ino:61:undefined reference to `_ZN2fl17InternalI2SDriver6createEv'```

lmancini:
Include the main library header file in order to use the `CONFIG_IDF_TARGET_*` flags.

### DIFF
--- a/src/platforms/esp/32/clockless_i2s_esp32s3.cpp
+++ b/src/platforms/esp/32/clockless_i2s_esp32s3.cpp
@@ -1,14 +1,9 @@
-
-
-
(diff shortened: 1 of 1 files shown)


zackees:
thank you

PR #1840, opened 2025-01-11
esp8266 clockless only define MIN if not already defined

dbu:
MIN is also defined in fl/math_macros.h, avoid the compiler warning.

### DIFF
--- a/src/platforms/esp/8266/clockless_block_esp8266.h
+++ b/src/platforms/esp/8266/clockless_block_esp8266.h
@@ -11,7 +11,10 @@
 
 #define FIX_BITS(bits) (((bits & 0x0fL) << 12) | (bits & 0x30))
 
+#ifndef MIN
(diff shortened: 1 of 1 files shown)


src/platforms/esp/8266/clockless_block_esp8266.h
```diff
@@ -11,7 +11,10 @@
 
 #define FIX_BITS(bits) (((bits & 0x0fL) << 12) | (bits & 0x30))
 
+#ifndef MIN
 #define MIN(X,Y) (((X)<(Y)) ? (X):(Y))
+#endif
```
dbu, line 16: if we can be sure that fl/math_macros.h is always loaded before this file, we could just delete the MIN macro. but that would create a rather implicit dependency...

i am not too familiar with C best practices, so not sure what the right solution is. this certainly is less risky of BC breaks.

zackees:
thanks

PR #1841, opened 2025-01-12
add more tests for brightness_bitshifter functions

Be-ing:
using fixed data in addition to the existing tests using randomly generated data.

I didn't realize these functions had some tests already before writing these. I figure they're helpful to add anyway; they helped me understand how these functions work.


zackees:
Bit representation of the numbers?

I think the test should have something like CRGB(r,g,b) and see that it maps to CRGB + brightness.

zackees:
My mistake, you are right.

PR #1850, opened 2025-01-17
Fix #1101 - check for num_leds == 0 and skip output routine

### LINKED ISSUE #1101: Zero length clockless led strip causes ARM SAMD21 to tightloop when calling `.show()`
[This if statement in the implementation for d21 chips](https://github.com/FastLED/FastLED/blob/3a1d880fb7abe5308950358cc1493c5b9e97e3eb/platforms/arm/d21/clockless_arm_d21.h#L29) causes a tightloop if the number of leds for a given CLEDController is set to 0. 

While having a controller control 0 LEDs may seem a little obtuse it is useful in the case of doing run-time configuration. I have worked around the bug simply by short circuiting a call to `.show()` much earlier if the number of leds is 0. I am unsure if this fix would be an acceptable one to merge into mainline.

kbob:
Thanks to [@apozharski](https://github.com/apozharski) for doing the hard work of isolating the bug.  Since I ran into this, I went ahead and submitted this PR.

I put the zero length check at the lowest level possible so it would catch all code paths.

### DIFF
--- a/src/platforms/arm/d21/clockless_arm_d21.h
+++ b/src/platforms/arm/d21/clockless_arm_d21.h
@@ -40,6 +40,9 @@ public:
     // This method is made static to force making register Y available to use for data on AVR - if the method is non-static, then
     // gcc will use register Y for the this pointer.
     static uint32_t showRGBInternal(PixelController<RGB_ORDER> pixels) {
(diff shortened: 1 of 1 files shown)


zackees:
Thank you!

PR #1854, opened 2025-01-19
Fix #1853 - RGBWEmulated does not work with SAMD21 clockless

### LINKED ISSUE #1853: RGBWEmulated does not work on SAMD21 clockless
The RGBWEmulated example does not work on the SAMD21 clockless controller (and probably others).   It runs, now that #1101 is fixed, but there is no output.

I have traced the problem to the fact that RGBWEmu does not call the device controller's init() method.  It also does not call beginShowLeds/endShowLeds, which probably breaks other drivers.

kbob:
The issue was that the RGBWEmulated controller did not call the device controller's `init` method.  Also, it did an update by calling `showPixels`.  `showPixels` bypasses the device's `beginUpdateLeds` and `endUpdateLeds` methods.

Keep in mind that the device controller is automatically linked into the global list of controllers (private to `CLEDController`).  So when show occurs, the device controller's show methods (`beginShowLeds`, `showLedsInternal` and `endShowLeds`) are invoked.  Because of the creation order, they are called immediately after RGBWEmulated's show methods.

I've changed RGBWEmulated to call the device controller at a higher level.  Instead of the device owning no LEDs and using LEDs passed in via showPixels, RGGWEmulated gives the LEDs immediately after allocating them (usually on the first show), then calls CLEDController::showLeds (with full brightness).  showLeds handles calling beginShowLeds, showLedsInternal, and endShowLeds.

To keep the device controller from displaying the same LEDs again when its show methods are invoked, RGBWController disables the controller (`setEnable(false)`).

Advantages of doing it this way.

  - less coupling between RGBWEmulated and CLEDController
  - works with controllers that use `beginShowLeds` and `endShowLeds`
  - works with controllers that are ill-behaved with empty LED arrays
  - works with controllers that incorrectly reference the LED array from `showPixels`

Disadvantages.

 - relies on `CLEDController::showLeds` which is not in the most-used code path
 - adds a few dozen instructions to each show (I guess.  I didn't measure.)

CRGBW also calls the device controller's `init` method, which should be non-controversial, and it resets all the parameters that would let CPixelController modify the RGB values.  Again, I changed the code to use higher-level interfaces to reduce coupling.

Sorry for the wall of text.  I'm new here and erring on the side of overexplanation.


kbob:
Is there some documentation I should be updating with this?  I didn't find any docs that described what RGBWEmulated is supposed to do or how to use it, just the example.

Ditto tests.  Should I be adding or updating a test?

kbob:
Thank you!

PR #1859, opened 2025-01-22
Remove some warnings

Jueff:
false warning when compiling ESP platforms not of type ESP32C3

src\platforms\esp\32\clockless_i2s_esp32s3.cpp:5:2: warning: #warning "esp_memory_utils.h is not available, are you on esp-idf 4? The parallel clockless i2s driver will not be available" [-Wcpp]

src\platforms\esp\32\rmt_4\idf4_rmt_impl.cpp:66:0: warning: "FASTLED_ASSERT" redefined
 #define FASTLED_ASSERT(format, errcode, ...)


zackees:
There are different changes in this PR. I'm going to pull your changes into a branch and cherry pick the ones for warning.

The change for RMT4 is terrifying. Can you tell me more about it? RMT4 is a legacy library and it's been through war. Every single one of those ugly hairs you see on it, has been produced through blood and death. And now you are telling me some of those are unnecessary. Sus.

zackees:
Also, and I don't say this to everyone, but I will say this to you. You are actually competent enough to fix RMT5 to do what you want RMT4 to do. God knows, the RMT5 bindings for FastLED are waaaaaaay simpler than all the magic we had to apply to get RMT4 to work.

zackees:
Also, fyi, have you seen this?

https://github.com/FastLED/PlatformIO-Starter

Jueff:
> The change for RMT4 is terrifying. Can you tell me more about it? RMT4 is a legacy library and it's been through war. Every single one of those ugly hairs you see on it, has been produced through blood and death. And now you are telling me some of those are unnecessary. Sus.

May long-term test now show, that previous hangs are done with changed code. I had four boards running for more than 72h without any hang. Before the change I had ~1-5 hand per 24h (the behavior reported in #1438).
Tihs bug is open since year, and only a workaround exists.

When doing deep analysis of the existing code I added a lot of debug code. That showed me, that sometimes a RMT channel doesn't raise interrupts. This happens when startOnChannel is called while other RMT channels already prduce interrupts. The accoring bit in RMT.int_st.val was never set for least one at of the active channels. And therefore fillNext()/doneOnChannel was never called. That was a hint, that something fails when changing the int_ena/int_clr.val registers. 

The change has been, that I moved the code that re-assignes a waiting controller to a channel, is moved away from method doneOnChannel called in interrupt handler. I moved it to showPixel, where previously the code only waited for gTx to be signalled.

Also I protected the modification of the RMT register with a lock.
Only both changes showed the expected result, having no more blocks.

Yes, I know, every change may break backward compatibility, but as you write, they have always the change to use previous version.

So I think, I have hunted a very good hidden bug, recovered and killed it :-).

Jakub-Bielawski:
Hi, I have also encountered an issue with unjustified warnings. Do you plan to implement the changes proposed by @Jueff  regarding the warnings in the near future, and make a release?

zackees:
> Hi, I have also encountered an issue with unjustified warnings. Do you plan to implement the changes proposed by @Jueff  regarding the warnings in the near future, and make a release? 

Yes, almost certain in will be in wednesdays release.

Jueff:
@zackees As I share your concern about breaking backward compatibility with my changes of the RMT4 driver I will close the existing pull request and place a new one, where the code change is optional using a define.
As this problem seems to occur only with old board version having ESP_IDF_VERSION_MAJOR<4 this new define will only be set for the older platforms.

zackees:
we good. Merged. I didn't realize how much testing you put in.

Thanks for crushing this long standing bug. Very much appreciated.

And good luck with your project.

PR #1860, opened 2025-01-22
Fix WS2816 on ObjectFLED

kbob:
Fixes the regression that WS2816 does not work on ObjectFLED.

    ObjectFLED was failing because beginShowLeds/endShowLeds were
    being called on disabled controllers.  This code puts guards
    (controller.getEnabled()) around those calls.  It also reorders
    the calls to all controllers' beginShowLeds -> all controllers'
    showPixels -> all controllers' endShowLeds, so that ObjectFLED
    can write all pins simultaneously.

N.B., this is still not thoroughly tested.  I know of one code path that doesn't work (CFastLED::showColor), and I've only tested two platforms.  So you can take this fix or you can wait for a more complete one.  I am continuing to test and debug, so I should have something solid in a day or two.

### DIFF
--- a/.github/workflows/check_teensy41_size.yml
+++ b/.github/workflows/check_teensy41_size.yml
@@ -13,6 +13,6 @@ jobs:
     uses: ./.github/workflows/build_template_binary_size.yml
     with:
       board: teensy41
-      max_size: 80000
+      max_size: 104000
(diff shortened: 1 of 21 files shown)


kbob:
I will redo this with the conflicts resolved.

PR #1861, opened 2025-01-23
Fix WS2816 on ObjectFLED and ESP32S3

kbob:
These changes make WS2816 work with the Teensy4's ObjectFLED controller and ESP32S3's IDF5 RMT controller.  They also fix CFastLED::showColor.

FWIW, my ESP32 board is a Seeed XAIO ESP32S3.  That seems to be the default in `platformio.ini`, so I'm not the only one using it.  The examples all use pin 3, and when I run them, the LED signals appears on the pin labeled A2.  The parallel examples use pins 1 and 3, and those come out on A0 and A2.  This seems to be consistent with Arduino's pin assignments, but it contradicts the silkscreen and [Seeed's own documentation](https://wiki.seeedstudio.com/xiao_esp32s3_getting_started/).  It's also consistent across FastLED/PlatformIO and FastLED/Arduino, and not dependent on using WS2816, so I think it's right.  It's just disconcerting.

### DIFF
--- a/src/FastLED.cpp
+++ b/src/FastLED.cpp
@@ -173,16 +173,31 @@ void CFastLED::showColor(const struct CRGB & color, uint8_t scale) {
 	int length = 0;
 	CLEDController *pCur = CLEDController::head();
 	while(pCur && length < MAX_CLED_CONTROLLERS) {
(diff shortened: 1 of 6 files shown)


zackees:
lol you are awesome.

Issue #1856, opened 2025-01-21
ESP32 RMT4 driver blocks FastLED.show()


Jueff:
Having that FastLED controller configuration

CLEDController& controller0 = FastLED.addLeds<NEOPIXEL, 27>(leds+  0,230); 
CLEDController& controller1 = FastLED.addLeds<NEOPIXEL, 32>(leds+230,  1); 
CLEDController& controller2 = FastLED.addLeds<NEOPIXEL, 16>(leds+231,  3); 
CLEDController& controller3 = FastLED.addLeds<NEOPIXEL, 14>(leds+234,  32); 
CLEDController& controller4 = FastLED.addLeds<NEOPIXEL, 18>(leds+266,  2); 

FastLED.show() block after some seconds.

Platform: ESP32 Arduino
Board: 1.0.4
FastLED: 3.9.9

When removing controller4 no more blocks occur. 
When downgrading to 3.7.0 no more blocks occur.

The sample code below reproduces the problem.
It also runs an extra task on other core to "revives' FastLED by calling rmt drivers GiveGTX_sem().

```
#include <FastLED.h>

#define NUM_LEDS 268

// Define the array of leds
CRGB leds[NUM_LEDS];
TaskHandle_t  taskHnd;
uint32_t lastShow = 0;
uint32_t lastBlink = 0;
uint32_t lastLog = 0;
uint32_t cnt = 0;
uint32_t failCount = 0;
bool displayLedStartMsg = true;
bool displayCheckStartMsg = true;
bool ledIsOn;

extern void IRAM_ATTR GiveGTX_sem() __attribute__((weak));
void setup() { 
  Serial.begin(115200);

  CLEDController& controller0 = FastLED.addLeds<NEOPIXEL, 27>(leds+  0,230); 
  CLEDController& controller1 = FastLED.addLeds<NEOPIXEL, 32>(leds+230,  1); 
  CLEDController& controller2 = FastLED.addLeds<NEOPIXEL, 16>(leds+231,  3); 
  CLEDController& controller3 = FastLED.addLeds<NEOPIXEL, 14>(leds+234,  32); 

  // **********************************************************
  // when commenting out the controller4 FastLED.show() won't block
  // **********************************************************

  CLEDController& controller4 = FastLED.addLeds<NEOPIXEL, 18>(leds+266,  2); 
  
  // create a task that monitor the alive state of the ESP
  // commenting out the next line displayed the alive check, FastLED.show() will then block until reboot
[...]

PR #1866, opened 2025-01-24
ESP32 RMT4 driver blocks FastLED.show() 

Jueff:
Fix for #1856 and #1438

introduce a new define RMT4_USE_THREADSAFE_REGISTERS 
with the define a thread safe variant of the RMT4 register handling is used
1.) gTxSem is now triggered every time a channel has finished sending data. Adding a waiting controller to the freed channel is no longer done within the interrupt.
Now inside showPixel method a loop takes care about waiting for all controllers to be finished and/or assigning controllers to channel that got free.
2.) I protect access to RMT registers with a lock

### DIFF
--- a/src/platforms/esp/32/rmt_4/idf4_rmt_impl.cpp
+++ b/src/platforms/esp/32/rmt_4/idf4_rmt_impl.cpp
@@ -124,6 +124,20 @@ extern "C"
 #endif
 #endif
 
+// -- use a thread safe version for register RMT handling, also start channels only from main thread, not within ISR
+#ifdef RMT4_USE_THREADSAFE_REGISTERS
+#include "freertos/semphr.h"
+#define OPTIONAL_RMT4_REGISTER_LOCK_ENTER     portENTER_CRITICAL(&rmt_spinlock)
+#define OPTIONAL_RMT4_REGISTER_LOCK_EXIT      portEXIT_CRITICAL(&rmt_spinlock)
+#define OPTIONAL_RMT4_REGISTER_LOCK_ISR_ENTER portENTER_CRITICAL_ISR(&rmt_spinlock)
+#define OPTIONAL_RMT4_REGISTER_LOCK_ISR_EXIT  portEXIT_CRITICAL_ISR(&rmt_spinlock)
+#else
+#define OPTIONAL_RMT4_REGISTER_LOCK_ENTER
+#define OPTIONAL_RMT4_REGISTER_LOCK_EXIT
+#define OPTIONAL_RMT4_REGISTER_LOCK_ISR_ENTER
+#define OPTIONAL_RMT4_REGISTER_LOCK_ISR_EXIT
+#endif
+
 namespace {
     bool gUseBuiltInDriver = false;
 }
@@ -201,6 +215,9 @@ static int gNumStarted = 0;
 static int gNumDone = 0;
 static int gNext = 0;
 
+#ifdef RMT4_USE_THREADSAFE_REGISTERS
+static portMUX_TYPE rmt_spinlock = portMUX_INITIALIZER_UNLOCKED;
+#endif
 static intr_handle_t gRMT_intr_handle = NULL;
 
 // -- Global semaphore for the whole show process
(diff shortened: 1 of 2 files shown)


zackees:
Not necessary.

Let's keep everyone on the same code path. You convinced me with your extensive testing that this is the right way to go about it.

Cheers

zackees:
This is the I2S controller, not the RMT controller.

because of reasons, we put the cpp file into an hpp file so it is only compiled if the user explicitly flips on the define at include time. Most of our users are on ArduinoIDE and this build system does not allow build defines without hacking the IDE.

I didn't want to do this, but I had to.

Your issue is surprising to me. Can you explain what your setup is?

RococoN8R:
Woops, apologies, I hadn't had coffee yet. 
I see your issue and need to find a work-around. 
For context, I'm a hobbyist and have gaps in my understanding.
I'm creating a platform leveraging HTTP, MQTT, WebSockets and a external signals.
I have a wrapper around much of FastLED managing sequences and animation states.
Given the complexity, I have a separate implementation file for user-created scripts.
It is part of the wrapper class, but FastLED is imported to support the user code and IntelliSense.
I anticipate needing access to other FastLED elements as other event-driven capabilities are added, so I'm not sure the issue goes away for me.
Just spit-balling given my lack of understanding, but could a flag default to single HPP file for the Arduino IDE environment and a build flag enable use of a split file model? I recognize it would create some redundant code, my existing installs involve 5 or more pins, and RMT just isn't doing it.

Keep up the great work!

zackees:
I don't think that's possible, but I may go back to doing a CPP file for the ESP i2s stuff.

You are trying to include FastLED in multiple cpp files yes?

You should know that if you are trying to hide implementation that I've put a lot of work into making this easy. While one cpp file will need to include FastLED.h, the rest should only have to rely on crgb.h in order to get the CRGB struct/class so you can do your pixel writing.

The convention I use is that gfx.cpp in my sketch includes the FastLED.h header, everywhere else includes crgb.h because typically pixel painting is done in multiple visualizer cpp's.

Before the 3.9.X series, crgb.h included FastLED.h. Now it does not and can be included anywhere you want.

Good luck.

RococoN8R:
Yes, I'm using FastLED.h in two .cpp files.I have a few years of custom effects and sequences built around the legacy architecture, and while I'm interested in adopting your 4.0 vision, I can't begin that exploration for 6 or more months. Unfortunately I can't tell you how many other developers will encounter similar constraints. Let me know your plans on the cpp file so I can determine how to best proceed.Sent from my phone... please excuse the topys.On Jan 25, 2025 12:06 PM, Zachary Vorhies ***@***.***> wrote:
I don't think that's possible, but I may go back to doing a CPP file for the ESP i2s stuff.
You are trying to include FastLED in multiple cpp files yes?
You should know that if you are trying to hide implementation that I've put a lot of work into making this easy. While one cpp file will need to include FastLED.h, the rest should only have to rely on crgb.h in order to get the CRGB struct/class so you can do your pixel writing.
The convention I use is that gfx.cpp in my sketch includes the FastLED.h header, everywhere else includes crgb.h because typically pixel painting is done in multiple visualizer cpp's.
Before the 3.9.X series, crgb.h included FastLED.h. Now it does not and can be included anywhere you want.
Good luck.

zackees:
> Yes, I'm using FastLED.h in two .cpp files.I have a few years of custom effects and sequences built around the legacy architecture, and while I'm interested in adopting your 4.0 vision, I can't begin that exploration for 6 or more months. Unfortunately I can't tell you how many other developers will encounter similar constraints. Let me know your plans on the cpp file so I can determine how to best proceed.Sent from my phone... please excuse the topys.On Jan 25, 2025 12:06 PM, Zachary Vorhies ***@***.***> wrote: I don't think that's possible, but I may go back to doing a CPP file for the ESP i2s stuff. You are trying to include FastLED in multiple cpp files yes? You should know that if you are trying to hide implementation that I've put a lot of work into making this easy. While one cpp file will need to include FastLED.h, the rest should only have to rely on crgb.h in order to get the CRGB struct/class so you can do your pixel writing. The convention I use is that gfx.cpp in my sketch includes the FastLED.h header, everywhere else includes crgb.h because typically pixel painting is done in multiple visualizer cpp's. Before the 3.9.X series, crgb.h included FastLED.h. Now it does not and can be included anywhere you want. Good luck. —Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you commented.Message ID: ***@***.***>

Moved to:

https://github.com/FastLED/FastLED/issues/1875

PR #1870, opened 2025-01-25
Fix WS2816 to bypass `protected` scope in various platform controller…

kbob:
… classes

Added a helper class to WS2816Controller that subclasses the platform's controller and allows access to the controller's protected methods. Ugly, I'm sorry.


zackees:
I'm going to accept this. I'm not a big fan of too much protected functions anyway. I'm coming to the conclusion that we can open up everything to the world because we are all adults.

PR #1871, opened 2025-01-25
Make RGBWEmulated work with ObjectFLED, ESP3D RMT5, and other controllers

kbob:
Working on the WS2816 support, I learned that some controllers care about the order in which their beginShowLeds, showPixels, and endShowLeds methods are called.  RGBWEmulatedController was not calling them in the correct order.  So I side-ported some code from WS2816 to fix RGBWEmu.

(Yes, those two controllers already have RGBW support.  But others probably have the same order dependency. ESP32 and Teensy4 are the test boards I have available.)


zackees:
Thank you very much!

PR #1874, opened 2025-01-27
Fix ATtiny4313/13 builds

sutaburosu:
`LIB8_ATTINY` was not being defined, which led to the invalid opcode errors.


zackees:
Attiny13 will never work with FastLED.

It only has 64 bytes of instruction memory.

PR #1879, opened 2025-01-30
Enable test compiles for ESP32 with ESP-IDF 3.3 LTS

Jueff:
add a test build task to avoid future ESP32 compile problems for projects based on  ESP-IDF 3.3 LTS

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -85,3 +85,8 @@ jobs:
     uses: ./.github/workflows/build_template.yml
     with:
       args: "esp32dev_idf44"
+
+  build_esp32dev_idf33:
+    uses: ./.github/workflows/build_template.yml
+    with:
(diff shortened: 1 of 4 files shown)


zackees:
This should be one commit, can you please `git squash` these into one?

Jueff:
thank you for the tip, I was wondering how combining commits can be done....

zackees:
Ah you've never done it before.

Okay, the first time absolutely sucks but once you figure it out and write a script for its solved for ever.

I can go ahead and git squash and look at the diff for you then approve this cl or not.

But I do suggest you learn how to git squash eventually. It's one of those advanced techniques that really makes a statement of professionalism when you do a PR and helps a PR get accepted very quickly. A bunch of commits has a code smell and is a quick indicator that the person is at a lower skill level.

Your PRs are so high quality I just assumed you knew how to do this already.

PR #1889, opened 2025-02-15
Remove unused variable; ret

Johboh:
- Fixing warning:
```
 warning: variable 'ret' set but not used [-Wunused-but-set-variable]
   85 |     esp_err_t ret = ESP_OK;
      |               ^~~
```
The function returns the error code from `rmt_new_led_strip_encoder_with_timings`, or a defined error.


zackees:
This is used in the macros and this PR causes build failures.

I've gone ahead and issued a fix for this false positive error using FASTLED_UNUSED(....)

PR #1899, opened 2025-03-07
Additional platform identifier define added for giga r1

RubiCubix:
Additional platform identifier define ARDUINO_GIGA added. Fixes issues with non recognizable platform when building with Arduino IDE


zackees:
Thanks!

PR #1906, opened 2025-03-19
fixing ESP IDF V 5.4 compiler error [-Werror=mismatched-new-delete]

rommo911:
compiling the library using latest EDP IDF V5.4 produce compiler warning as error [-Werror=mismatched-new-delete] 
this is due to allocating  using 
new uint8_t[width * height]  ( using [] ) 
while deleter is not using [] operators  in scoped_ptr  
fix : using scoped_array instead


zackees:
Thank you very much! This is great and very much appreciated!

rommo911:
you are welcome !

Issue #1909, opened 2025-03-27
FastLED incompatible with ArduinoSTL on AVR due to `operator new(size_t, void* ptr) _NO_EXCEPT`


Dantali0n:
Due to the definition of `new` in `inplacenew.h`:

```c
#if defined(__AVR__) || !defined(__has_include)
#ifndef __has_include
#define _NO_EXCEPT
#else
#define _NO_EXCEPT noexcept
#endif
inline void* operator new(size_t, void* ptr) _NO_EXCEPT {
    return ptr;
}
```

ArduinoSTL can no longer be used together with FastLED. I would propose to add an additional preprocessor check that allows to exclude this inline definition.

```c
#if (defined(__AVR__) || !defined(__has_include)) && !defined(FASTLED_NO_NEW)
```

Dantali0n:
Seems github didn't connect the issue to the pull automatically:
https://github.com/FastLED/FastLED/pull/1910

PR #1910, opened 2025-03-27
1909: Add preproccesor check to prevent redefinition of new()

Dantali0n:
Closes: #1909

This adds a new define check using `FASTLED_HAS_NEW` to prevent redefinition of `operator new()` on AVR as that is also defined by the `ArduinoSTL` library.


zackees:
Thank you!

PR #1924, opened 2025-04-23
Updates component dependencies in CMake configuration for esp-idf v5.4 ESP32(S3) needed dependency 

rommo911:
added 'esp_lcd' in the CMake since its needed and not included by default in ESP32-S3 compile using IDF v5.4


zackees:
Thanks!

PR #1927, opened 2025-05-18
fix: use dma if enabled

Joe91:
Didn't dig much deeper, but it seems, that this makes at least for my setup a difference (ESP32-WROVER-E).
Sadly the Clockless SPI doesn't work on my ESP32 version. Was not able to find out why this is not possible on this board, but will dig deeper on the exact reason.

Thanks a lot for your great work!


zackees:
thanks! Looks like default is still off. So this is great. Looks like an oversight on my part.

PR #1931, opened 2025-05-26
Fix: Avoid FL_ALIGN_PROGMEM redefinition warning

lsellens:
Adds #ifndef FL_ALIGN_PROGMEM to prevent redefinition when platforms/null_progmem.h already defines it. Fixes a warning seen on platforms like RP2040 with FastLED 3.9.17.


lsellens:
I pushed this to wrong branch. I'll close and fix it.

PR #1939, opened 2025-06-09
FireMatrix Fix

4wheeljive:
Proposed fixes to what I believe may be swapped uses of HEIGHT and WIDTH in several parts of the sketch (which canceled each other out and would not be noticeable with a square matrix as in this example).


zackees:
Boom, this is great! Thank you!

PR #1945, opened 2025-06-24
Update animartrix_detail.hpp

4wheeljive:
Fix polar_waves


4wheeljive:
I realized there were 10 more instances of what I believe are swapped x and y references. So I'm closing out this and submitting a new PR that includes all 11 suggested fixes.

PR #1948, opened 2025-06-24
Fix ESP32C2 build action

nomakewan:
This should fix the ESP32C2 build action by updating the `platformio.ini` file with an updated SDK definition. This definition is required to avoid the `**** missing SConscript file ‘/home/cw/.platformio/packages/framework-arduinoespressif32-libs/esp32c2/pioarduino-build.py’*` error, per [this discussion](https://community.platformio.org/t/use-of-arduino-esp32-in-platformio/45382/22).

It also fixes some improper indentation in `build_wasm_compilers.yml` which was causing errors when the action tried to start.

### DIFF
--- a/.github/workflows/build_wasm_compilers.yml
+++ b/.github/workflows/build_wasm_compilers.yml
@@ -4,7 +4,7 @@ on:
     push:
       branches:
         - master
-    paths:
+      paths:
       - 'src/**'
       - 'examples/wasm/**'
       - 'examples/WasmScreenCoords/**'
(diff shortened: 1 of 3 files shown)


nomakewan:
It would appear that the workflow that checks against pull requests is not working properly:

badcheckout

As you can see in the "Checking out the ref" section, for some reason it is trying to check out `refs/remotes/origin/master` instead of `refs/remotes/pull/1948/merge`. So the changes in this PR are not being reflected in the CIs because the CIs are just checking out `master` from `FastLED/FastLED` instead.

zackees:
Thanks for the tip on the builders, I put in a fix.

PR #2013, opened 2025-08-04
Code generation to support debugging on AVRs 

felias-fogg:
Not that the new code (using the constraint `[b1] "=&a" (b1)` instead of 	`[b1] "+a" (b1)`) is active if the code is compiled without -Os, which is the standard optimization level for AVR code. The new code generation will also emit a warning.


zackees:
You've got a claude.md change.

Can you revert that file?

PR #2014, opened 2025-08-04
Changed code generation to support AVR debugging

felias-fogg:
Again ...

### DIFF
--- a/FEATURE.md
+++ b/FEATURE.md
@@ -1,0 +1,835 @@
+# FEATURE: Efficient Multi-Platform Build System Using pio run
+
+## Current Problem
+
+FastLED currently uses `pio ci` for multi-platform compilation testing, which is extremely inefficient:
+
+- **Each `pio ci` call creates a new project from scratch**
+- **Downloads dependencies repeatedly** for each example/board combination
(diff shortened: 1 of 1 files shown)


zackees:
Okay so it looks like you've tested this in release and debud mode and you've found that it works in both ways. How thoroughly did you test this? The reason I ask is that there's other avr platforms that likely share this code path.

Right now I see that you are using an if-def conditional. The pattern I like to use is two phase macros:

```C++

#ifndef FASTLED_AVR_USES_OPTIMIZED_TIMING
#ifdef __OPTIMIZE_SIZE__
#define FASTLED_AVR_USES_OPTIMIZED_TIMING 1
#else
#define FASTLED_AVR_USES_OPTIMIZED_TIMING 0
#endif
#endif

#if FASTLED_AVR_USES_OPTIMIZED_TIMING
...
#else
...
#endif
```

That way if something goes wrong for one of the avr chipsets you didn't realize needed the old behavior, the users have an escape hatch to manually set the right code path.

Obviously I could do this change myself, but you probably want the contributor badge from a merged cl so I'll let you do the correction.

PR #2089, opened 2025-09-24
Fix Audio advanced.h drawSpectrumBars()

4wheeljive:
If matrix WIDTH < (2 x NUM_BINS) , then `int barWidth =  WIDTH / NUM_BANDS` = 1 (or even 0). 
So `barWidth - 1` = 0,
So the loop at line 198 ` for (int x = 0; x < barWidth - 1; x++)`  never executes.

### DIFF
--- a/examples/Audio/advanced/advanced.h
+++ b/examples/Audio/advanced/advanced.h
@@ -195,7 +195,7 @@ void drawSpectrumBars(FFTBins* fft, float /* peak */) {
         int barHeight = magnitude * HEIGHT;
         int xStart = band * barWidth;
         
-        for (int x = 0; x < barWidth - 1; x++) {
+        for (int x = 0; x < MAX(barWidth, 1); x++) {
             for (int y = 0; y < barHeight; y++) {
                 uint8_t colorIndex = fl::map_range<float, uint8_t>(
                     float(y) / HEIGHT, 0, 1, 0, 255


zackees:
thank you.

zackees:
It's a little rough in this part of the codebase. Any changes you propose for fixups is greatly appreciated.

4wheeljive:
You're welcome. I've been plugging away trying to figure out how to use various parts of the audio functionality library you're putting together, which I know is very much a work in progress. (I've read https://github.com/FastLED/FastLED/blob/master/FEATURE.md, which I believe was behind much of the new audio_reactive stuff.)

What I've done so far in my test/prototype project (ESP32-S3 with INMP441) is:
1) Begin more or less with the **Audio Advanced** example
2) Figure out how to insert portions of the **AudioInput** example (and the audio_input library) to use I2S for audio capture
3) Add in hooks to the **AudioReactive** library

What I've started most recently is refactoring each of the Audio Advanced visualizers to pull data from AudioReactive::AudioData objects instead of audio.h-based AudioSample objects.

I've looked quite a bit at https://github.com/FastLED/FastLED/blob/master/tests/test_audio_reactive_simple.cpp to help me discover what some of the AudioReactive capabilities are and take a stab at using them. And I've gotten a lot of help from Claude. But it's been somewhat slow going.

One thing that would be helpful is to know a bit more about the status of the AudioReactive. Are you planning to spend much/any time on this over the next few weeks/months? 

One thing in particular...

I see that the AudioReactive::AudioReactiveConfig struct includes a `bool noiseGate `element; but as far as I can tell, this is purely a placeholder at the moment. In my project, I would like to employ an early-in-the-chain noise gate/filter/floor of some sort, and I'm wondering: **Do you have any percolating thoughts about how you might incorporate noise filtering into AudioReactive?** If so, I'd love to learn about that. Otherwise, I'll just keep working on rolling my own solution that I will place between `currentSample = audioSource->read()` and `audio.processSample(currentSample)`

I will be more than happy to share what I've been working on. 

Thanks,
Jeff

4wheeljive:
https://youtu.be/SM-4ZamYaYM
FYI, here's a video I shot last night of my test project in action. This is using the I2S input, going through FL Audio Reactive processing, and into a refactored Radial Spectrum visualizer from the Audio Advanced example. A good place to jump in for a quick view might be around 1:45.

zackees:
> https://youtu.be/SM-4ZamYaYM FYI, here's a video I shot last night of my test project in action. This is using the I2S input, going through FL Audio Reactive processing, and into a refactored Radial Spectrum visualizer from the Audio Advanced example. A good place to jump in for a quick view might be around 1:45.

This is UNBELIEVABLE awesome to see. Good work!!

You may notice that the Animartrix examples have a global time value that you can adjust in the wasm demos. It either feeds into the FxEngine's time value or it modifies Animartrix's time value directory, i think it's then former.

If you connect these two concepts, then you already know how to make animartrix audio reactive.

zackees:
This is an old video of the FFT when it was in a broken state (it was missing audio frames and you'll see this in the missed percussion hits, but this has been subsquently fixed).

But if you don't mind looking at an old broken impl of FFT, this might inspire you:

https://www.youtube.com/watch?v=_6Msl1PyyRU

4wheeljive:
Interesting about the Animartrix global speed setting. I was never sure exactly what that did or how it related to other speed factors in the sketch; so my current implementation simply has a fixed `animartrixEngine.setSpeed(1)`.  I'll definitely need to explore this further (especially as timing becomes more important with audio elements)!

As far as where to put the db conversion tuning for INMP411, I need to learn more about how some of the back end stuff is currently working. I believe what you're asking about is the inmp441_rms_to_dB() function in platforms/esp/32/audio/sound_util.h and sound_util.cpp. Is that correct? I may circle back with some specific questions on this.

Thx for sharing your FFT video. Very cool! And indeed inspiring. We definitely need some good audio-enabled scrolling visualizers like that!

zackees:
> I believe what you're asking about is the inmp441_rms_to_dB() function in platforms/esp/32/audio/sound_util.h and sound_util.cpp. Is that correct? I may circle back with some specific questions on this.

Correct.

> Thx for sharing your FFT video. Very cool! And indeed inspiring. We definitely need some good audio-enabled scrolling visualizers like that!

Check out `examples/Audio/simple/**` this is where the FFT demo lives.

4wheeljive:
FYI, I am giving myself a deep-dive crash course on audio and digital signal processing basics (things like PCM, RMS, FFT, spectral density, etc.) Figured I really need to get my head around some of this stuff in order to make much progress here. Many things are starting to make a lot more sense!

zackees:
I've already figured it out. It's the concept `Sound2Midi`. AI is pounding away right now to make this amazing.

4wheeljive:
Holy shit! This is game-changing!

zackees:
Oh... yeah. It's coming back. The previous api was just an experiment to figure out where i needed to go. It turns out that sound2midi doesn't have a good general algorithm for typically music.

It needs to have a collection of highly specialized algorithms. Vocals, beat detection, down beat, spectrum, spectral flux, attack, plus sound to midi for simple polyphonic types.

I'll come back to it once i get the builds straightened out and the Software SPI 32 and Octo HW spi released.

Issue #1885, opened 2025-02-05
HD108 support


Blaster1920:
Since the last release brought the HD107 to fastled, I've tried my best to write some code to port the HD108 as well, but I'm hitting a wall. I can see The pins are sending out data but it's not quite right yet, as the leds won't get set.
This is the barebone code that should make implementing the HD108 a little bit easier:

```
template <
    uint8_t DATA_PIN,
    uint8_t CLOCK_PIN,
    EOrder RGB_ORDER = RGB,
    uint32_t SPI_SPEED = DATA_RATE_MHZ(40)
>
class HD108Controller : public CPixelLEDController<RGB_ORDER> {  
    typedef SPIOutput<DATA_PIN, CLOCK_PIN, SPI_SPEED> SPI;
    SPI mSPI;

public:
    HD108Controller() {}

    virtual void init() override {
        mSPI.init();
    }

protected:
    virtual void showPixels(PixelController<RGB_ORDER> & pixels) override {
        mSPI.select();
        startFrame();
        while (pixels.has(1)) {
            uint8_t rGain = 1, gGain = 1, bGain = 1;  // Default gains to avoid LED flickering
            uint8_t r8, g8, b8;
            uint16_t r, g, b;

            // Corrected: Use FastLED's standard scaling function
            pixels.loadAndScaleRGB(&r8, &g8, &b8);

            // Scale from 8-bit (0-255) to 16-bit (0-65535) for HD108
            r = (uint16_t(r8) << 8) | r8;
            g = (uint16_t(g8) << 8) | g8;
            b = (uint16_t(b8) << 8) | b8;

            // Send data in the required format
            writeLedFrame(rGain, gGain, bGain, r, g, b);
            pixels.stepDithering();
            pixels.advanceData();
        }
        endFrame(pixels.size());
        mSPI.waitFully();
        mSPI.release();
    }

private:
    inline void startFrame() {
        mSPI.writeByte(0x00);
        mSPI.writeByte(0x00);
        mSPI.writeByte(0x00);
        mSPI.writeByte(0x00);
    }

    inline void endFrame(int nLeds) {
        int nBytes = (nLeds / 16) * 8;
        for (int i = 0; i < nBytes; i++) {
            mSPI.writeByte(0xFF);
        }
    }

inline void writeLedFrame(uint8_t rGain, uint8_t gGain, uint8_t bGain, uint16_t r, uint16_t g, uint16_t b) {
    uint8_t frame[8] = {
[...]

Issue #1637, opened 2024-06-22
How do I make usage of the Hardware SPI for my ESP32?


PelksUnderscore:
This is my first time asking a question and im not even sure if im asking it correctly in the right place of this project. 
In my Arduino IDE program whenever I start it up I get the message "No hardware SPI pins defined. All SPI access will default to bitbanged output" and after a bit of research I stumbled upon this post https://github.com/FastLED/FastLED/pull/1047 and did what it said and now my question is did I implement it correctly on my ESP32 when the code and message look like this?
Code:
#define FASTLED_ALL_PINS_HARDWARE_SPI
#define FASTLED_ESP32_SPI_BUS HSPI
#include <FastLED.h>

#define HSPI_MOSI_DATA 13
#define HSPI_SCLK_CLK 14
#define COLOR_ORDER BGR
#define LED_TYPE SK9822
#define Brightness 30
#define NUM_LEDS 41

  FastLED.addLeds<LED_TYPE,HSPI_MOSI_DATA,HSPI_SCLK_CLK,BGR>(leds,NUM_LEDS);
  FastLED.setBrightness(Brightness);

Message:
In file included from c:\Users\XXX\Documents\Arduino\libraries\FastLED\src/platforms/esp/32/fastled_esp32.h:6,
                 from c:\Users\XXX\Documents\Arduino\libraries\FastLED\src/platforms.h:42,
                 from c:\Users\XXX\Documents\Arduino\libraries\FastLED\src/FastLED.h:62,
                 from F:\POVI180IAPPIDUALIV6\POVI180IAPPIDUALIV6.ino:3:
c:\Users\XXX\Documents\Arduino\libraries\FastLED\src/platforms/esp/32/fastspi_esp32.h:2:17: note: #pragma message: ESP32 Hardware SPI support added
 #pragma message "ESP32 Hardware SPI support added"
                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

zackees:
With these chips it's pretty much all hardware spi with their fancy async dma controllers.

The hardware spi optimization is more of a relic of 10 years ago when the first arduino was launched and software spi was much slower than realtime hardware spi.

Unless you are running the original arduino then don't worry about it.

PR #2119, opened 2025-10-29
hd108: Initial working state

arfoll:
I'm throwing this quickly here for feedback and maybe it's even useful for someone who accidentally bought such an LED strip (didn't expect these strips where so exotic!).

It's a really basic implementation that simply scales the 8bit values to 16. It looks pretty good to me but I have very little reference. Maybe HD108 should rather be HD108_SD.... If people are interested I can make a quick video showing how this looks.

Tested on:
- Teensy 4.1 using 30mhz SPI (40 should work but my setup is far from ideal noise wise, so it's a little flickery at above 30, I will redo some tests with better wiring later)
- Also on an Uno R4, obviously there it can't keep up with the 20mhz SPI but it does what it can :) (was just easier to debug with my logic analyser)
- a 43 LED long strip (future me shall use a smaller test strip...)
- This is the strip I used, https://www.superlightingled.com/hd108-ic-individually-addressable-led-rgb-full-color-changing-digital-flexible-dc5v-led-strip-lights-p-4522.html

A few disclaimers:
- it's the first RGB LED strip I use so I have no reference point, I also cheated to try get the blues a little brighter, not sure if this is the right place for it
- I've never used fastled before, and I will run the linters etc, just didn't get round to it on this winpc and figured feedback is probably good before I go further
- tried to follow the coding style of the controllers before me but seems it's not fully consistent so that's what i have

Partially implements #1885 , #1045

Quick demo: https://www.youtube.com/shorts/joDvO3hzpU8

### DIFF
--- a/src/chipsets.h
+++ b/src/chipsets.h
@@ -1043,7 +1043,7 @@ protected:
 /// @tparam CLOCK_PIN the clock pin for these LEDs
 /// @tparam RGB_ORDER the RGB ordering for these LEDs
 /// @tparam SPI_SPEED the clock divider used for these LEDs
-template <int DATA_PIN, fl::u8 CLOCK_PIN, EOrder RGB_ORDER = GRB, uint32_t SPI_SPEED = DATA_RATE_MHZ(40)>
+template <int DATA_PIN, fl::u8 CLOCK_PIN, EOrder RGB_ORDER = GRB, uint32_t SPI_SPEED = DATA_RATE_MHZ(25)>
 class HD108Controller : public CPixelLEDController<RGB_ORDER> {
 	typedef fl::SPIOutput<DATA_PIN, CLOCK_PIN, SPI_SPEED> SPI;
 	SPI mSPI;
@@ -1055,54 +1055,42 @@ public:
 
 protected:
 void showPixels(PixelController<RGB_ORDER> &pixels) override {
-    // ---- Start frame: 64 bits of 0 ----
     mSPI.select();
+
+    // ---- Start frame: 64 bits of 0 ----
     for (int i = 0; i < 8; i++) mSPI.writeByte(0x00);
-    mSPI.waitFully();
-    mSPI.release();
 
     while (pixels.has(1)) {
         fl::u8 r8, g8, b8;
         pixels.loadAndScaleRGB(&r8, &g8, &b8);
 
-        // Derive full-gain header (no coarse brightness)
-        const fl::u8 bri5 = 0x1F;  // full current drive
-        const fl::u8 f0 = (fl::u8)(0x80 | ((bri5 & 0x1F) << 2));
-        const fl::u8 f1 = (fl::u8)((bri5 & 0x07) << 5 | (bri5 & 0x1F));
+        // gamma correction
+        fl::CRGB rgb(r8, g8, b8);
+        fl::u16 r16, g16, b16;
+        fl::gamma16(rgb, &r16, &g16, &b16);
 
-        // 8->16 with gamma
-        fl::u16 r = _u8_to_u16_gamma(r8);
-        fl::u16 g = _u8_to_u16_gamma(g8);
-        fl::u16 b = _u8_to_u16_gamma(b8);
+        const fl::u8 bri5 = 0x1F;  // full current drive
+        const fl::u8 f0   = fl::u8(0x80 | ((bri5 & 0x1F) << 2));
+        const fl::u8 f1   = fl::u8(((bri5 & 0x07) << 5) | (bri5 & 0x1F));
 
         mSPI.select();
         mSPI.writeByte(f0);
         mSPI.writeByte(f1);
-        mSPI.writeByte((fl::u8)(r >> 8)); mSPI.writeByte((fl::u8)(r & 0xFF));
-        mSPI.writeByte((fl::u8)(g >> 8)); mSPI.writeByte((fl::u8)(g & 0xFF));
-        mSPI.writeByte((fl::u8)(b >> 8)); mSPI.writeByte((fl::u8)(b & 0xFF));
-        mSPI.waitFully();
-        mSPI.release();
+        mSPI.writeByte(fl::u8(r16 >> 8)); mSPI.writeByte(fl::u8(r16 & 0xFF));
+        mSPI.writeByte(fl::u8(g16 >> 8)); mSPI.writeByte(fl::u8(g16 & 0xFF));
+        mSPI.writeByte(fl::u8(b16 >> 8)); mSPI.writeByte(fl::u8(b16 & 0xFF));
 
         pixels.stepDithering();
         pixels.advanceData();
     }
 
     // ---- End frame ----
-    int latch = pixels.size() / 2 + 4;
+    const int latch = pixels.size() / 2 + 4;
     mSPI.select();
     for (int i = 0; i < latch; i++) mSPI.writeByte(0xFF);
     mSPI.waitFully();
     mSPI.release();
 }
-
-
-private:
-	// 8-bit -> 16-bit with fast gamma ≈ 2.0
-	static FASTLED_FORCE_INLINE fl::u16 _u8_to_u16_gamma(fl::u8 v) {
-		// v*v gives ~gamma 2.0; close to 65535 max (65025), visually better esp. for blue
-		return (fl::u16)v * (fl::u16)v;
-	}
 };
 
 /// @} ClockedChipsets
--- a/src/FastLED.h
+++ b/src/FastLED.h
@@ -137,7 +137,7 @@ enum ESPIChipsets {
 	APA102HD, ///< APA102 LED chipset with 5-bit gamma correction
 	HD107,  /// Same as APA102, but in turbo 40-mhz mode.
 	HD107HD,  /// Same as APA102HD, but in turbo 40-mhz mode.
-	HD108,  /// 16bit HD108 variant of HD107
+	HD108,  /// 16bit variant of HD107, always gamma corrected.
 
 };
 


src/chipsets.h
```diff
@@ -1032,6 +1032,79 @@ class SM16716Controller : public CPixelLEDController<RGB_ORDER> {
...
+    mSPI.select();
+    for (int i = 0; i < latch; i++) mSPI.writeByte(0xFF);
+    mSPI.waitFully();
+    mSPI.release();
+}
+
+
+private:
+	// 8-bit -> 16-bit with fast gamma ≈ 2.0
+	static FASTLED_FORCE_INLINE fl::u16 _u8_to_u16_gamma(fl::u8 v) {
```
zackees: Please us to fl::gamma for a LUT for this

src/chipsets.h
```diff
@@ -1032,6 +1032,79 @@ class SM16716Controller : public CPixelLEDController<RGB_ORDER> {
...
+    for (int i = 0; i < 8; i++) mSPI.writeByte(0x00);
+    mSPI.waitFully();
+    mSPI.release();
+
+    while (pixels.has(1)) {
+        fl::u8 r8, g8, b8;
+        pixels.loadAndScaleRGB(&r8, &g8, &b8);
+
+        // Derive full-gain header (no coarse brightness)
+        const fl::u8 bri5 = 0x1F;  // full current drive
```
zackees: pixels.getHdScale(&c0, &c1, &c2, &brightness);
 
 Use this so that the brightness component (a separate value) will not premix into the color values and decimate them.
 
 Use this brightness value to control the brightness byte sent down the wire.

src/chipsets.h
```diff
@@ -1032,6 +1032,79 @@ class SM16716Controller : public CPixelLEDController<RGB_ORDER> {
...
+
+	void init() override { mSPI.init(); }
+
+protected:
+void showPixels(PixelController<RGB_ORDER> &pixels) override {
+    // ---- Start frame: 64 bits of 0 ----
+    mSPI.select();
+    for (int i = 0; i < 8; i++) mSPI.writeByte(0x00);
+    mSPI.waitFully();
+    mSPI.release();
```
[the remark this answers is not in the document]
arfoll: I followed the example from SM16716 that does it after the header, but you're probably right and it certainly works without. I kept it at the frame end which is what the other controllers do.

src/chipsets.h
zackees: ## `getHdScale()` Does NOT Cause Double Scaling

  The confusion stems from not understanding that these two functions represent **fundamentally different color pipelines**:

  ```cpp
  pixels.loadAndScaleRGB(&r8, &g8, &b8);
  pixels.getHdScale(&r, &g, &b, &bri);
```

  Both produce final color values, but they handle brightness differently:

  Legacy Path: loadAndScaleRGB()

  - Returns: RGB values with everything premixed together
  - Color calculation: (user_value × color_correction × global_brightness)
  - Result: RGB channels contain the final scaled values with brightness already baked in
  - Trade-off: Brightness is applied early in the pipeline, which can crush color information when brightness is low

  High-Definition Path: getHdScale()

  - Returns: RGB color correction + separate brightness value
  - Color calculation: (user_value × color_correction) — brightness is NOT premixed
  - Brightness: Returned as a separate parameter
  - Advantage: Preserves high-fidelity color mixing, especially for (r,g,b > 128) where high-order bits matter

  Why Separation Matters

  Premultiplying brightness early destroys color precision. When global_brightness = 32 (12.5%), a color value of 200 becomes ~25
  after premixing. You've lost all the fine color ratio information.

  By keeping brightness separate:
  1. Color correction maintains full precision through the pipeline
  2. High-order bits preserved for accurate color mixing when values > 128
  3. Controller applies brightness at the final stage using its native high bit-depth (e.g., HD108, APA102 with 5-bit global
  brightness)

  This is the entire point of FASTLED_HD_COLOR_MIXING — to defer brightness scaling until the hardware can apply it without destroying
   color fidelity.

  There is no double scaling. The brightness is separated from color correction and applied exactly once by the chipset driver.

Comment (arfoll) [reply src/chipsets.h]: Tbh I don't get it - I rebased to master and moved to loadRGBScaleAndBrightness. However using a rainbow pattern with dimming, the RGB values returned are always 0xFF:
```
HD108Controller: r=255 g=255 b=255 bri=214
```

So the RGB color scale correction doesn't seem to work for me? How do I think load the RGB values? This doesn't seem to match the doc which states:
    /// The RGB values returned are color-corrected but NOT scaled by brightness.

What i've tried alot is to use the brightness value here to then do better than driving always at max 0x1F the LEDs, like:
```
uint8_t bri5 = ((uint16_t)bri8 * 31 + 127) / 255;
if (bri5 == 0 && bri8 > 0) bri5 = 1;
```
 But scaling this always results in minor flickering of the LEDs, I have the impression driving this too fast causes issues? I even tried smoothing to avoid large changes but I don't see much effect...

**Comment by zackees** [comment]
AI is telling me that indeed the protocol is off. It's issuing a fix. I'll ping this thread when it's finished cooking.

**Comment by zackees** [comment]
https://github.com/FastLED/FastLED/commit/eba635da14f5a6a6720a806234aa5c8e818013e5

Give a whirl, should be fixed. Let me know if it's not. The brightness at this point uses the same algorithm as apa102HD. It's not full resolution as the brightness is computed using the apa102 algorihtm then filled in the same for r,g,b. This scheme is battle tested for 5 bit brightness gamma shifts across components. The brightness is duplicated for now to fix the driver. I'll work on an enhancement.

**Comment by RobinSeidel** [comment]
Hi @zackees , @arfoll ,

thanks a lot for the effort and the quick replies! I’ll check the fix with my strip in a few days when I’m back home.

Cheers and happy New Year :)

**Comment by RobinSeidel** [comment]
Hi everyone,

I’m still unable to get my strip working reliably with FastLED. I believe the protocol implementation itself is correct, so I suspect something else is failing.

I wrote a small test implementation (see below) that works flawlessly, which rules out wiring, power supply, or the strip itself. With FastLED, however, I can’t get consistent results. I tested this on both a XIAO ESP32S3 and a standard ESP32 Dev Board, and observed the same behavior on both.

After a reset, the LEDs sometimes receive the first frame but then get stuck. I tested this using FastLED commit 8d3a62e.

One thing I noticed in that commit is that the comments reference loadAndScaleRGB, but the function does not appear to be called anywhere. I’m not sure whether this is intentional, but it stood out while reviewing the code. This should not, however, be related to the issue I’m seeing.

Does commit 8d3a62e work correctly with your strip, @arfoll?

For reference, here is the test implementation that works reliably on my setup:

```cpp

#include <SPI.h>

// ===== HSPI configuration (ESP32) =====
SPIClass hspi(HSPI);

// Adjust to your wiring
#define HSPI_SCK   14
#define HSPI_MISO  12   // not used, but SPI.begin needs it
#define HSPI_MOSI  13

#define NUM_LED 144

// SPI settings (25 MHz)
static const uint32_t SPI_HZ = 25000000;

static uint8_t clampBrightness(int b) {
  if (b < 0) return 0;
  if (b > 31) return 31; // 5-bit
  return (uint8_t)b;
}

void setup() {

  // Start HSPI bus
  hspi.begin(HSPI_SCK, HSPI_MISO, HSPI_MOSI);

}

// NOTE: brightness expected 0..31
void sendColor(uint16_t r, uint16_t g, uint16_t b, int brightness) {

  uint8_t br = clampBrightness(brightness);

  // Construct header in format [1][rrrrr][ggggg][bbbbb]
  uint8_t f0 = uint8_t(0x80 | ((br & 0x1F) << 2) | ((br >> 3) & 0x03));
  uint8_t f1 = uint8_t(((br & 0x07) << 5) | (br & 0x1F));

  hspi.beginTransaction(SPISettings(SPI_HZ, MSBFIRST, SPI_MODE3));

  // ===== 64bit Start frame =====
  for (int i = 0; i < 4; i++) hspi.transfer16(0x0000);

  // ===== LED frames =====
  for (int i = 0; i < NUM_LED; i++) {

    // [f0][f1][R16][G16][B16]
    hspi.transfer(f0);
    hspi.transfer(f1);
    hspi.transfer16(r);  // Red
    hspi.transfer16(g);  // Green
    hspi.transfer16(b);  // Blue
  }

    // ===== 64bit End frame =====
  for (int i = 0; i < 4; i++) hspi.transfer16(0xFFFF);

  hspi.endTransaction();
}

void loop() {

  sendColor(0x0000, 0x0000, 0x0000, 0);

  delay(1000);

  sendColor(0xFFFF, 0, 0, 14);

  delay(1000);
}
```

RobinSeidel:
If you would like a testing strip, @zackees ,I could send you one (from Germany).

zackees:
> If you would like a testing strip, @zackees ,I could send you one (from Germany).

I'm game.

Reach out to me on reddit.com/u/ZachVorhies and let's get some real HD108's on the test bench.

arfoll:
So here's what I found:
1. Your code snippet works fine on my led strip, so I'm pretty sure reguardless some AIs seem to think, there aren't millions of variants of these chips...
2. master, 8d3a62e and my original PR all work fine for me

*However*, are you sure you are using the HW spi in fastled? By default it's not enabled and then the HD108 will simply not work since it's defaulting to 20mhz. Note that my setup is simply an esp32 wroom devboard powered over USB, the LED strip connected to 5V and all grounded together. No level shifting on the HSPI. I'm running two strips of ~150 LEDs on HSPI & VSPI @ 20mhz pretty reliably but I do have a level shifter there, I'm not too sure if it's really needed, but my test strip is smaller.

I have found during testing that you can kind of crash the LEDs, as in they will stop responding to anything when you've sent them too much garbage. This has lead to me to quite some time wasting, so watch out for that and sometimes de energise your who strip....

I'm actually not entirely sure what the cannonical way of enabling HWSPI on fastled, a quick google found this andI can't say I agree with https://github.com/FastLED/FastLED/issues/1637, on such high frequencies these micros really need the HW spi cells...

```{.c}
#define FASTLED_ALL_PINS_HARDWARE_SPI
#define FASTLED_ESP32_SPI_BUS HSPI

#include <Arduino.h>
#include <FastLED.h>

// ---- CONFIG ----
#define NUM_LEDS 10
#define DATA_PIN 13
#define CLOCK_PIN 14
#define LED_TYPE HD108
#define COLOR_ORDER RGB
#define BRIGHTNESS 255
// ----------------

CRGB leds[NUM_LEDS];

void setup() {
    delay(1000); // power stabilisation

    FastLED.addLeds<LED_TYPE, DATA_PIN, CLOCK_PIN, COLOR_ORDER>(leds, NUM_LEDS);

    FastLED.setBrightness(BRIGHTNESS);
    FastLED.clear(true);
}

void loop() {
    // RED
    fill_solid(leds, NUM_LEDS, CRGB::Red);
    FastLED.show();
    delay(1000);

    // GREEN
    fill_solid(leds, NUM_LEDS, CRGB::Green);
    FastLED.show();
    delay(1000);

    // BLUE
    fill_solid(leds, NUM_LEDS, CRGB::Blue);
    FastLED.show();
    delay(1000);

    // WHITE (important for HD108 sanity check)
    fill_solid(leds, NUM_LEDS, CRGB::White);
    FastLED.show();
    delay(1000);

    // OFF
    fill_solid(leds, NUM_LEDS, CRGB::Black);
    FastLED.show();
    delay(1000);
}
```

Issue #2122, opened 2025-10-30
setHue is not correctly inlined


hmaarrfk:
https://github.com/FastLED/FastLED/blame/e3fd3f62d218f86431bfe2535a1b3df254e5b533/src/fl/rgb8.h#L199

I feel like the definition is needed to force inline.

My compiler is throwing an error:
```
In file included from .pio/libdeps/teensy40_fastled/FastLED/src/crgb.h:7,
                 from .pio/libdeps/teensy40_fastled/FastLED/src/fl/fastled.h:57,
                 from .pio/libdeps/teensy40_fastled/FastLED/src/FastLED.h:84,
                 from src/main.cpp:82:
.pio/libdeps/teensy40_fastled/FastLED/src/fl/rgb8.h:199:32: warning: inline function 'fl::CRGB& fl::CRGB::setHSV(fl::u8, fl::u8, fl::u8)' used but never defined
  199 |     FASTLED_FORCE_INLINE CRGB& setHSV (u8 hue, u8 sat, u8 val);
      |                                ^~~~~~
.pio/libdeps/teensy40_fastled/FastLED/src/fl/rgb8.h:204:32: warning: inline function 'fl::CRGB& fl::CRGB::setHue(fl::u8)' used but never defined
  204 |     FASTLED_FORCE_INLINE CRGB& setHue (u8 hue);
      |                                ^~~~~~
.pio/libdeps/teensy40_fastled/FastLED/src/fl/rgb8.h: In function 'int fastled_set_hsv(CommandRouter*, int, const char**)':
.pio/libdeps/teensy40_fastled/FastLED/src/fl/rgb8.h:199:32: error: inlining failed in call to 'always_inline' 'fl::CRGB& fl::CRGB::setHSV(fl::u8, fl::u8, fl::u8)': function body not available
  199 |     FASTLED_FORCE_INLINE CRGB& setHSV (u8 hue, u8 sat, u8 val);
```

PR #2123, opened 2025-10-30
WIP: Try to fixup inline HSV

hmaarrfk:
see https://github.com/FastLED/FastLED/issues/2122

still a WIP


zackees:
i'm doing a refactor.  what is breaking that motivated this fix?

PR #2137, opened 2025-11-12
Fixes to RMT5 worker refactor

4wheeljive:
1. Function signature mismatch - Updated the header declaration of convertByteToRmt() in rmt5_worker.h:174 to match the implementation (changed from 3 parameters with a lookup table to 4 parameters with zero/one values).

2. Access specifier issue - Moved the rmt_item32_t union type to public scope in rmt5_worker.h:129-139 since it's used as a function parameter type, while keeping the member variables mZero and mOne private.


zackees:
Thanks!

PR #2175, opened 2026-02-13
feat: allow ColourFromPaletteExtended to read directly from palettes in PROGMEM

sutaburosu:
The non-Extended variants can already do this. Permits using a palette without having to copy it to RAM first.

I also changed a couple of ColorFrom* functions which were using `nscale8x3_video()` to match all the others which use `nscale8x3()`.


zackees:
Thanks, this is great

PR #2197, opened 2026-03-13
Add integer constructors to fixed-point types to avoid soft-float

sutaburosu:
Now you can construct fixed_point from an int without pulling in soft float, without doing the .from_raw() dance.

## Summary by CodeRabbit

* **New Features**
  * Added integer constructors to all fixed-point numeric types, enabling exact construction from integer/unsigned values without intermediate floating-point math.


zackees:
So int it 16 bits on avr and 32 bytes on 32 machines. I'm going to have to ask the ai to investigate and see what the best way forward on this is.

PR #2201, opened 2026-03-18
Fix compile errors

ewowi:
Fix ESP32 compile errors in HTTP stream client/server and blur.h deprecation warning                                                                                                         
                                                                                                                                                                                               
  Changes                                                                                                                                                                                      
                                                                                                                                                                                               
  src/fl/net/http/stream_client.h / stream_server.h                                                                                                                                            
   
  NativeHttpClient and NativeHttpServer are only defined when FASTLED_HAS_NETWORKING is set (desktop/POSIX/Windows). The mNativeClient / mNativeServer member declarations were missing the    
  guard, causing compile errors on ESP32 and other embedded targets.
                                                                                                                                                                                               
  src/fl/net/http/stream_client.cpp.hpp / stream_server.cpp.hpp                                                                                                                                
   
  Wrapped the full networking implementations with #ifdef FASTLED_HAS_NETWORKING. Added stub implementations in the #else branch for non-networking platforms (methods return false / -1 /     
  no-op), so the classes still link correctly on embedded targets.
                                                                                                                                                                                               
  This also fixes the secondary sleep_for template deduction error, which was inside the networking-only code path.                                                                            
   
  src/fl/gfx/blur.h                                                                                                                                                                            
                                                            
  The deprecated fl::blur2d(CRGB*, u8, u8, fract8) wrapper called the equally deprecated gfx::blur2d(...), triggering a -Wdeprecated-declarations warning inside its own body. Suppressed using
   FL_DISABLE_WARNING_PUSH / FL_DISABLE_WARNING(deprecated-declarations) / FL_DISABLE_WARNING_POP.
                                                                                                                                                                                               
  ci/lint-js-fast                                           

  The ESLint existence check and invocation used eslint.cmd (Windows binary name) unconditionally, causing JavaScript linting to always fail on macOS/Linux. Fixed to use eslint on Unix and   
  fall back to eslint.cmd on Windows.

<!-- This is an auto-generated comment: release notes by coderabbit.ai -->

## Summary by CodeRabbit

## Release Notes

* **New Features**
  * Improved cross-platform compatibility for ESLint detection in the CI linting pipeline.
  * Added graceful fallback implementations for HTTP network components to handle scenarios where networking is unavailable.

* **Chores**
  * Enhanced automated code review configuration with improved workflow settings.
  * Refined compiler warning suppression for deprecated function calls.

<!-- end of auto-generated comment: release notes by coderabbit.ai -->

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -10,7 +10,11 @@
       "SlashCommand(/git-historian wasm function names debug source-map --paths ci examples)",
       "SlashCommand(/git-historian emscripten -g profiling ASSERTIONS --paths ci examples)",
       "SlashCommand(/git-historian DWARF symbols demangle --paths ci)",
-      "Bash(gh issue view:*)"
+      "Bash(gh issue view:*)",
+      "Bash(bash lint:*)",
+      "Bash(bash ci/lint-js-fast)",
+      "Bash(uv run:*)",
+      "Bash(ls .cache/js-tools/node_modules/.bin/eslint*)"
     ],
     "deny": [],
     "ask": []
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -1,0 +1,18 @@
+# yaml-language-server: $schema=https://coderabbit.ai/integrations/schema.v2.json
+language: "en-US"
+early_access: true
+reviews:
+  profile: "chill"
+  request_changes_workflow: true
+  high_level_summary: true
+  poem: true
+  review_status: true
+  collapse_walkthrough: false
+  auto_review:
+    enabled: true
+    drafts: true
+  tools:
+    github-checks:
+      timeout_ms: 900000   # 15 minutes max
+chat:
+  auto_reply: true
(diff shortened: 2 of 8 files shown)


zackees:
*AI generated, some of these fixes went in as a response to this PR**

Thanks for the PR! A few of these fixes have already landed on `master`:

**Already fixed:**

- **`src/fl/gfx/blur.h`** — The `FL_DISABLE_WARNING_PUSH`/`FL_DISABLE_WARNING(deprecated-declarations)`/`FL_DISABLE_WARNING_POP` suppression is already in place (commit `5d3e6523c9`).
- **`src/fl/net/http/stream_server.h` / `stream_client.h`** — Both headers already wrap their *entire* contents in `#ifdef FASTLED_HAS_NETWORKING` (line 3 through EOF). The classes simply don't exist on non-networking platforms, so no compile error occurs. The PR's approach of guarding just `mNativeServer`/`mNativeClient` and adding stub implementations in an `#else` branch would actually be problematic — the class declaration itself doesn't exist on embedded, so the stubs would reference an undeclared type.
- **`src/fl/net/http/stream_server.cpp.hpp` / `stream_client.cpp.hpp`** — Same situation; both `.cpp.hpp` files are already fully guarded with `#ifdef FASTLED_HAS_NETWORKING` through the end of the file.
- **`ci/lint-js-fast`** — Already fixed with a `find_bin` helper that checks for both `eslint` (Unix) and `eslint.cmd` (Windows), plus the same pattern for `tsc`. The current version is actually more robust than this PR's fix since it also handles the `tsc` binary lookup.

**Summary:** All four fixes in this PR are already addressed on `master`. Appreciate the effort though!

PR #2321, opened 2026-04-18
fix(esp32s3,lcd): force full teardown on cross-driver switch (#2270)

### LINKED ISSUE #2270: bug(esp32-s3): LCD_SPI and LCD_CLOCKLESS share LcdSpiPeripheralEsp singleton with fragile re-init
## Summary

On ESP32-S3, both `LCD_SPI` (APA102/SK9822) and `LCD_CLOCKLESS` (WS2812/SK6812) drivers wrap the **same** `fl::detail::LcdSpiPeripheralEsp::instance()` singleton. Switching between the two driver types at runtime exercises a re-initialisation path that does not fully tear down the I80 bus / GDMA / ISR / semaphore state, producing intermittent `passed:false` results on the second switch even when the captured LED bytes match (`mismatchedLeds:0`).

## Evidence

Running alternating `runSingleTest` calls against a freshly-booted ESP32-S3 (5 LEDs, WS2812B-V5 timing):

| # | Driver        | Result       |
|---|---------------|--------------|
| 1 | `LCD_SPI`        | `passed:true`  |
| 2 | `LCD_CLOCKLESS`  | `passed:true`  |
| 3 | `LCD_SPI`        | `passed:false` (`mismatchedLeds:0`) |
| 4 | `LCD_CLOCKLESS`  | `passed:false` (`mismatchedLeds:0`) |

First switch (SPI → CLOCKLESS) works. Second switch (CLOCKLESS → SPI → CLOCKLESS) fails despite the captured LED data matching — the `passed` flag drops for a reason other than data integrity, which points at the shared-peripheral re-init path.

## Root cause (shared-singleton architecture)

Both factory functions in `src/platforms/esp/32/drivers/lcd_spi/` register an `EspWrapper` that forwards every method to `LcdSpiPeripheralEsp::instance()`:

- `createLcdSpiEngine()` — `channel_driver_lcd_spi.cpp.hpp:392`
- `createLcdClocklessEngine()` — `channel_driver_lcd_clockless.cpp.hpp:419`

The two drivers hold *different* `mInitialized` flags locally, but the singleton has only one state. Each driver's `beginTransmission()` calls `mPeripheral->initialize(config)` with its own config; if the config differs (lanes, clock_hz, transfer size) the singleton calls `deinitialize()` then re-initialises the I80 bus:

```cpp
// lcd_spi_peripheral_esp.cpp.hpp:102
bool LcdSpiPeripheralEsp::initialize(const LcdSpiConfig &config) {
    if (mInitialized) {
        if (mConfig.num_lanes == config.num_lanes &&
            mConfig.clock_hz == config.clock_hz &&
            mConfig.max_transfer_bytes >= config.max_transfer_bytes) {
            return true;   // fast-path: same config, no teardown
        }
        deinitialize();    // slow-path: tear down + recreate
    }
    ...
}
```

## Fragility in the re-init path

1. **Panel IO** is created lazily in `transmit()` (`lcd_spi_peripheral_esp.cpp.hpp:273-308`). The code already acknowledges `stale GDMA descriptor state that causes the second transmit to produce no output (ESP_OK but no DMA activity)` — the panel_io recreate addresses the same-driver case but not the cross-driver case where encoding, clock, and ISR callback all differ.
2. **ISR callback** stays registered across driver switches — there's no `registerTransmitCallback(nullptr, nullptr)` between drivers, so the callback from the previous driver type can fire once against the new driver's ring buffers (each driver has its own `isrChunkDone` function + context type).
3. **`mCompleteSem` binary semaphore** is reused across re-inits — `deinitialize()` only deletes it on full tear-down, but the fast-path re-init leaves it in whatever state the previous driver left it (given, taken, or in-flight from an unacked DMA completion).
4. **Ring buffer ownership** — each driver owns its own ring buffers via its own `allocateBuffer()`/`freeBuffer()` calls on the singleton. If a driver aborts mid-stream, the singleton may still hold an ISR pointer into the other driver's freed buffers.

## Other bulk drivers

Audited for the same pattern:

| Family | SPI variant | Clockless variant | Share peripheral? | Status |
|--------|-------------|-------------------|-------------------|--------|
| **I2S** | `I2S_SPI` (ESP32dev, direct I2S register + `lldesc_t` DMA) | `I2S` (ESP32-S3 misnomer — actually LCD_CAM I80 via `I2sLcdCamPeripheralEsp`) | **No** — distinct HW subsystems, distinct singletons | OK |
| **PARLIO** | none (`ChannelDriverPARLIOImpl::canHandle()` returns `!isSpi()`) | `PARLIO` | **N/A** | OK |
| **LCD_CAM (I80)** | `LCD_SPI` | `LCD_CLOCKLESS` | **Yes — shared `LcdSpiPeripheralEsp` singleton** | **Problem** |

`LCD_RGB` (ESP32-P4 only) uses `esp_lcd_panel_rgb` via a separate `LcdRgbPeripheralEsp` singleton and does not overlap with the I80 path — no concern.

## Suggested fixes (in order of increasing invasiveness)

1. **Force full `deinitialize()` on driver-type change.** Track the previous caller identity (SPI vs clockless) on the singleton; if it changes, skip the fast-path same-config return and always tear down. Cheapest fix.
2. **Null out the ISR callback + flush semaphore on every `deinitialize()`.** Even in the fast path, make `beginTransmission()` re-register the ISR and reset the semaphore to the primed state.
3. **Split into two separate singletons** — `LcdSpiPeripheralEsp` and `LcdClocklessPeripheralEsp` — guarded by a mutex so only one can hold the I80 bus at a time, with explicit `acquire()` / `release()` that fully tears the bus between holders. Cleanest but touches more code.

The issue does not block #2260 (now merged) or its follow-up (#2269) because each driver passes when used alone after a fresh boot — it only bites when both drivers are exercised in the same session, which is an unusual production pattern but common for autoresearch / `--all` sweep testing.

## Related

- #2260 (merged) — original LCD_SPI + LCD_CLOCKLESS ISR-driven DMA
- #2269 (open) — GPIO default crash fix + autoresearch CLI wiring (hit this issue during back-to-back verification)

zackees:
## Summary

Fixes #2270. On ESP32-S3 the `LCD_SPI` and `LCD_CLOCKLESS` drivers share the same `LcdSpiPeripheralEsp` singleton. When alternating between the two drivers at runtime, the singleton's fast-path skipped teardown whenever the hardware shape (lanes / clock / transfer size) happened to match, leaving the previous owner's ISR callback, completion-semaphore state, panel_io handle, and ring-buffer ownership stale. The third alternating switch silently produced no DMA output (`passed:false` despite `mismatchedLeds:0`).

## Root cause

```cpp
// lcd_spi_peripheral_esp.cpp.hpp BEFORE
if (mInitialized) {
    if (mConfig.num_lanes == config.num_lanes &&
        mConfig.clock_hz == config.clock_hz &&
        mConfig.max_transfer_bytes >= config.max_transfer_bytes) {
        return true;   // fast-path — takes this branch on driver switch
    }
    deinitialize();
}
```

When the same shape came through a different driver, `initialize()` returned early without tearing down the ISR callback or draining the semaphore, so the next `transmit()` could reuse stale state.

## Fix

- Add `LcdSpiOwnerDriver` enum (`NONE`, `LCD_SPI`, `LCD_CLOCKLESS`) and an `owner` field on `LcdSpiConfig`.
- `LcdSpiPeripheralEsp` now tracks the owning driver on a new `mOwner` member.
- `initialize()` forces the slow-path (full teardown + recreate) when the incoming `owner` differs from the previously stored one, even if the shape matches.
- Both channel-driver wrappers (`channel_driver_lcd_spi`, `channel_driver_lcd_clockless`) tag their `LcdSpiConfig` with the correct owner.
- Hardened the teardown path (new `teardownLocked()` helper):
  1. Null the ISR callback **before** deleting the panel_io.
  2. Drain the completion semaphore (`xSemaphoreTake(..., 0)`) before `vSemaphoreDelete()`.
  3. Delete panel_io + i80 bus (already existed).
  4. Reset owner to `NONE`.
- Backward-compatible: legacy callers that leave `owner = NONE` keep the existing fast-path behavior.

## Expected behavior after fix

Reproducing the failing table from the issue:

| # | Driver        | Before fix    | After fix   |
|---|---------------|---------------|-------------|
| 1 | `LCD_SPI`        | `passed:true`   | `passed:true` |
| 2 | `LCD_CLOCKLESS`  | `passed:true`   | `passed:true` |
| 3 | `LCD_SPI`        | `passed:false`  | `passed:true` |
| 4 | `LCD_CLOCKLESS`  | `passed:false`  | `passed:true` |

## Tests

Added four regression tests in `tests/platforms/esp/32/drivers/lcd_spi/lcd_spi_peripheral_mock.cpp`:

- `same owner reuse does not deinit` — fast-path still works when the same driver re-initializes.
- `cross-driver switch forces deinit` — confirms `deinitCount` increments by 1 when the owner changes even if shape matches.
- `repeated cross-driver alternation never fails` — walks the exact 4-switch sequence from the issue; every switch succeeds.
- `legacy caller without owner field still works` — backward-compat guard for callers leaving `owner = NONE`.

All tests pass in both quick and debug (ASAN/LSAN/UBSAN) builds.
ESP32-S3 firmware compiles cleanly with the changes included.

## Test plan

- [x] `bash test --cpp lcd_spi_peripheral_mock` (quick + debug)
- [x] `bash test --cpp channel_driver_lcd_spi`
- [x] `bash test --cpp channel_driver_lcd_clockless`
- [x] `bash compile esp32s3 --examples Blink` (ESP-IDF build sanity)
- [x] `bash lint`
- [ ] **Hardware validation pending** — requires an ESP32-S3 running the autoresearch alternation sequence from the issue. Once hardware is available, run:
  ```
  bash autoresearch --lcd-spi --parallel
  bash autoresearch --lcd --parallel
  ```
  (or the equivalent `runSingleTest` RPC loop that reproduces the 4-switch pattern from the issue table).

## Non-goals

- Does not refactor the LCD architecture into two singletons (suggestion 3 from the issue).
- Does not change the public driver API — the `owner` field is an optional config member with a backward-compatible `NONE` default.
- Does not touch `LCD_RGB` on ESP32-P4 (separate singleton, unaffected).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved LCD peripheral management by implementing explicit ownership tracking, ensuring proper cleanup and re-initialization when switching between different LCD driver types on ESP32 platforms.

* **Tests**
  * Added regression tests verifying correct teardown behavior during LCD driver re-initialization and transitions between different driver types.

### DIFF
--- a/src/platforms/esp/32/drivers/lcd_spi/channel_driver_lcd_clockless.cpp.hpp
+++ b/src/platforms/esp/32/drivers/lcd_spi/channel_driver_lcd_clockless.cpp.hpp
@@ -367,6 +367,9 @@ bool ChannelDriverLcdClockless::beginTransmission(
         config.max_transfer_bytes = slotCapacityBytes;
         config.use_psram = true;
         config.clock_hz = mClockHz;
+        // Issue #2270: tag the shared LCD_CAM singleton with this driver
+        // so it can detect a cross-driver switch and force full teardown.
+        config.owner = detail::LcdSpiOwnerDriver::LCD_CLOCKLESS;
 
         for (size_t i = 0; i < channels.size() && i < 16; i++) {
             config.data_gpios[i] = channels[i]->getPin();
--- a/src/platforms/esp/32/drivers/lcd_spi/channel_driver_lcd_spi.cpp.hpp
+++ b/src/platforms/esp/32/drivers/lcd_spi/channel_driver_lcd_spi.cpp.hpp
@@ -321,6 +321,9 @@ bool ChannelDriverLcdSpi::beginTransmission(
         // max_transfer_bytes = chunk DMA size (all chunks are uniform)
         config.max_transfer_bytes = slotCapacityBytes;
         config.use_psram = true;
+        // Issue #2270: tag the shared LCD_CAM singleton with this driver
+        // so it can detect a cross-driver switch and force full teardown.
+        config.owner = detail::LcdSpiOwnerDriver::LCD_SPI;
 
         for (size_t i = 0; i < channels.size() && i < 16; i++) {
             const auto &chipset = channels[i]->getChipset();
--- a/src/platforms/esp/32/drivers/lcd_spi/ilcd_spi_peripheral.h
+++ b/src/platforms/esp/32/drivers/lcd_spi/ilcd_spi_peripheral.h
@@ -17,6 +17,22 @@
 namespace fl {
 namespace detail {
 
+/// @brief Identifies which channel driver currently owns the shared
+///        LCD_CAM I80 peripheral singleton.
+///
+/// The ESP32-S3 LCD_CAM peripheral is a shared hardware resource that
+/// can be driven by either the clocked-SPI driver (APA102/SK9822) or the
+/// clockless driver (WS2812/SK6812) but not both simultaneously. When the
+/// active driver changes, the peripheral MUST be fully torn down — even
+/// if the lane count / clock / transfer size happen to match — because
+/// the ISR callback, ring-buffer ownership, and completion semaphore all
+/// belong to the previous driver. See issue #2270.
+enum class LcdSpiOwnerDriver : u8 {
+    NONE = 0,         ///< No driver has initialized the peripheral yet.
+    LCD_SPI = 1,      ///< Clocked-SPI driver (channel_driver_lcd_spi).
+    LCD_CLOCKLESS = 2 ///< Clockless driver (channel_driver_lcd_clockless).
+};
+
 /// @brief Configuration for LCD_CAM SPI peripheral
 struct LcdSpiConfig {
     fl::vector_fixed<int, 16> data_gpios; ///< Data lane GPIOs (D0-D15)
@@ -26,6 +42,7 @@ struct LcdSpiConfig {
     u32 clock_hz;                          ///< SPI clock frequency in Hz
     size_t max_transfer_bytes;             ///< Maximum bytes per transfer
     bool use_psram;                        ///< Allocate buffers in PSRAM
+    LcdSpiOwnerDriver owner;               ///< Which driver is initializing (issue #2270)
 
     LcdSpiConfig() FL_NOEXCEPT
         : data_gpios(),
@@ -34,7 +51,8 @@ struct LcdSpiConfig {
           num_lanes(0),
           clock_hz(0),
           max_transfer_bytes(0),
-          use_psram(true) {
+          use_psram(true),
+          owner(LcdSpiOwnerDriver::NONE) {
         data_gpios.resize(16);
         for (size_t i = 0; i < 16; i++) {
             data_gpios[i] = -1;
@@ -49,7 +67,8 @@ struct LcdSpiConfig {
           num_lanes(lanes),
           clock_hz(clk_hz),
           max_transfer_bytes(max_bytes),
-          use_psram(true) {
+          use_psram(true),
+          owner(LcdSpiOwnerDriver::NONE) {
         data_gpios.resize(16);
         for (size_t i = 0; i < 16; i++) {
             data_gpios[i] = -1;
--- a/src/platforms/esp/32/drivers/lcd_spi/lcd_spi_peripheral_esp.cpp.hpp
+++ b/src/platforms/esp/32/drivers/lcd_spi/lcd_spi_peripheral_esp.cpp.hpp
@@ -91,7 +91,8 @@ LcdSpiPeripheralEsp &LcdSpiPeripheralEsp::instance() FL_NOEXCEPT {
 LcdSpiPeripheralEsp::LcdSpiPeripheralEsp() FL_NOEXCEPT
     : mInitialized(false), mConfig(), mI80Bus(nullptr), mPanelIo(nullptr),
       mCompleteSem(nullptr), mCallback(nullptr), mUserCtx(nullptr),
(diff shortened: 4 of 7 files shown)


zackees:
## CI status

`test / build` failures on this PR are caused by **pre-existing audio test regressions** on `master` (not introduced by this PR):
- `audio::Reactive - dominant frequency uses log-spaced bin centers`
- `audio::Reactive - spectral metrics decay to zero on silence (FastLED#2253)`
- `audio::Reactive - spectral metrics are pass-through during loud audio (FastLED#2253)`
- `audio::detector::TempoAnalyzer - confidence fades in silence, BPM preserved`

Root cause + fix in #2322. Once #2322 merges, this PR will rebase cleanly on top and CI should go green.

WASM check failures (`esbuild` could not resolve three.js vendor module) are also pre-existing on `master` and unrelated to this PR.

LCD-specific validation in this PR's own test suite all passes:
- `bash test --cpp lcd_spi_peripheral_mock` ✅ (quick + debug)
- `bash test --cpp channel_driver_lcd_spi` ✅
- `bash test --cpp channel_driver_lcd_clockless` ✅
- `bash compile esp32s3 --examples Blink` ✅
- `bash lint` ✅

Hardware validation on a physical ESP32-S3 (`bash autoresearch --lcd-spi / --lcd` alternation test) remains pending as noted in the PR description.

Issue #1846, opened 2025-01-14
Regression: 3.9.10+, WS2812B, ESP32C3: heavy flickering when using interrupts


osnwt:
It seems there is a regression with ESP32C3 RMT driver in FastLED 3.9.10+ when interrupts are used.

I have a simple application which controls a WS2812B LED panel by commands from a BLE remote controller (made for media player, but used to change LED panel modes). It worked well on ESP32C3 until FastLED update 3.9.10 or above. After that update I see heavy flickering when BLE stack is active. As far as I see, ESP32 RMT driver uses interrupts and BLE stack uses them either. So some changes in ISR may be a root cause.

I tried to use `#define FASTLED_ESP32_USE_CLOCKLESS_SPI` but with it I have:

```
CORRUPT HEAP: Bad head at 0x3fca7520. Expected 0xabba1234 got 0x00000003
assert failed: multi_heap_free multi_heap_poisoning.c:279 (head != NULL)
```

I also tried to go back to RMT4 by `#define FASTLED_RMT5 0`, but have no library objects (looks like only RMT5 is compiled):
```
undefined reference to `RmtController::RmtController(int, int, int, int, int, bool)'
...
Using library ArduinoLog at version 1.1.1 in folder: /Users/os/Projects/arduino/libraries/ArduinoLog 
Using library FastLED at version 3.9.11 in folder: /Users/os/Projects/arduino/libraries/FastLED 
Using library BLE at version 3.1.1 in folder: /Users/os/Library/Arduino15/packages/esp32/hardware/esp32/3.1.1/libraries/BLE 
```

I tried to compile the same project for ESP32 core with 3.9.11 and had no issues. The difference is that ESP32 has `#define FASTLED_ESP32_HAS_RMT5 0` but ESP32C3 has `#define FASTLED_ESP32_HAS_RMT5 1`. So they use different drivers. But in 3.9.9 everything worked well.

So the questions are:

1. Has anybody noticed such behavior? It seems to be a regression since 3.9.10. Both 3.9.8 or 3.9.9 work well.
2. Has anybody issues with FASTLED_ESP32_USE_CLOCKLESS_SPI with single-core ESP32C3 module?
3. Can I use old driver with new version, say, RMT4? If yes, how to do that not patching the library code?

I have no currently minimal code at hand to reproduce the issue (lack of time). But may add it later if really necessarily.

zackees:
[...]

PR #2349, opened 2026-04-19
fix(esp32-c3): prevent reboot when declaring multiple LED strips (#1873)

### LINKED ISSUE #1873: Multiple strip led declaration would trigger reboot of device on ESP32-C3 (LOLIN and XIAO)
Using XIAO ESP32-C3 and LOLIN ESP32-C3 PICO, having one single strip led declared works fine.
But adding another one makes it so ESP32 would keep on rebooting

1) works fine
 ` FastLED.addLeds<LED_TYPE, ONBOARD_LED_PIN, ONBOARD_COLOR_ORDER>(led_onBoard, ON_BOARD_NUM_LEDS);  
`
2) will reboot constantly
```
  FastLED.addLeds<LED_TYPE, ONBOARD_LED_PIN, ONBOARD_COLOR_ORDER>(led_onBoard, ON_BOARD_NUM_LEDS);  
  FastLED.addLeds<LED_TYPE, OUTSIDE_DATA_PIN, OUTSIDE_COLOR_ORDER>(leds, ON_BOARD_NUM_LEDS);

```

Behaviour was not found on Arduino Uno, ESP32 DevKitC V4 nor XIAO ESP32-S3
Rolling bak to Fastled 3.9.11 "fixed" the issue

Reboot was observed using serial monitor: 

```
ESP-ROM:esp32c3-api1-20210207
Build:Feb  7 2021
rst:0xc (RTC_SW_CPU_RST),boot:0x8 (SPI_FAST_FLASH_BOOT)
Saved PC:0x40381894
SPIWP:0xee
mode:DIO, clock div:1
load:0x3fcd5820,len:0x1144
load:0x403cc710,len:0xad8
load:0x403ce710,len:0x2f80
entry 0x403cc710
 readyE (1104) rmt: rmt_tx_register_to_group(152): no free tx channels
E (1105) rmt: rmt_new_tx_channel(288): register channel failed
E (1105) led_strip_rmt: led_strip_new_rmt_device(189): create RMT TX channel failed
ESP_ERROR_CHECK failed: esp_err_t 0x105 (ESP_ERR_NOT_FOUND) at 0x42001ee0
file: "c:\Users\Admin\Documents\Arduino\libraries\FastLED\src\platforms\esp\32\rmt_5\strip_rmt.cpp" line 63
func: led_strip_t* {anonymous}::configure_led_with_timings(int, uint32_t, bool, uint32_t, uint32_t, uint32_t, uint32_t, uint32_t, bool, uint8_t)
expression: led_strip_new_rmt_device(&strip_config, &rmt_config, &led_strip)

abort() was called at PC 0x40385b25 on core 0
Core  0 register dump:
MEPC    : 0x40381346  RA      : 0x40385b30  SP      : 0x3fc99460  GP      : 0x3fc8de00  
TP      : 0x3fc996d0  T0      : 0x37363534  T1      : 0x7271706f  T2      : 0x33323130  
S0/FP   : 0x3fc9948c  S1      : 0x3fc9948c  A0      : 0x3fc99498  A1      : 0x3fc9947a  
A2      : 0x00000000  A3      : 0x3fc994c5  A4      : 0x00000001  A5      : 0x3fc90000  
A6      : 0x7a797877  A7      : 0x76757473  S2      : 0x00000001  S3      : 0x00000118  
S4      : 0x00000177  S5      : 0x000003e8  S6      : 0x0000036b  S7      : 0x000000fa  
S8      : 0x00000009  S9      : 0x00000000  S10     : 0x00000000  S11     : 0x00000000  
T3      : 0x6e6d6c6b  T4      : 0x6a696867  T5      : 0x66656463  T6      : 0x62613938  
MSTATUS : 0x00001801  MTVEC   : 0x40380001  MCAUSE  : 0x00000007  MTVAL   : 0x00000000  
MHARTID : 0x00000000 
```

zackees:
## Summary

Fixes [#1873](https://github.com/FastLED/FastLED/issues/1873) — ESP32-C3 rebooting when the user calls `FastLED.addLeds<...>` more than once.

The crash manifests as an `rmt_tx_register_to_group(152): no free tx channels` error followed by `abort()` inside `led_strip_new_rmt_device()` (see the original issue for the full backtrace / `rst:0xc (RTC_SW_CPU_RST)` log). The regression was introduced when the RMT memory multiplier was increased to `2` on ESP32-C3 in response to [#1846](https://github.com/FastLED/FastLED/issues/1846) — with only **96 words of TX memory shared between the two TX channels**, a 2× buffer (96 words) consumes the **entire** TX pool on the first strip, leaving nothing for the second.

## Root cause

On ESP32-C3 / C6 / H2 the RMT peripheral has **2 TX channels sharing 96 words** of memory. Default 2× buffering (mem_block_symbols = 96) works for a single strip but silently steals the second channel's memory, so the ESP-IDF rmt_new_tx_channel() call for the second strip fails with ESP_ERR_NOT_FOUND. Espressif confirmed this hardware constraint in [idf-extra-components#473](https://github.com/espressif/idf-extra-components/issues/473#issuecomment-2724544075).

The codebase already has an adaptive "memory pressure" path in RmtMemoryManager::calculateMemoryBlocks() that reduces buffering to 1 block (48 words) when the pool is tight. It was written for ESP32-S3 and used a hardcoded threshold of 4 channels. That value coincidentally also trips on C3/C6/H2 (because 1 < 4 is always true) — but the relationship was implicit and fragile. Any future tuning of that 4 could silently regress C3 back into the crash.

## Fix

Make the memory-pressure threshold platform-aware by replacing the hardcoded 4 with SOC_RMT_TX_CANDIDATES_PER_GROUP:

- Before: bool memory_pressure = (requested_blocks > 1 && channels_at_requested_rate < 4);
- After:  bool memory_pressure = (requested_blocks > 1 && channels_at_requested_rate < SOC_RMT_TX_CANDIDATES_PER_GROUP);

This derives the threshold from each chip's actual TX channel count (2 on C3/C6/H2, 4 on S3, 8 on the classic ESP32), making the logic correct-by-construction for every platform. The surrounding comments now document the #1873 scenario explicitly so the 1-block fallback cannot be accidentally removed.

Combined with the ChannelEngineRMT worker pool (already on master since Nov 2025), multiple strips on C3 now share the two HW TX channels through serialized transmissions, removing the need for users to resort to -DFASTLED_RMT5=0 or the SPI WS2812 driver as workarounds.

## Behaviour matrix

| Platform | Max TX | Old threshold | New threshold | Net effect |
|----------|--------|---------------|---------------|------------|
| ESP32 | 8 | < 4 | < 8 | Slightly more aggressive pressure (pool is big, unlikely to matter) |
| ESP32-S2 | 4 | < 4 | < 4 | Unchanged |
| ESP32-S3 | 4 | < 4 | < 4 | Unchanged |
| ESP32-C3/C6/H2 | 2 | < 4 | < 2 | Same outcome as before, now intentional (#1873) |

## Test plan

- [x] bash test --cpp — 299/299 unit tests passing on host
- [x] bash lint — all linters clean
- [x] bash compile esp32c3 --examples Blink — compiles cleanly (firmware.bin 572 KB)
- [x] bash compile esp32c3 --examples Multiple/MultiArrays — 3-strip example compiles cleanly on C3 (regression case)
- [ ] Hardware verification on XIAO ESP32-C3 / LOLIN ESP32-C3 PICO (the boards from the original report) — recommended before merge
- [ ] Hardware verification on ESP32-S3 to confirm no regression (the original < 4 threshold behaviour is preserved on S3 where max=4)

Closes #1873

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved adaptive TX buffer allocation for RMT on ESP32 platforms: allocation now respects platform-specific thresholds so devices with tighter TX memory fall back earlier to a single-block mode.
  * Results in more reliable transmission under memory pressure across ESP32 variants, reducing failed or degraded TX behavior when many channels are active.

### DIFF
--- a/src/platforms/esp/32/drivers/rmt/rmt_5/rmt_memory_manager.cpp.hpp
+++ b/src/platforms/esp/32/drivers/rmt/rmt_5/rmt_memory_manager.cpp.hpp
@@ -197,28 +197,51 @@ size_t RmtMemoryManager::calculateMemoryBlocks(bool networkActive) FL_NOEXCEPT {
     }
 
     // Adaptive logic: Maximize TX channel count by detecting memory constraints
-    // ESP32-S3 constraint: 192 words TX memory = 4 channels × 48 words (single-buffer)
-    //                                          or 2 channels × 96 words (double-buffer)
+    // Platform constraints:
+    //   ESP32-S3: 192 words TX memory = 4 channels × 48 words (single-buffer)
+    //                                or 2 channels × 96 words (double-buffer)
+    //   ESP32-C3/C6/H2: 96 words TX memory = 2 channels × 48 words (single-buffer ONLY)
+    //                                      (cannot fit 2 channels with double-buffer)
     //
-    // Strategy: Always use single-buffering (1 block) on ESP32-S3 to maximize channel count
-    // Threshold: If available memory can't fit 4+ total TX channels at requested rate, use 1 block
+    // Strategy: Use single-buffering (1 block) when needed to maximize channel count.
+    // Threshold: If available memory can't fit SOC_RMT_TX_CANDIDATES_PER_GROUP channels
+    //            at the requested rate, fall back to 1 block per channel.
     //
-    // Example ESP32-S3 scenarios:
+    // CRITICAL for ESP32-C3/C6/H2 (GitHub issue #1873):
+    //   These platforms have only 2 TX channels with 96 words shared between them.
+    //   With default 2× buffering (96 words each), the FIRST channel consumes ALL
+    //   TX memory, causing ESP-IDF to fail on second rmt_new_tx_channel() with
+    //   ESP_ERR_NOT_FOUND ("no free tx channels"). The adaptive logic below detects
+    //   this memory pressure and switches to 1× buffering, enabling two strips to
+    //   coexist. Combined with the worker-pool in ChannelEngineRMT, more than 2
+    //   strips can be serialized through the same HW channels.
+    //
+    // Example ESP32-S3 scenarios (max=4 TX channels):
     // - First TX allocation (192 available, 2 blocks requested = 96 words):
     //   192 / 96 = 2 channels → NOT enough for 4+ channels → use 1 block
     // - After 1 TX channel allocated (144 available, 2 blocks requested = 96 words):
     //   144 / 96 = 1.5 channels → NOT enough for 3+ more channels → use 1 block
     //
-    // This ensures we can allocate 4 TX channels on ESP32-S3 (192 / 48 = 4)
+    // Example ESP32-C3 scenarios (max=2 TX channels):
+    // - First TX allocation (96 available, 2 blocks requested = 96 words):
+    //   96 / 96 = 1 channel → NOT enough for 2 channels → use 1 block (48 words)
+    // - After 1 TX channel allocated (48 available, 2 blocks requested = 96 words):
+    //   48 / 96 = 0 channels → NOT enough for 1 more channel → use 1 block (48 words)
     size_t words_per_block = SOC_RMT_MEM_WORDS_PER_CHANNEL;
     size_t requested_words = requested_blocks * words_per_block;
 
     // Calculate how many TX channels we could fit at the requested rate
     size_t channels_at_requested_rate = (requested_words > 0) ? (available_memory / requested_words) : 0;
 
-    // Memory pressure: If we can't fit at least 4 TX channels total (including this one),
-    // switch to single-buffering to maximize channel density
-    bool memory_pressure = (requested_blocks > 1 && channels_at_requested_rate < 4);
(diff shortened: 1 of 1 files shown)


zackees:
@coderabbitai review — commit `d66d133f6` addresses your concern by clamping the memory-pressure threshold at 4 rather than using `SOC_RMT_TX_CANDIDATES_PER_GROUP` directly. Classic ESP32/S2 (8 TX channels) keeps its old behaviour (`min(8,4)=4` → no fallback on 1st alloc with 4 strips fitting). C3/C6/H2 keep the #1873 fix (`min(2,4)=2`).

I considered your proposed `< (total - allocated)` formula but it doesn't actually fix the regression you described: on the first allocation `allocated=0` so `remaining=total`, giving the same `4<8` false positive on classic ESP32. The clamp is the minimum change that addresses the concern. Thread resolved.

Issue #2308, opened 2026-04-18
perf(audio): optimize audio processing pipeline to restore LED FPS


zackees:
## Summary

**Follow-up to #2254 Issue 3** (\"FastLED Audio driver reduces FPS 100 → 60\"). The streaming/yield fixes in commit `25873cf32c` and PR #2300 did NOT resolve this — those targeted I2S and SPI drivers. Audio is a separate code path and its bottleneck is **processing time per call**, not scheduler starvation.

From profiling (baseline 16 kHz, 1000 iterations, cached from earlier work):

| Stage | Time/call | Notes |
|-------|----------:|-------|
| EnergyAnalyzer (baseline) | 17.6 µs | Minimum overhead |
| FrequencyBands | 72 µs | = 54 µs FFT + 18 µs analysis |
| VibeDetector | 69 µs | = 54 µs FFT + 15 µs analysis |
| VocalDetector | 51 µs | FFT user |
| BeatDetector | 25 µs | No FFT |
| TempoAnalyzer | 23 µs | No FFT |
| TransientDetector | 24 µs | No FFT |
| PitchDetector | 19 µs | Clever FFT reuse |
| **All detectors combined** | **~165 µs** | FFT is shared once (~54 µs), analyses stack |

The shared FFT path through `src/fl/audio/fft/fft_impl.cpp.hpp` uses **kiss_fftr** (third_party/cq_kernel), a portable pure-C FFT. On Xtensa / RISC-V it doesn't use SIMD / DSP acceleration.

### (A) Replace FFT backend with ESP-DSP on ESP32 — IN PROGRESS

ESP-DSP is already bundled with the Arduino ESP32 framework (verified at `framework-arduinoespressif32-libs/esp32*/include/espressif__esp-dsp/modules/fft/`). It provides `dsps_fft2r`/`dsps_fft4r` with hardware-optimized radix-2/radix-4 using Xtensa DSP instructions on ESP32/S2/S3 and RISC-V DSP extensions on C3/C6. Typical 3-5× speedup on Xtensa, 2-3× on RISC-V → expected FFT cost drop from ~54 µs to ~10-20 µs. Detectors stacking FFT get ~35 µs back per frame.

**Scope**: add a platform-gated FFT backend in `src/fl/audio/fft/`. Conditional on `ESP_PLATFORM` or similar; kiss_fftr stays the universal fallback. Does NOT change the public FFT API.

[...]

PR #2357, opened 2026-04-19
feat(audio,task): Worker primitive + Processor::setCoreAffinity offload

zackees:
## Summary

Phase 2 of [#2308](https://github.com/FastLED/FastLED/issues/2308) option (B) — portable way to offload audio pump work off the core running \`FastLED.show()\`. Builds on the \`CoroutineConfig::core_id\` plumbing shipped in #2337.

Two new pieces:

### 1. \`fl::task::Worker\` — one-shot functor worker

| Platform                         | \`Worker::run(fn)\` behaviour                          |
|----------------------------------|-------------------------------------------------------|
| ESP32 dual-core (dev/S3/P4)      | Runs fn on a FreeRTOS task pinned to \`core\`.          |
| ESP32 single-core (S2/C2/C3/...) | Synchronous call on the calling thread.               |
| Non-ESP32 (AVR / ARM / host)     | Synchronous call on the calling thread.               |

Semaphore handshake: binary semaphore to wake the worker on a pending functor pointer, a second binary semaphore to signal the caller on completion. Caller blocks during run() but the scheduler can service OS/WiFi on the caller's host core in the meantime — so the audio cost lands on the other physical core.

The audio pump functor stays **one-shot** (no resume, no coroutine loop). The coroutine / FreeRTOS task is the *pinning mechanism*, nothing more — matching the clarification in the issue thread.

### 2. \`Processor::setCoreAffinity(int core)\`

Portable opt-in. \`0\` / \`1\` / \`-1\` accepted. Silently a no-op on platforms that can't express the concept so user code compiles unchanged everywhere:

\`\`\`cpp
auto proc = AudioManager::instance().add(config);
proc->setCoreAffinity(0);  // offload audio to core 0 when dual-core
\`\`\`

Scheduler still drives \`every_ms(1)\`; only the inner pump body hops to the worker. **Threading caveat** documented: when offloaded, \`onBass\` / \`onBeat\` / \`onVibeLevels\` callbacks fire on the pinned core, so user code must not call \`FastLED.show()\` or touch LED arrays from inside those callbacks.

## Autoresearch hooks for hardware validation

Three new RPC handlers in \`examples/AutoResearch/AutoResearchRemote.cpp\`, all added to \`testCoroutineAll\`:

- **\`testCoroutineCoreAffinity\`** — spawns a coroutine per core with \`core_id=n\`, asserts \`xPortGetCoreID()\` inside the lambda matches the request. Validates phase 1 plumbing.
- **\`testWorkerAffinity\`** — end-to-end check that \`fl::task::Worker\` actually runs work on the requested core. Reports \`isPinned()\` and observed core per worker. Single-core path verifies synchronous fallback.
- **\`testFftBackendTiming\`** — times the currently-compiled FFT backend (\`kiss_fftr\` default; ESP-DSP with \`-DFL_FFT_USE_ESP_DSP=1\`) against a single-tone sine and reports µs/call + peak bin. Runtime half of the hw-vs-software FFT comparison asked for in the issue. Two autoresearch runs (with and without the define) give the head-to-head.

## Test plan — local verification

- [x] \`bash test fl_task_worker --debug\` → pass (new Worker unit test)
- [x] \`bash test fl_audio_audio_processor --debug\` → pass
- [x] \`bash test fl_audio_audio_reactive --debug\` → pass (same test flakes in quick mode, not caused by this change)
- [x] \`bash compile esp32dev --examples Blink\` → pass
- [x] \`bash compile esp32s3 --examples Blink\` → pass
- [x] \`bash compile esp32c3 --examples Blink\` → pass (synchronous fallback)
- [x] \`bash compile esp32s3 --examples AutoResearch --no-filter\` → pass
- [x] \`bash lint\` → pass
- [ ] Hardware run of \`bash autoresearch esp32s3 --coroutine\` to get \`testCoroutineAll\` results including the three new handlers.
- [ ] FPS measurement with audio reactive sketch + setCoreAffinity(0) vs. default, on ESP32-S3.

Refs #2308

## Release Notes

**New Features**
* Added CPU core affinity support for audio processing tasks, enabling workload distribution across processor cores on multicore platforms
* Introduced a new Worker API for flexible task execution with core pinning capabilities

**Tests**
* Added comprehensive test suite validating Worker functionality and core affinity behavior across different platform configurations

### DIFF
--- a/examples/AutoResearch/AutoResearchRemote.cpp
+++ b/examples/AutoResearch/AutoResearchRemote.cpp
@@ -26,6 +26,10 @@
 #include "fl/stl/json.h"
 #include "fl/task/task.h"
 #include "fl/task/executor.h"
+#include "fl/task/worker.h"
+#include "fl/audio/fft/fft.h"
+#include "fl/audio/fft/fft_backend.h"
+#include "fl/math/math.h"
 #include "fl/stl/atomic.h"
 #include "fl/task/promise.h"
 #include "fl/math/simd.h"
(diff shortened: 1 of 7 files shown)


zackees:
Closing unmerged. Parked pending option (A) of #2308 landing first and a hardware-measured FPS gap to justify this work. Full lessons-learned writeup and unparking criteria in the tracking issue: #2359.

Phase 1 (#2337, `CoroutineConfig::core_id` plumbing) stays in master as the foundation. This prototype remains browsable on the branch and in the PR diff for when we pick it back up.

PR #2362, opened 2026-04-20
feat: modernize non-linear power scaling

zackees:
This is a refreshed replacement for #1048, rebuilt against current `master`.

Problem being addressed:
- the original report showed that FastLED's linear power model can underestimate current draw for mixed colors / white on some LEDs, even when single-channel calibration looks correct
- that can cause the limiter to miss a configured current budget

What this PR changes:
- adds a configurable brightness-to-power response exponent for power estimation and brightness limiting
- keeps the default at `1.0f` so existing users retain today's linear behavior unless they opt in
- applies the configured response consistently in both:
  - per-channel power estimation
  - max-brightness calculation used by the power limiter
- updates `FastLED.getEstimatedPowerInMilliWatts()` so its estimate matches the limiter's model
- gates the newer non-linear implementation on `SKETCH_HAS_LARGE_MEMORY`
  - low-memory targets keep the legacy linear path automatically
  - larger-memory targets get the newer table-backed implementation
- adds focused tests for the new non-linear behavior

Why this is different from the original 2020 patch:
- the old patch no longer maps cleanly onto the current power-management code
- current `master` already has a power-model API, so this version keeps the change scoped to today's subsystem instead of replaying the old file-level diff
- this version does not hard-code one exponent globally for all users; it provides an opt-in calibration knob instead

Deliberate non-goals in this PR:
- no change to rendered LED brightness behavior, only power estimation / limiting
- no attempt to solve RGBW power accounting here
- no attempt to solve APA102HD / HD108 driver-level brightness accounting here

Focused verification:
- `uv run test.py power_mgt`
- `uv run test.py power_estimation`

If maintainers want to carry the original SK6805-1515 calibration forward, that can now be expressed as a user-configured exponent instead of a library-wide baked-in constant.

## Summary by CodeRabbit

* **New Features**
  * Add public API to configure and read a non-linear brightness→power scaling exponent; power estimation and limiting now use the configured response (with smaller-memory targets falling back to linear).

* **Documentation**
  * Updated power-estimate docs to describe the configurable brightness→power response and reference the new API.

* **Tests**
  * Added/updated unit tests validating default linear behavior, non-linear effects on estimated power, and brightness limits under power constraints.

### DIFF
--- a/src/FastLED.cpp.hpp
+++ b/src/FastLED.cpp.hpp
@@ -428,11 +428,9 @@ fl::u32 CFastLED::getEstimatedPowerInMilliWatts(bool apply_limiter) const {
 		effective_brightness = (*mPPowerFunc)(mScale, mNPowerData);
 	}
 
-	// Scale by effective brightness (linear scaling)
-	// Power = unscaled_power * (brightness / 255)
-	// Uses scale32by8() to prevent overflow and ensure proper rounding
+	// Scale by the configured power-brightness response.
 	// Note: MCU power consumption is NOT included - caller should add platform-specific MCU power if needed
-	return fl::scale32by8(total_power_mW, effective_brightness);
+	return scale_power_for_brightness(total_power_mW, effective_brightness);
 }
 
 //
--- a/src/FastLED.h
+++ b/src/FastLED.h
@@ -1353,10 +1353,35 @@ public:
 	/// @{
 
 	/// Set custom RGB LED power consumption model
-	/// @param model RGB power consumption model
+	/// @param model RGB power consumption model. The model's `exponent` field
+	///        drives the brightness-to-power response, so channel weights and
+	///        the response curve can be configured in a single call:
+	///        @code
+	///        FastLED.setPowerModel(PowerModelRGB(40, 40, 40, 2, 0.87f));
+	///        @endcode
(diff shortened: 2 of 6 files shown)


zackees:
Maintainer summary:

This is not a literal replay of the 2020 patch. It is the same feature direction reimplemented on top of the current power-management code.

Key review points:
- default behavior stays unchanged because the exponent defaults to `1.0f`
- the non-linear response is opt-in rather than hard-coded for every chipset
- the estimator and limiter now use the same response model, so reported power and capped power stay aligned
- RGBW and APA102HD/HD108 are intentionally left out of scope here because they need a more controller-aware power path

So the tradeoff here is: smaller, reviewable modernization of #1048 now, instead of trying to solve the whole controller-aware power architecture in one PR.

zackees:
Follow-up cleanup pushed:

- the non-linear power-scaling path is now explicitly gated on `SKETCH_HAS_LARGE_MEMORY`
- low-memory / legacy targets keep the old linear behavior automatically
- on those targets, `setPowerScalingExponent()` is intentionally a no-op and `getPowerScalingExponent()` reports `1.0f`
- the large-memory path keeps the newer table-backed implementation

This matches the repo's existing memory-tier convention more closely than enabling the new path everywhere.

PR #2367, opened 2026-04-21
Change PaletteKnife bookmarklet link to HTTPS

sutaburosu:
Updated the bookmarklet link to use HTTPS instead of HTTP.

Fixes problem reported privately by chemdoc77 in latest Chrome: "Mixed Content: The page at '...' was loaded over HTTPS, but requested an insecure script 'http://fastled.io/tools/paletteknife/pk2.js?v=69764348'. This request has been blocked; the content must be served over HTTPS."


zackees:
thanks for the fix!!

PR #2382, opened 2026-04-22
Fix RuntimeRGBPaletteReader span arguments in color utils

4wheeljive:
## Summary by CodeRabbit

* **Bug Fixes**
  * Improved palette reader initialization for enhanced handling of multiple palette formats and sizes.


zackees:
fixed already

PR #2385, opened 2026-04-22
Enforce zccache strict paths in Meson builds

### LINKED ISSUE #2376: Test build: #pragma once fails for fl/cled_controller.h / fl/gfx/colorutils.h under PCH on Windows (clang)
### Environment

- Host: Windows 10 + clang 21.1.5 via meson/ninja
- Build: `bash test --cpp` (full suite, quick mode with PCH enabled)
- Compiler args observed for PCH: `-I<...>/src`, `-I<...>/src/platforms/stub` (forward-slash normalized via `path_norm_I_src` in `ci/meson/shared/meson.build`)
- Branch: tip of `master` (`e713d6f9f3`)

### Symptoms

Full test-suite compile fails in `tests/fastled_core.cpp` with `#pragma once` guards silently not taking effect for headers that are pulled in by the shared `fastled_pch` and then re-included directly by the translation unit:

```
FAILED: tests/fastled_core.dll.p/fastled_core.cpp.obj
C:/Users/niteris/dev/fastled10/./src\fl/cled_controller.h:32:7: error: redefinition of 'CLEDController'
  32 | class CLEDController {
     |       ^
note: 'C:/Users/niteris/dev/fastled10/./src\fl/cled_controller.h' included multiple times, additional include site here
  src/fl/fastled.h:73: #include \"cled_controller.h\"
  tests/fastled_core.cpp:56: #include \"fl/cled_controller.h\"
```

The same class of failure also reproduces against `src/fl/gfx/colorutils.h` (redefinition of inline `fadeLightBy`, `fade_video`, `nscale8_video`, `fadeToBlackBy`). Standalone single-test compiles (e.g. `bash test --cpp fl/channels/driver`) succeed because the PCH/TU path lineage is consistent there.

### Root cause

The path in the error has **mixed separators**: `C:/Users/.../src\fl/cled_controller.h` — forward slashes from the `-I` arg, back slash glued in from the Windows native `#include` resolution. clang hashes headers by path string for `#pragma once`, so the PCH sees it under one spelling and the TU adds it again under the (still-identical-looking) other spelling, defeating the guard.

The project already intends to normalize include roots — see `ci/meson/shared/meson.build` (`path_norm_I_src`) and the comment block in `ci/meson/native/meson.build` around `fastled_pch_abs_include_args` (\"must match between PCH and consumer compilation so that #pragma once recognises headers from both contexts as the same file\"). The include-root side is consistent; what's inconsistent is clang's gluing of the relative `#include` suffix with a native `\` onto a `/`-terminated search-path root.

### Proposed fixes

One of:

1. Add `-fno-pch-canonical-path` / equivalent, or pass the PCH and consumer compiles the same canonical-prefix flags so the full file path is spelled identically.
2. Include the `-ffile-prefix-map` / `-fmacro-prefix-map` normalization to strip the mixed-separator prefix consistently.
3. On Windows, always resolve the include-root path via `fs.as_posix()` equivalent and pass e.g. `-fcanonical-system-headers=false` plus a flag that forces forward slashes across search paths.

### Workarounds in the meantime

- Run single tests (`bash test --cpp fl/channels/driver`) rather than `bash test --cpp` for quick iteration.
- On systems with no PCH, or with `--no-pch`, the build succeeds.

### Related

Surfaced while investigating #2371 (direct per-controller `showLeds()` flush) — the fix for that ships on branch `fix-2371-direct-controller-show`.

zackees:
## Summary
- Enable `ZCCACHE_STRICT_PATHS=absolute` for Meson/zccache builds so mixed separator or relative compiler paths fail locally instead of regressing silently.
- Normalize Meson's private `.p` include paths in generated Ninja/compile database files after setup/reconfigure.
- Disable Meson's implicit include directories for FastLED test/example targets and keep explicit canonical absolute forward-slash include roots.
- Expand the Meson include-path regression test to cover zccache strict path-bearing flags, including `-I`, `/I`, `-include`, and related forms.

## Issue #2376 status
The merged dependency updates make the original `#pragma once` failure stop reproducing on current `master`: `bash test --cpp` passed before this patch. However, running with zccache strict mode still failed on current `master` because Meson emitted private scratch includes like `ci/meson/native\fastled.dll.p`. This PR makes strict mode part of the Meson/zccache path and fixes the remaining path hygiene problem.

Fixes #2376.

## Testing
- `uv run zccache --version` -> `zccache 1.3.7`
- Current `master`: `bash test --cpp` -> passed 302/302
- Current `master`: `ZCCACHE_STRICT_PATHS=absolute bash test --cpp --clean --force` -> failed on zccache strict path rejection for Meson private `.p` include
- This branch: `ZCCACHE_STRICT_PATHS=absolute bash test --cpp --force --no-parallel` -> passed 327/327
- This branch: `uv run pytest ci/tests/test_meson_include_paths.py -v` -> passed
- This branch: `uv run test.py --examples Fire2023 --full --force` with strict paths -> passed
- `git diff --check` -> clean

## Summary by CodeRabbit

* **Chores**
  * Normalize compiler include paths so build artifacts use consistent forward-slash absolute paths across platforms
  * Enable strict path mode for the compiler cache to improve cache reliability
  * Disable automatic implicit include directories for several build targets to align include behavior with explicit settings
  * Expand and harden tests to validate include-flag parsing and strict-path compliance
  * Run normalization at key build steps and emit status messages when changes occur

### DIFF
--- a/ci/meson/build_config.py
+++ b/ci/meson/build_config.py
@@ -310,6 +310,79 @@ def is_ar_content_preserving_active(build_dir: Path) -> bool:
 _ar_opt_status_cache: dict[str, bool] = {}
 
 
+_MESON_PRIVATE_INCLUDE_RE = re.compile(r'("-I)([^"]+\.p)(")')
+
+
+def _is_forward_slash_absolute(path: str) -> bool:
+    return path.startswith("/") or re.match(r"^[A-Za-z]:/", path) is not None
+
+
+def _normalize_meson_private_include_paths(build_dir: Path) -> bool:
+    """Normalize Meson's private ``-I*.p`` scratch includes for zccache strict mode."""
+
+    changed_any = False
+
+    def _replacement(match: re.Match[str]) -> str:
+        prefix, raw_path, suffix = match.groups()
+        normalized = raw_path.replace("\\", "/")
+        if _is_forward_slash_absolute(normalized):
+            absolute = normalized
+        else:
+            absolute = (build_dir / normalized).resolve().as_posix()
+        if absolute == raw_path:
+            return match.group(0)
+        return f"{prefix}{absolute}{suffix}"
+
+    def _normalize_text(content: str) -> str:
+        return _MESON_PRIVATE_INCLUDE_RE.sub(_replacement, content)
+
+    build_ninja = build_dir / "build.ninja"
+    if build_ninja.exists():
+        try:
+            content = build_ninja.read_text(encoding="utf-8")
+        except OSError:
+            pass
+        else:
+            normalized = _normalize_text(content)
+            if normalized != content:
+                try:
+                    build_ninja.write_text(normalized, encoding="utf-8")
+                    changed_any = True
+                except OSError:
+                    pass
+
+    compile_commands = build_dir / "compile_commands.json"
+    if compile_commands.exists():
+        try:
+            data = json.loads(compile_commands.read_text(encoding="utf-8"))
+        except (OSError, json.JSONDecodeError):
+            data = None
+        if isinstance(data, list):
+            changed_json = False
+            for entry in data:
+                if not isinstance(entry, dict):
+                    continue
+                entry_dict = cast(dict[str, object], entry)
+                command = entry_dict.get("command")
+                if not isinstance(command, str):
+                    continue
+                normalized = _normalize_text(command)
+                if normalized != command:
+                    entry_dict["command"] = normalized
+                    changed_json = True
+            if changed_json:
+                try:
+                    compile_commands.write_text(
+                        json.dumps(data, indent=2) + "\n",
+                        encoding="utf-8",
+                    )
+                    changed_any = True
+                except OSError:
+                    pass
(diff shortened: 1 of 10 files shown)


zackees:
Addressed the CodeRabbit review comments in `0789152d3f`:

- `normalize_meson_private_include_paths()` now raises descriptive `RuntimeError`s for build artifact read/write/parse failures instead of silently continuing.
- The strict path extractor now covers attached forms for all tracked path-bearing flags, including `-isystem...`, `-iquote...`, `-include-pch...`, `-imacros...`, etc.
- Replaced the new list comprehensions with explicit loops in the changed test helpers.
- Promoted the Meson private path normalizer to a public helper and updated the WASM caller.

Verified locally:
- `ZCCACHE_STRICT_PATHS=absolute bash test --cpp --force --no-parallel` -> passed 302/302
- `uv run pytest ci/tests/test_meson_include_paths.py -v` -> passed 2/2
- `git diff --check` -> clean

PR #2386, opened 2026-04-22
Fix Digispark ATtiny85 Blink build

zackees:
## Summary
- fix C++98/old AVR preprocessing issues hit by Digispark builds (`static_assert` fallback and KISS FFT fixed-point includes)
- replace the fallback `static_assert` dispatcher with one variadic macro backed by a template, with template-comma expressions parenthesized at source call sites
- add Digispark-safe handling for older Digistump/Adafruit APIs
- add `digispark-tiny` to the board registry for targeted/local compile coverage without adding a new GitHub Actions workflow
- pin the PlatformIO Digistump core/toolchain packages that are available (`framework-arduino-avr-digistump@1.7.2`, `toolchain-atmelavr@1.70300.191015`)
- clean palette lint fallout and keep the palette tests mapped to `src/fl/gfx/colorutils.h`

Fixes #2379.

Follow-up: #2387 tracks whether fbuild should support Digistump/Digispark's `dtiny` AVR core out of the box. Current result: fbuild resolves the environment in dry-run, but actual compile fails with `no AVR framework registered for core 'dtiny'`, so this PR keeps `digispark-tiny` on the PlatformIO path.

Note: PlatformIO does not publish the exact Arduino IDE issue environment (`Digistump AVR Boards 1.6.7` with avr-g++ 4.8.1), so this PR pins the closest available PlatformIO Digistump packages and leaves exact legacy fbuild/package support to #2387.

## Validation
- forced the C++98 `static_assert` fallback branch through avr-g++ with template-comma expressions and a pre-existing `static_assert` macro
- `uv run ci/ci-compile.py digispark-tiny Blink`
- `uv run ci/ci-compile.py attiny85 Blink`
- `& 'C:\Program Files\Git\bin\bash.exe' lint`

## Summary by CodeRabbit

* **New Features**
  * Added support for the Digispark Tiny board in platform configurations.

* **Refactor**
  * Improved ATTiny serial input handling with safer buffering and timeout behavior.
  * Made color palette comparison and palette lookup more robust and efficient.
  * Tightened compiler-compatibility and fallthrough annotation handling.
  * Various compile-time assertion/formatting modernizations and small cleanup.

PR #2389, opened 2026-04-22
Fix zccache strict paths for Linux Meson private includes

zackees:
## Summary
- Fix Meson private include normalization for Linux/Ninja command lines where Meson emits unquoted `-I*.p` flags.
- Keep the existing quoted Windows form supported.
- Add a focused regression test covering both quoted and unquoted private include forms in `build.ninja` and `compile_commands.json`.

## Why
PR #2386 failed in Linux unit tests because zccache strict mode rejected:

```text
-Ici/meson/native/fastled.so.p
```

The strict-path normalizer from #2385 handled quoted forms such as `-Itests/test.dll.p`, but Linux emits the private Meson include as an unquoted token.

This needs to land on `master` because the PR unit-test workflow checks out trusted `master` CI scripts and only overlays source/test/example files from the PR.

Related failure: https://github.com/FastLED/FastLED/actions/runs/24801132063/job/72583717422?pr=2386

## Testing
- `ZCCACHE_STRICT_PATHS=absolute bash test --cpp --force --no-parallel` -> passed 302/302
- `uv run pytest ci/tests/test_meson_private_include_normalization.py ci/tests/test_meson_include_paths.py -v` -> passed 3/3
- `git diff --check` -> clean

## Summary by CodeRabbit

* **Chores**
  * CI now uses repository-local Meson build configuration for same-repo PRs, improving consistency for builds and checks; private Meson include paths are normalized more robustly to produce correct absolute include entries.
* **Tests**
  * Added a regression test that verifies normalization of both quoted and unquoted private include arguments, ensuring originals are replaced by normalized absolute paths.

### DIFF
--- a/.github/workflows/build_linux.yml
+++ b/.github/workflows/build_linux.yml
@@ -73,6 +73,12 @@ jobs:
           if [ -f "pr-code/CMakeLists.txt" ]; then
             cp pr-code/CMakeLists.txt trusted/
           fi
+          if [ "${{ github.event.pull_request.head.repo.full_name }}" = "${{ github.repository }}" ]; then
+            # Same-repository CI helper PRs need to exercise their own Meson setup code.
+            if [ -f "pr-code/ci/meson/build_config.py" ]; then
+              cp pr-code/ci/meson/build_config.py trusted/ci/meson/build_config.py
+            fi
+          fi
           # Change working directory to trusted for all subsequent steps
           cd trusted
         shell: bash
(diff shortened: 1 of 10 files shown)


zackees:
Addressed CodeRabbit's path-resolution test feedback in 0cf2ce5dad89cbc35ef8c2ff238a1a49c57cb0d0 and resolved the review thread. @coderabbitai review

PR #2419, opened 2026-04-28
fix(fl): address CodeRabbit follow-up findings from PR #2414

zackees:
## Summary

Follow-up to #2414, addressing the three CodeRabbit findings the prior commits in that PR deferred to a human reviewer.

- **`src/fl/channels/rx.cpp.hpp`** — Append `FL_NOEXCEPT` to all twelve `RxDevice::create<>` out-of-class definitions in the `FL_IS_ESP32`, `FL_IS_TEENSY_4X`, and `#else` fallback blocks so they match the `FL_NOEXCEPT`-qualified declarations in `rx.h`. Stub block was already correct. (C++17 requires matching exception specs between declarations and definitions of explicit template specializations — currently latent because `FL_NOEXCEPT` expands to nothing, but becomes a hard compile error the moment that macro is switched on for any platform.)
- **`src/fl/ui/button.h` + `button.cpp.hpp`** —
  - `clickedCount()` no longer ignores `mButtonInput`. The `Listener` now edge-detects rising transitions of `mButtonInput->clicked()`, increments a local counter, and `clickedCount()` returns `mImpl.clickedCount() + mListener.realButtonClickCount()`. Restores parity with `clicked()` / `isPressed()`, which already fall back to `mButtonInput`.
  - Move `mListener` to be the **last** private member so it is destroyed first and unregisters from `EngineEvents` before `mButtonInput` and the callback lists are torn down. Fixes a destruction-order hazard where `Listener::onBeginFrame()` could observe a half-destroyed `UIButton`.
- **`src/fl/sensors/ui_button_integration.cpp.hpp`** — Append `FL_NOEXCEPT` to the `UIButton::addRealButton` and `UIDropdown::addNextButton` out-of-class definitions to match their declarations.

## Test plan

- [x] `bash lint` — clean
- [x] `bash test --cpp` — 302/302 passed
- [ ] CI green on master after merge

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved real button click detection to ensure all physical button presses are properly registered and counted.

* **Chores**
  * Enhanced exception safety specifications across button and device creation functions.

### DIFF
--- a/src/fl/channels/rx.cpp.hpp
+++ b/src/fl/channels/rx.cpp.hpp
@@ -85,7 +85,7 @@ namespace fl {
 
 // RMT device specialization for ESP32
 template <>
-fl::shared_ptr<RxDevice> RxDevice::create<RxDeviceType::RMT>(int pin) {
+fl::shared_ptr<RxDevice> RxDevice::create<RxDeviceType::RMT>(int pin) FL_NOEXCEPT {
     auto device = RmtRxChannel::create(pin);
     if (!device) {
         return fl::make_shared<DummyRxDevice>("RMT RX channel creation failed");
@@ -95,7 +95,7 @@ fl::shared_ptr<RxDevice> RxDevice::create<RxDeviceType::RMT>(int pin) {
 
 // ISR device specialization for ESP32
 template <>
-fl::shared_ptr<RxDevice> RxDevice::create<RxDeviceType::ISR>(int pin) {
+fl::shared_ptr<RxDevice> RxDevice::create<RxDeviceType::ISR>(int pin) FL_NOEXCEPT {
     auto device = GpioIsrRx::create(pin);
     if (!device) {
         return fl::make_shared<DummyRxDevice>("GPIO ISR RX creation failed");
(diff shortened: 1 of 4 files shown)


zackees:
## Post-merge expert review

Reviewed this PR's four changes against the broader codebase:

| Change | Verdict |
|---|---|
| `rx.cpp.hpp` — append `FL_NOEXCEPT` to all 12 `RxDevice::create<>` defs | ✅ Fine. All defs match `rx.h` decls (lines 433-436) and the primary template at line 278. |
| `button.h` — move `mListener` to last private member | ✅ Fine. Init list `mImpl(name), mListener(this)` is in declaration order (no -Wreorder), and mListener now destructs first → unregisters from EngineEvents before mButtonInput / callback lists. |
| `button.h` — `removeClickedCallback` alias + per-list comment | ✅ Fine. |
| `ui_button_integration.cpp.hpp` — `FL_NOEXCEPT` on `addRealButton` / `addNextButton` defs | ✅ Fine. Match decls. `FL_NOEXCEPT` reaches the TU via `fl/sensors/button.h` → `fl/stl/noexcept.h`. |
| `button.cpp.hpp` — `clickedCount()` edge-detection on `mButtonInput->clicked()` | 🚨 **Bug**: see below. |

### 🚨 Bug found in the `clickedCount()` edge counter

The new edge counter in `UIButton::Listener::onBeginFrame`:

```cpp
const bool real_clicked = mOwner->mButtonInput->clicked();
if (real_clicked && !mRealButtonClickedLastFrame) {
    ++mRealButtonClickCount;
}
mRealButtonClickedLastFrame = real_clicked;
```

silently relies on `IButtonInput::clicked()` being **transient** (true for one frame after a press, false otherwise). But the only concrete impl, `fl::Button`, has a **sticky-true** `clicked()`: `mClickedThisFrame` is set to `true` on the first press transition and **never reset** anywhere (`grep mClickedThisFrame` shows two writes — initialization to `false` and one `= true`).

Result: `clickedCount()` on a `UIButton` with an attached real `Button` will increment **exactly once** per lifetime — the first press — and stay at 1 forever, regardless of how many times the user actually presses the button.

The same root cause silently breaks `UIDropdown::Listener::onBeginFrame` (`src/fl/ui/dropdown.cpp.hpp:29`, pre-existing): `mNextButton->clicked()` becomes sticky-true → `nextOption()` is called every frame after the first press → the dropdown auto-advances nonstop. This wasn't introduced here but is exposed by the same defective contract.

### Why CI didn't catch it

No host-side unit tests cover `Button` / `UIButton::clickedCount()` / `UIDropdown` polling. The bug manifests only on real hardware where a physical button is pressed across multiple frames.

### Fix

Filed as #2425 — patches `Button::Listener::onEndFrame` to write `mClickedThisFrame = (rising-edge of isPressed)` unconditionally each frame, restoring the transient semantic the field's own comment already documents (`// This is true if clicked this frame.`). That single change repairs both call sites simultaneously.

Issue #2421, opened 2026-04-30
[Tracking] ESP32-S3 timing regression + binary bloat from Channel API (from #2420)


zackees:
## Summary

Tracking issue for the timing-precision and binary-size regression on ESP32-S3 (and likely all ESP32 family + similar Channel-API platforms) reported in #2420 by @nomakewan.

This issue exists to coordinate investigation, organize fix proposals, and track shipping of the near-term mitigation.

## Reported symptoms (from #2420)

- **Timing drift**: A loop that should take 2500 ms now averages ~2600 ms with random spikes >3000 ms on master vs 3.10.3. Sketch is minimal: `FastLED.show()` every 5 ms via `millisDelay`, 35 WS2812B LEDs, NEOPIXEL pin only, `setMaxRefreshRate(0)`, `setCpuFrequencyMhz(160)`.
- **Binary bloat**: Same sketch grew from **344 KB → 783 KB** (+439 KB / 2.27x) between 3.10.3 and `master#ea5d8ec`.
- Bare `delay()` is *not* swapped to `fl::delay` on Arduino targets (gated to `FASTLED_STUB_IMPL` at `src/FastLED.h:1648`), so async pumping in `fl::delay` is not the cause.

### Timing jitter — `vTaskDelay(1)` floor in every `show()`

The new Channel API inserts two FreeRTOS-blocking waits per `FastLED.show()` call that did not exist in 3.10.3:

1. **Begin-frame wait**: `CFastLED::show()` -> `onBeginFrame()` (FastLED.cpp.hpp:233) fires `EngineEvents::onBeginFrame` -> `ChannelManager::onBeginFrame()` -> `waitForReady()` (manager.cpp.hpp:401-403).
2. **End-frame wait**: `onEndFrame()` (FastLED.cpp.hpp:280) -> `ChannelManager::onEndFrame()` -> `driver->show()` then `waitForReadyOrDraining()` (manager.cpp.hpp:405-415).

Both go through `ChannelManager::waitForCondition` (manager.cpp.hpp:326-336), which busy-loops:

```cpp
while (!condition()) {
    if (timeoutMs && (millis() - startTime >= timeoutMs)) return false;
    task::run(250, task::ExecFlags::SYSTEM);   // <- blocks >=1 FreeRTOS tick
}
```

[...]

Issue #2420, opened 2026-04-30
[Bug] Compounded timing inaccuracy on master compared to 3.10.3


nomakewan:
### Board / MCU

ESP32-S3

### LED Chipset

WS2812B / NeoPixel

### FastLED Version

master#ea5d8ec

### Arduino / PlatformIO version

PlatformIO 6.1.19

### Operating System

Windows

### Description

Compared to timing performance in `3.10.3`, timing performance in `master` is massively worse. This is even for incredibly simple operations such as writing an update to a single pixel in a loop that does nothing other than write to a new pixel on a specific delay. It appears that timing calls are significantly negatively affected by the current code in `master`.

Note that for timing we use the incredibly common library `millisDelay` which is a part of the PIO package `powerbroker2/SafeString`. With all previous versions of FastLED this has never been a problem, and allows for very tight millisecond-precision timing.

Using rudimentary measurements with a stopwatch, merely switching from `3.10.3` to `master` using the same demo code added up to a whopping 500ms of additional delay compared to the normal sequence. The normal sequence should take a total of 2500ms to run, and switching to `master` caused it to randomly exceed 3000ms. On average, rather than taking 2500ms to run, the same code would typically be around 2600ms. However, as mentioned, this timing drift appears to be highly random.

Additionally, the binary size _absolutely exploded_ by changing from `3.10.3` to `master`. With the exact same example code, `3.10.3` clocked in at 344 kilobytes while `master` clocks in at 783 kilobytes.

### Steps to reproduce

1. Target FastLED `master#ea5d8ec`.
2. Compile minimal sketch.
3. See that timing precision is significantly more unreliable than `3.10.3`.

### Minimal sketch

```cpp
#include <Arduino.h>
#include <millisDelay.h>
#include <FastLED.h>

uint8_t i_post_fade = 255;
#define LED_PIN 4
#define FAST_LED_UPDATE_MS 5
#define NUM_LEDS 35 // 35-LED NeoPixel ring
CRGB leds[NUM_LEDS];
uint8_t i_fast_led_delay = FAST_LED_UPDATE_MS;
millisDelay ms_fast_led;
millisDelay ms_delay_post;

void setup() {
  FastLED.addLeds<NEOPIXEL, LED_PIN>(leds, NUM_LEDS).setCorrection(TypicalLEDStrip);
  FastLED.setMaxRefreshRate(0);
  FastLED.show();
  setCpuFrequencyMhz(160);
  Serial.begin(115200);
[...]

PR #2423, opened 2026-05-01
feat(fl/stl): add FL_NO_UNWIND macros, apply to ESP32 unity headers

zackees:
## Summary

Adds `FL_NO_UNWIND` / `FL_NO_UNWIND_BEGIN` / `FL_NO_UNWIND_END` macros and applies the TU-level pair to both ESP32 unity-build headers. Targets the binary-size half of #2421 by suppressing GCC unwind-table emission for the FastLED ESP32 driver code, with a debug-aware gate so panic backtraces stay intact in development builds.

Closes part of #2421. Related: #2420.

## Why

Symbol-level analysis of @nomakewan's reproducer build (`master#ea5d8ec`, ESP32-S3, NEOPIXEL, 35 LEDs) showed `libFastLED.a` contributing **180 KB of `.eh_frame`** — half of the entire `.flash.rodata` section. A direct count of FDE entries showed:

- 16,773 total FDEs in the binary
- **13,272 (79.1%) are orphan FDEs** covering zero-byte ranges — left behind by GNU LD when `--gc-sections` strips dead virtual-dispatch code but fails to cull the corresponding `.eh_frame` entry
- Only 21% point to live code

Full breakdown: https://github.com/FastLED/FastLED/issues/2421#issuecomment-4356403093

`FL_NOEXCEPT` does not suppress this — `noexcept` is a language contract, `.eh_frame` is a codegen artifact controlled by `-fasynchronous-unwind-tables` (default ON in GCC, used by panic handlers / debuggers / profilers, not just C++ exceptions).

## What changed

- `src/fl/stl/compiler_control.h` (+89): three new macros following the same pattern as the existing `FL_DISABLE_WARNING_PUSH/POP` and `FL_BEGIN_OPTIMIZE_FOR_EXACT_TIMING` blocks
- `src/platforms/esp/32/_build.cpp.hpp` (+5): wrap entire unity TU with `FL_NO_UNWIND_BEGIN`/`_END`
- `src/platforms/esp/32/drivers/_build.cpp.hpp` (+5): same

## Gating logic — preserves backtraces by default

The macros are **no-ops (preserve `.eh_frame`) whenever** any of these are set:

**FastLED debug indicators:** `!RELEASE`, `FASTLED_TESTING`, `FASTLED_FORCE_DBG`, `FASTLED_DEBUG`, `FL_DEBUG`, `!NDEBUG`

**ESP32 sdkconfig (only when `FL_IS_ESP32` is defined):**
- `CONFIG_ESP_SYSTEM_PANIC_PRINT_BACKTRACE` (ESP-IDF default ON)
- `CONFIG_ESP_SYSTEM_PANIC_GDBSTUB`
- `CONFIG_ESP_DEBUG_OCDAWARE`
- `CONFIG_COMPILER_OPTIMIZATION_DEBUG` / `CONFIG_COMPILER_OPTIMIZATION_NONE`

**Arduino-ESP32 verbosity:** `CORE_DEBUG_LEVEL >= 3`

**User overrides (last word):** `FASTLED_FORCE_UNWIND_TABLES` (force-keep) / `FASTLED_FORCE_NO_UNWIND_TABLES` (force-strip)

This means the size win has to be **opted into** with `-DFASTLED_FORCE_NO_UNWIND_TABLES` in `build_flags`. The default behavior on the reporter's exact build (`CORE_DEBUG_LEVEL=3` + ESP-IDF default panic backtrace) is **no change** — backtraces are preserved, binary size is unchanged from current master. Users who want the ~180 KB win pay for it explicitly.

## Why opt-in is the right default

A wrapped function shows `???` in `esp32_exception_decoder` output. For a release build of a single-purpose installation that just blinks LEDs, that's an acceptable trade for 180 KB; for a development build or a customer-facing product where crash diagnostics matter, it isn't. Defaulting to "preserve" is the conservative choice.

## Why `compiler_control.h` and not a new `codegen.h`

The original spec proposed `src/fl/stl/codegen.h`, but `src/fl/stl/compiler_control.h` already hosts every other `_Pragma`-based macro pair in the codebase (`FL_DISABLE_WARNING_PUSH/POP`, `FL_BEGIN_OPTIMIZE_FOR_EXACT_TIMING/_END`, `FL_OPTIMIZATION_LEVEL_O3_BEGIN/_END`). Same shape, same conventions, same compiler-detection macros (`FL_IS_GCC`, `FL_GCC_VERSION`, `FL_STRINGIFY`) — so the natural home was here, not a parallel new file.

## Why just `optimize("no-unwind-tables")` and not also `no-asynchronous-unwind-tables`

`-fno-unwind-tables` is the strictly stronger GCC flag and implies `-fno-asynchronous-unwind-tables`. The redundant pair is a common belt-and-suspenders pattern in tutorials but adds nothing on GCC. One flag is enough. (Discussion: https://github.com/FastLED/FastLED/issues/2421#issuecomment-4357768687.)

## Why the GCC-only gate (and not clang)

`#pragma GCC optimize` is accepted by clang but emits an "unknown pragma" warning at minimum, and clang's per-function optimize attribute support is uneven (well-known caveats with LTO and template instantiation). Restricting to `FL_IS_GCC` keeps clang builds clean.

## Verification

- `bash lint` — clean
- `bash test --cpp` — 302/302 host tests pass (~107s)
- `bash compile esp32s3 --examples Blink` — succeeds (~437s), produces 810.81 KB firmware.bin (default mode, backtraces preserved — exactly as designed)
- Pragma compiles without warning under Xtensa GCC 13.2 from pioarduino's `toolchain-xtensa-esp-elf`

## Test plan

- [x] Lint clean
- [x] Host tests (302/302) pass
- [x] ESP32-S3 default-mode compile succeeds, binary unchanged from current master
- [ ] **Need community verification**: build with `-DFASTLED_FORCE_NO_UNWIND_TABLES` and confirm ~180 KB savings
- [ ] **Need community verification**: panic-decode a forced crash inside `Channel::showPixels` in default mode (backtrace should be intact) and in opt-in mode (frame should show `???`, downstream frames should still resolve)
- [ ] @nomakewan if you're up for it, retest your #2420 sketch with this branch + `-DFASTLED_FORCE_NO_UNWIND_TABLES` in `build_flags` and report the new binary size

## What this does NOT fix

This PR addresses ONE half of #2421 (the binary bloat). It does NOT fix:

- The per-`show()` `vTaskDelay(1)` timing regression (separate `FASTLED_TIGHT_TIMING` proposal in #2421)
- The unconditional driver compilation in `_build.cpp.hpp` (per-driver feature gates, ~125 KB more savings)
- The cross-TU references that pull in `fl.audio` / `fl.remote` / `fl.net` / `fl.fx` / `fl.gfx` for one symbol each (will be a follow-up issue)

Those land in subsequent PRs against #2421.

## Summary by CodeRabbit

* **Chores**
  * Added build-time controls to optionally suppress unwind table emission on embedded builds, reducing binary size.
  * Applied those controls across platform unity-builds so the setting consistently affects aggregated implementation files.
  * Backtrace/debug builds automatically preserve unwind information; unsupported toolchains remain unaffected and can no-op the controls.

* **Tests**
  * Added tests verifying the new unwind-control behavior and safe usage when the controls are no-ops.

### DIFF
--- a/src/fl/stl/compiler_control.h
+++ b/src/fl/stl/compiler_control.h
@@ -224,6 +224,133 @@
   #define FL_OPTIMIZATION_LEVEL_O0_END   /* nothing */
 #endif
 
+// =============================================================================
+// FL_NO_UNWIND family — suppress DWARF .eh_frame unwind-table emission
+// =============================================================================
+//
+// Public macros provided by this section:
+//
+//   FL_NO_UNWIND          per-function attribute (place before return type)
+//   FL_NO_UNWIND_BEGIN    start of TU/block-scoped pragma stripping
+//   FL_NO_UNWIND_END      paired with _BEGIN
+//   FL_NO_UNWIND_ACTIVE   query macro: 1 if stripping is active, else 0
+//
+// Why: ~180 KB savings on ESP32-S3 builds where 79% of FDE entries are
+// orphan FDEs left behind after --gc-sections strips dead virtual-dispatch
+// code (see issue #2421).
+//
+// SAFETY: A function compiled with stripped unwind tables cannot be unwound
+// through by the ESP32 panic handler — backtraces stop at that frame and
+// resume at the next frame with valid CFI. Apply only to hot dispatch or
+// driver code where the size win matters; do NOT apply to setup, loop, init,
+// or error-handling paths where a usable backtrace matters.
+//
+// Per-function form:
+//   FL_NO_UNWIND void hot_dispatch() noexcept { ... }
+//
+// TU-level form (preferred for unity-build .cpp.hpp wrappers):
+//   FL_NO_UNWIND_BEGIN
+//   #include "platforms/.../_build.cpp.hpp"
+//   FL_NO_UNWIND_END
+//
+// =============================================================================
+// User control hierarchy (highest priority first)
+// =============================================================================
+//
+//   1. Direct macro override.
+//      Pre-define any of FL_NO_UNWIND / FL_NO_UNWIND_BEGIN / FL_NO_UNWIND_END
+//      before including this header — yours will be respected, the others
+//      will be stubbed to no-op. Pre-define FL_NO_UNWIND_ACTIVE yourself if
+//      you want to assert what your override does.
+//
+//   2. Hard policy override (overrides everything below):
+//        #define FASTLED_FORCE_NO_UNWIND_TABLES 1   // always strip
+//        #define FASTLED_FORCE_UNWIND_TABLES    1   // always preserve
+//      FORCE_NO wins if both are accidentally set.
+//
+//   3. Debug auto-detect — preserve unwind tables when ANY of:
+//        !defined(RELEASE)
+//        !defined(NDEBUG)
+//        defined(FASTLED_TESTING) / FASTLED_FORCE_DBG / FASTLED_DEBUG / FL_DEBUG
+//        on ESP32: CONFIG_ESP_SYSTEM_PANIC_PRINT_BACKTRACE,
+//                  CONFIG_ESP_SYSTEM_PANIC_GDBSTUB,
+//                  CONFIG_ESP_DEBUG_OCDAWARE,
+//                  CONFIG_COMPILER_OPTIMIZATION_DEBUG,
+//                  CONFIG_COMPILER_OPTIMIZATION_NONE,
+//                  CORE_DEBUG_LEVEL >= 3
+//
+//   4. Default (release build, no diagnostic indicators) — strip.
+
+// -----------------------------------------------------------------------------
+// Stage 1: resolve FL_NO_UNWIND_ACTIVE policy (user can pre-define).
+// -----------------------------------------------------------------------------
+#ifndef FL_NO_UNWIND_ACTIVE
+  #if defined(FASTLED_FORCE_NO_UNWIND_TABLES) && FASTLED_FORCE_NO_UNWIND_TABLES
+    // Hard force ON wins over everything below.
+    #define FL_NO_UNWIND_ACTIVE 1
+  #elif defined(FASTLED_FORCE_UNWIND_TABLES) && FASTLED_FORCE_UNWIND_TABLES
+    // Hard force OFF wins over auto-detect.
+    #define FL_NO_UNWIND_ACTIVE 0
+  #elif !defined(RELEASE) \
+        || defined(FASTLED_TESTING) \
+        || defined(FASTLED_FORCE_DBG) \
+        || defined(FASTLED_DEBUG) \
+        || defined(FL_DEBUG) \
+        || !defined(NDEBUG) \
+        || (defined(FL_IS_ESP32) && ( \
+               (defined(CONFIG_ESP_SYSTEM_PANIC_PRINT_BACKTRACE) && CONFIG_ESP_SYSTEM_PANIC_PRINT_BACKTRACE) \
+            || (defined(CONFIG_ESP_SYSTEM_PANIC_GDBSTUB)         && CONFIG_ESP_SYSTEM_PANIC_GDBSTUB) \
+            || (defined(CONFIG_ESP_DEBUG_OCDAWARE)               && CONFIG_ESP_DEBUG_OCDAWARE) \
+            || (defined(CONFIG_COMPILER_OPTIMIZATION_DEBUG)      && CONFIG_COMPILER_OPTIMIZATION_DEBUG) \
+            || (defined(CONFIG_COMPILER_OPTIMIZATION_NONE)       && CONFIG_COMPILER_OPTIMIZATION_NONE) \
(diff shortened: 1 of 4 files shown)


nomakewan:
The new binary size for the example code from #2420 with `-D "FASTLED_FORCE_NO_UNWIND_TABLES"` in `build_flags` and targeting this branch is 770 KB if `-D "CORE_DEBUG_LEVEL=0"` and 784 KB if `-D "CORE_DEBUG_LEVEL=3"`.

zackees:
## Refactored macro structure for robustness + user override

Force-pushed (PR still one commit, `21a71a3b02`). Changes from prior version:

### What's new

1. **User can pre-define any of the macros directly.** Each `FL_NO_UNWIND` / `FL_NO_UNWIND_BEGIN` / `FL_NO_UNWIND_END` is now guarded with `#ifndef`, so a user can pre-define their own (e.g. a custom pragma combo) before including `fl/stl/compiler_control.h` and the header will respect it.

2. **New `FL_NO_UNWIND_ACTIVE` query macro.** Always defined to `0` or `1` after include. Lets calling code branch on whether stripping is actually happening, e.g. for log lines or conditional compilation. Also pre-overridable.

3. **Cleaner two-stage policy resolution:**
   - **Stage 1**: resolve `FL_NO_UNWIND_ACTIVE` (skipped if user pre-defined)
   - **Stage 2**: emit each macro independently with `#ifndef` guards (skipped if user pre-defined)
   
   Eliminates the previous coupled `_FL_BUILD_WANTS_BACKTRACE` sentinel, which used a reserved-identifier namespace (`_` + uppercase, technically reserved by the C++ standard).

4. **Documented user-control hierarchy** (highest priority first):
   1. Direct macro override (pre-define `FL_NO_UNWIND` / `FL_NO_UNWIND_BEGIN` / `FL_NO_UNWIND_END` / `FL_NO_UNWIND_ACTIVE`)
   2. Hard policy override: `FASTLED_FORCE_NO_UNWIND_TABLES=1` (always strip) or `FASTLED_FORCE_UNWIND_TABLES=1` (always preserve)
   3. Debug auto-detect (`!RELEASE`, `!NDEBUG`, FastLED debug macros, ESP32 panic-config, `CORE_DEBUG_LEVEL>=3`)
   4. Default: release build with no diagnostic indicators → strip

5. **Self-correcting `FL_NO_UNWIND_ACTIVE`**: if user requests stripping (Stage 1 sets it to 1) but we're not on GCC and so can't honor it (Stage 2 emits no-ops), the query macro is corrected back to 0 to accurately reflect what was emitted.

### Code shape

```cpp
// Stage 1: policy
#ifndef FL_NO_UNWIND_ACTIVE
  #if defined(FASTLED_FORCE_NO_UNWIND_TABLES) && FASTLED_FORCE_NO_UNWIND_TABLES
    #define FL_NO_UNWIND_ACTIVE 1
  #elif defined(FASTLED_FORCE_UNWIND_TABLES) && FASTLED_FORCE_UNWIND_TABLES
    #define FL_NO_UNWIND_ACTIVE 0
  #elif <debug auto-detect compound>
    #define FL_NO_UNWIND_ACTIVE 0
  #else
    #define FL_NO_UNWIND_ACTIVE 1
  #endif
#endif

// Stage 2: emission
#if FL_NO_UNWIND_ACTIVE && defined(FL_IS_GCC)
  #ifndef FL_NO_UNWIND       /* ... attribute ... */ #endif
  #ifndef FL_NO_UNWIND_BEGIN /* ... pragmas  ... */ #endif
  #ifndef FL_NO_UNWIND_END   /* ... pop      ... */ #endif
#else
  #ifndef FL_NO_UNWIND       /* no-op */ #endif
  // ...
  #if FL_NO_UNWIND_ACTIVE  // user wanted strip but we're non-GCC
    #undef FL_NO_UNWIND_ACTIVE
    #define FL_NO_UNWIND_ACTIVE 0
  #endif
#endif
```

### Tests added

`tests/fl/stl/compiler_control.cpp` gets a new `FL_NO_UNWIND family` test case covering:
- `FL_NO_UNWIND_ACTIVE` is exactly 0 or 1
- Per-function `FL_NO_UNWIND` attribute compiles + runs
- TU-level `FL_NO_UNWIND_BEGIN`/`_END` compiles + runs (sample function defined inside the block)
- Macros are always defined (even when no-op)
- Nested `FL_NO_UNWIND_BEGIN`/`_END` blocks compile

### Verification

- `bash lint --cpp` clean
- `bash test compiler_control` — 1/1 binary, all subcases pass
- `bash test --cpp` — 301/301 host test binaries pass

### Net diff vs prior PR version

```
src/fl/stl/compiler_control.h     | 172 ++++++++++++++++++++++++-------------
tests/fl/stl/compiler_control.cpp |  51 +++++++++++
```

Macros header grew (more comments + the cleaner two-stage structure); test coverage added for the new behaviors.

PR #2435, opened 2026-05-10
Feat/channels bus traits drivers

zackees:
## Summary by CodeRabbit

* **New Features**
  * Added a generic channel interface so listeners can handle different channel implementations.

* **Refactor**
  * Updated channel event callbacks to use the generic interface type for broader compatibility.

* **Documentation**
  * Updated examples and README snippets to show the new callback parameter types and minor log formatting.

* **Tests**
  * Adjusted tests to use the interface-based callbacks.

* **Bug Fixes**
  * Improved handling of keyboard interrupts during lock release/cleanup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FastLED/FastLED/pull/2435)

### DIFF
--- a/src/FastLED.h
+++ b/src/FastLED.h
@@ -636,7 +636,7 @@ public:
 	/// @return Reference to the ChannelEvents singleton
 	/// @note Use to register callbacks for channel lifecycle events
 	/// @code
-	/// int id = FastLED.channelEvents().onChannelCreated.add([](const fl::Channel& ch) { ... });
+	/// int id = FastLED.channelEvents().onChannelCreated.add([](const fl::IChannel& ch) { ... });
 	/// @endcode
 	static fl::ChannelEvents& channelEvents();
 
--- a/src/fl/channels/README.md
+++ b/src/fl/channels/README.md
@@ -175,13 +175,13 @@ void setup() {
     auto& events = FastLED.channelEvents();
 
     // Called when channel is created
-    events.onChannelCreated.add([](const fl::Channel& ch) {
+    events.onChannelCreated.add([](const fl::IChannel& ch) {
         Serial.printf("Channel created: %s\n", ch.name().c_str());
     });
 
     // Called when channel data is enqueued to driver
-    events.onChannelEnqueued.add([](const fl::Channel& ch, const fl::string& driver) {
-        Serial.printf("%s → %s\n", ch.name().c_str(), driver.c_str());
+    events.onChannelEnqueued.add([](const fl::IChannel& ch, const fl::string& driver) {
+        Serial.printf("%s -> %s\n", ch.name().c_str(), driver.c_str());
     });
 
     // Create channel (triggers onChannelCreated)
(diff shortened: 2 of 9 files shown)


zackees:
## Squashed and addressed CodeRabbit feedback

Force-pushed a single squashed commit replacing the prior 2-commit history.

### Addressed

**`test.py` — KeyboardInterrupt handling in watchdog cleanup** (Major / Poor tradeoff)
You're right that re-raising `KeyboardInterrupt` inside `_release_held_build_locks()` defeats the watchdog's `os._exit()`. Removed the `raise` from both call sites:
- `_release_held_build_locks()` itself (both inner and outer `except KeyboardInterrupt`).
- `deadman_timer()`'s call-site wrapper.

The function now satisfies the KBI001 lint rule (`handle_keyboard_interrupt(ki)` called) but no longer propagates, so the unconditional `os._exit(3)` afterward always runs. Added explanatory comments at both sites.

### Skipped with rationale

**`uart/bus_traits.h` — `fl::make_shared` allocation failure** (Critical)
This is a project-wide pattern, not new in this PR. The same `fl::make_shared<UartPeripheralEsp>()` + `fl::make_shared<ChannelEngineUART>(peripheral)` sequence (no null check) already exists in `src/platforms/esp/32/drivers/channel_manager_esp32.cpp.hpp:233-235` (the file my BusTraits is mirroring). Every other driver factory in the codebase has the same shape.

Fixing this requires either:
1. A project-wide `fl::aligned_alloc` change to abort/assert on failure, OR
2. Refactoring every singleton factory to return `fl::optional<fl::shared_ptr<T>>`.

Both are out of scope for a Phase 2/3a additive PR; the risk profile of my code is identical to the surrounding codebase. Worth a follow-up issue if not already tracked.

**`#pragma once` -> FastLED compiler control macros** (Minor x4: rmt_4, rmt_5, i2s_spi, ichannel.h)
The cited rule appears to be a hallucination. I checked:
- `src/fl/stl/compiler_control.h:1` itself uses `#pragma once`.
- Every existing channels header uses `#pragma once` (`bus.h`, `bus_traits.h`, `channel.h`, `channel_events.h`, `chipset_helpers.h`, `config.h`, `data.h`, `driver.h`, `id_tracker.h`, ...).
- No `FASTLED_BEGIN_HEADER` / `FL_HEADER_BEGIN` / `FL_COMPILER_CONTROL_HEADER` macros exist in `compiler_control.h`.

`#pragma once` is the established codebase convention. Changing 4 of my new headers to a non-existent macro pair would introduce inconsistency, not fix it.

### Verification after squash

- `bash lint` clean.
- `bash test --cpp` passes 303/303.

zackees:
@zackees teensy41 failure is pre-existing on master — same PROGMEM section type conflict (`'RainbowColors_p' causes a section type conflict with 'fl::sinCosPairedLut'`) failing on every master CI run since at least commit ebb48e610f (5+ commits ago). Unrelated to IChannel changes here.

Filed as follow-up PR per your instruction: #2436. The fix special-cases `FL_PROGMEM` to expand to nothing on Teensy 4 (`__IMXRT1062__`) — Teensy 4 has a unified address space and the linker script places `.rodata` in flash automatically, so the section attribute is unnecessary and is the sole cause of the conflict.

Issue #2493, opened 2026-05-19
ESP32-P4 PARLIO: FastLED.show() appears to block for ~16.7 ms before WS2812 data burst begins


4wheeljive:
### Board / MCU

Other

### LED Chipset

WS2812B / NeoPixel

### FastLED Version

0a4ff478a3755cecaadccbb7b56a8c6a0aa4e0fe

### Arduino / PlatformIO version

pioarduino 1.4.4

### Operating System

Windows

### Description

Report authored by ChatGPT

## Summary

I’m using the FastLED ESP32-P4 PARLIO driver with WS2812B output and seeing behavior that suggests `FastLED.show()` is blocking for a fixed ~16.7 ms period, even though the actual WS2812 data burst for 256 LEDs is only ~7.6 ms, as expected.

The observed behavior looks like a fixed ~60 Hz wait / pacing / synchronization interval inside `FastLED.show()`, rather than time spent actively transmitting LED data.

This is significant because the `FastLED.show()` stage appears to block the main animation loop for ~16.7 ms regardless of whether I add a delay after `show()`, which effectively caps or strongly limits the overall frame rate.

---

## Hardware / setup

- MCU: ESP32-P4
- LED type: WS2812B
- Driver: FastLED PARLIO driver
- LED arrangement: 12 parallel strips of 256 LEDs each
- Scope: FNIRSI DPOS350P
- CH1: profiling GPIO pin
- CH2: LED data pin for one strip
- Grounds connected to system ground
- FastLED commit: 3.10.3+sha.0a4ff47
- Board: EPS32-P4

---

## Expected behavior

For a 256-pixel WS2812B strip:

```text
256 LEDs × 24 bits × ~1.25 µs/bit ≈ 7680 µs
```

So the actual data burst per strip should be approximately:

```text
~7.6–7.8 ms
```

With 12 strips driven in parallel, the expected data push time should still be roughly the same as for one 256-LED strip, assuming the PARLIO output is truly parallel.

I would therefore expect `FastLED.show()` to either:

1. block for approximately the actual data transmission time, around ~7.6–7.8 ms, or
2. queue an asynchronous transfer and return relatively quickly, depending on intended driver design.

Instead, I’m observing a consistent ~16.7 ms `FastLED.show()` duration.

---

## Profiling code

I wrapped `FastLED.show()` with both my software profiler and a GPIO pin:

```cpp
PROFILE_START("led_show");

digitalWrite(PROFILE_PIN, HIGH);
[...]

Issue #2254, opened 2026-04-12
ESP32-S3: I2S driver starves websockets, SPI output incorrect, Audio driver FPS impact


zackees:
## Summary

Hoisted from [#2167 (comment)](https://github.com/FastLED/FastLED/issues/2167#issuecomment-4229140580) by @ewowi to track remaining driver issues separately from the Channels API discussion.

Testing on **ESP32-S3** with **latest master** reveals three distinct issues across the LED drivers when used via the Channels API:

## Issue 1: I2S driver starves websockets (~1 FPS)

**Status:** Regression — this was previously fixed (delayMicroseconds busy-wait fix, Mar 9) but has resurfaced.

- When using I2S as the channel driver, websocket throughput drops to ~1 FPS
- **LEDs show the correct output** — the rendering itself is fine
- The I2S driver appears to be monopolizing the RTOS task, starving the network task layer
- This affects boards smaller than ESP32-P4 (tested on ESP32-S3, previously also on ESP32-D0)
- RMT does **not** exhibit this problem

**Root cause hypothesis** (from @zackees): The network stack has two layers — ISR layer and task layer. The task layer is what's getting starved. A `vTaskDelay(1)` or `delay(1)` between frames in the main loop may alleviate it, suggesting the I2S driver needs cooperative yield points.

## Issue 2: SPI driver — incorrect LED output

**Status:** Partially fixed in #2251 (buffer growth on channel reuse), but still not rendering correctly.

- SPI driver does **not** slow down websockets (good)
- However, LEDs are **not showing the right output** on ESP32-S3
- Video evidence: https://github.com/user-attachments/assets/afe42d51-a635-42a4-a838-bca5b56bfdf9

This is a separate rendering bug from the buffer issue fixed in #2251.

## Issue 3: FastLED Audio driver reduces FPS (100 → 60)

**Status:** Ongoing side-effect.

- With the FastLED Audio driver active, FPS drops from ~100 to ~60
- @ewowi asks: is the audio driver running on the same RTOS task as the LED driver code? If so, the slowdown is expected, and running audio on a separate task would resolve it.

## Driver behavior matrix (ESP32-S3, latest master)

| Driver | Websocket speed | LED output | Notes |
|--------|----------------|------------|-------|
| **RMT** | Normal | Correct | Works as expected |
| **SPI** | Normal | **Incorrect** | Rendering bug persists after #2251 |
| **I2S** | **~1 FPS** | Correct | Task starvation regression |

## Environment

[...]

PR #2497, opened 2026-05-22
perf(esp32): smart yield gate for driver polling loops

zackees:
## Summary

Driver polling loops on ESP32 (currently just one: `I2sLcdCamPeripheralEsp::waitTransmitDone`) previously called `vTaskDelay(1)` unconditionally so lower-priority network tasks (lwIP / WiFi / BT) could drain. That costs one full FreeRTOS tick (~1 ms at the default 1 kHz tick rate) per iteration even on headless / no-WiFi builds where no network task exists to yield to.

This PR introduces `fl::platforms::esp_smart_yield()` — an inline helper that gates the deep yield on the existing `ICoroutineRuntime::needsDeepYield()` mechanism. Two layers:

- **Link-time**: weak symbols on `esp_wifi_get_mode` and `esp_bt_controller_get_status`. If neither component is linked, the deep-yield branch is dead code.
- **Runtime**: when symbols are linked, the active-mode check is cached at 1 Hz inside `needsDeepYield()`. Per-call cost is a singleton lookup + one bool compare.

When the gate fires, we use **`vTaskDelay(1)` — exactly one FreeRTOS tick**, the smallest interval that releases the caller from the Ready set so the scheduler can dispatch lower-priority tasks. Sub-tick requests degenerate to `taskYIELD()` under FreeRTOS, which is what the deep-yield path exists to avoid.

## Changes

- `src/platforms/esp/32/coroutine_esp32.impl.hpp`: extend the existing `isWiFiActive()` → `isNetworkActive()` so `needsDeepYield()` now ORs WiFi-mode-active and BT-controller-not-idle. New weak extern for `esp_bt_controller_get_status` under `SOC_BT_SUPPORTED`. Behavior preserved on WiFi-only configurations.
- `src/platforms/esp/32/yield_gate.hpp` (new): the inline `fl::platforms::esp_smart_yield()` helper, with the two-layer-gate rationale and the one-tick yield magnitude reasoning in the header doc.
- `src/platforms/esp/32/drivers/i2s/i2s_lcd_cam_peripheral_esp.cpp.hpp`: the sole FastLED-side `vTaskDelay(1)` poll loop now calls `esp_smart_yield()`.

## Scope notes

- The full `grep` for FastLED-side `vTaskDelay(N)` polling loops with small `N` across `src/platforms/esp/32/drivers/**` returned exactly one site (the I2S LCD_CAM `waitTransmitDone` loop). PARLIO and RMT5 delegate to ESP-IDF's `parlio_tx_unit_wait_all_done` / `rmt_tx_wait_all_done`, which are semaphore-backed inside IDF — no FastLED-side polling to gate.
- The headline ESP32-P4 PARLIO latency reported in #2493 is **not** addressed here; that needs hardware-measured investigation into the IDF wait path. The smart-yield helper is reusable if future drivers introduce poll loops.

## Refs

- Issue [#2493](https://github.com/FastLED/FastLED/issues/2493) (ESP32-P4 PARLIO `FastLED.show()` blocking analysis) — motivating context.
- Issue [#2254](https://github.com/FastLED/FastLED/issues/2254) (original WiFi-starvation root-cause of the round-up-to-one-tick rule).

## Test plan

- [x] `bash compile esp32s3 --examples Blink` — pass
- [x] `bash compile esp32dev --examples Blink` — pass
- [x] `bash compile esp32p4 --examples Blink` — pass
- [x] `bash lint` — clean
- [x] `bash test --cpp` — 304/304 (cached; host tests don't compile ESP-only paths)
- [ ] Hardware validation on ESP32-S3 with I2S LCD_CAM driver (needs hardware — no behavioral change expected on no-network builds; ms-level frame budget improvement on builds without WiFi/BT linked)
- [ ] CI green on branch

## Summary by CodeRabbit

* **Refactor**
  * Improved ESP32 task scheduling to detect WiFi and Bluetooth activity and use more appropriate yielding during I2S operations.
* **New Features**
  * Added a network-aware yield helper that chooses between light and deep yields based on active network stacks.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FastLED/FastLED/pull/2497?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/src/platforms/esp/32/coroutine_esp32.impl.hpp
+++ b/src/platforms/esp/32/coroutine_esp32.impl.hpp
@@ -125,7 +125,10 @@ private:
 #endif
 #if SOC_BT_SUPPORTED
         if (esp_bt_controller_get_status != nullptr) {
-            if (esp_bt_controller_get_status() != ESP_BT_CONTROLLER_STATUS_IDLE) {
+            // Match NetworkDetector::isBluetoothActive() — controller is active
+            // only when ENABLED (not merely INITED), so we don't pay the deep
+            // yield for a bt-stack-init'd-but-radio-off configuration.
+            if (esp_bt_controller_get_status() == ESP_BT_CONTROLLER_STATUS_ENABLED) {
                 return true;
             }
         }
--- a/src/platforms/esp/32/coroutine_esp32.impl.hpp
+++ b/src/platforms/esp/32/coroutine_esp32.impl.hpp
@@ -48,6 +48,17 @@ FL_EXTERN_C_BEGIN
 // Weak symbol: resolves to nullptr when WiFi component is not linked
 FL_LINK_WEAK esp_err_t esp_wifi_get_mode(wifi_mode_t *mode);
 #endif
+#ifndef SOC_BT_SUPPORTED
+#define SOC_BT_SUPPORTED 0
+#endif
+#if SOC_BT_SUPPORTED
+#include "esp_bt.h"
+// Weak symbol: resolves to nullptr when BT controller component is not linked.
+// Returning anything other than ESP_BT_CONTROLLER_STATUS_IDLE indicates the
+// controller is initialized/enabled and may have ready BT tasks competing
+// for CPU.
+FL_LINK_WEAK esp_bt_controller_status_t esp_bt_controller_get_status(void);
+#endif
 FL_EXTERN_C_END
 
 namespace fl {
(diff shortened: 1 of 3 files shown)


zackees:
## Adversarial-review follow-up — three findings fixed in f4c7a15220

I ran an embedded-systems sub-agent against this PR with a hostile-review brief. It found three real issues; all fixed in the latest push.

### CRITICAL — namespace bug at `yield_gate.hpp:61` (would have broken ESP32-S3 build)

Original code called `fl::ICoroutineRuntime::instance()`. The class is actually declared in `namespace fl::platforms` (`src/platforms/coroutine_runtime.h:13-14,26`), so the qualifier failed name lookup. Every other call site in the codebase uses `fl::platforms::ICoroutineRuntime::instance()` (`await.h:48`, `executor.cpp.hpp:137,148`, `js_fetch.cpp.hpp:112,132`) — corrected to match.

**Important correction to my earlier "compile passed" claim:** the local `bash compile esp32s3 / esp32dev / esp32p4` runs all exited 0, but reading the actual output revealed the underlying `fbuild` had failed early on an Arduino-ESP32 framework cache issue (`WString.h` / `soc/soc_caps.h` missing, `daemon error: stream error`) — never reached FastLED code. The exit code was being swallowed somewhere in the wrapper. **None of my local compile checks were valid.** Relying on this PR's CI for actual compile validation now.

### HIGH — "exactly one FreeRTOS tick" doc claim was wrong

`vTaskDelay(1)` blocks for **between 1 and 2 tick periods** depending on phase relative to the next tick interrupt — *at least* one tick, not *exactly* one tick. Per the [FreeRTOS vTaskDelay docs](https://www.freertos.org/Documentation/02-Kernel/04-API-references/02-Task-control/01-vTaskDelay). Doc comments in `yield_gate.hpp` updated; the behavioral claim that motivates the helper (`vTaskDelay` blocks, `taskYIELD` doesn't) is unchanged.

### MEDIUM — BT controller status check was looser than the codebase's own existing check

I used `esp_bt_controller_get_status() != ESP_BT_CONTROLLER_STATUS_IDLE`. The existing `NetworkDetector::isBluetoothActive()` at `network_detector.cpp.hpp:247` uses `== ESP_BT_CONTROLLER_STATUS_ENABLED` (strict) and explicitly comments _"Controller is active only when enabled (not just initialized)"_. The `INITED`-but-not-`ENABLED` state means the BT stack is initialized but the radio is off — no scheduled BT tasks to yield to. My check would have over-triggered the deep yield in that state. Aligned with the existing strict check.

### Verified claims (per the agent)

- PARLIO/RMT5 do indeed delegate to IDF's `xQueueReceive`-backed waits — scope claim correct.
- `FL_LINK_WEAK` is `__attribute__((weak))` (`fl/stl/compiler_control.h:282`); the redeclare-as-weak pattern after the IDF `#include` is established (used by `NetworkDetector` for both `esp_wifi_get_mode` and `esp_bt_controller_get_status`).

### Items the agent flagged that I'm not changing (with reasoning)

- **Dual-core data race on `s_lastCheck` / `s_cached` static locals**: word-sized aligned reads/writes are atomic on Xtensa/RISC-V; worst case is one extra `isNetworkActive()` probe. Matches the existing single-core idiom in the same function pre-PR. Not promoting to `fl::atomic` without measured cause.
- **`WIFI_MODE_NULL` after `esp_wifi_stop()` without deinit**: false-positive for one tick per yield. Matches existing `NetworkDetector::isWiFiActive()` behavior — fixing it would diverge from the codebase's pattern.
- **First call after boot pays full probe**: `s_lastCheck = 0` so the first call always updates the cache. Documented honestly in the updated header comment.

zackees:
Closing per author request. Branch and PR superseded; reopen if needed.

PR #2567, opened 2026-05-30
fix(ci): skip --no-undefined on macOS (ld64.lld rejects GNU flag)

### LINKED ISSUE #2566: macOS CI broken: ld64.lld rejects -Wl,--no-undefined in quick/release modes
## Symptom

Every macOS CI job (e.g. `unit tests macos`, `example tests macos`) on `master` is failing during the link step:

```
ld64.lld: error: unknown argument '--no-undefined'
clang++: error: linker command failed with exit code 1
```

Failing run example: https://github.com/FastLED/FastLED/actions/runs/26676862991

## Root cause

`ci/meson/native/meson.build` (around the `runner_link_args` block, lines ~317-355) unconditionally appends `-Wl,--no-undefined` in `quick` and `release` modes for any non-Windows platform:

```meson
# macOS and Linux: dynamic linking (glibc doesn't support static + dlopen)
runner_link_args = core_link_args + deploy_link_args + [
  '-Wl,--no-undefined',  # Catch missing libraries early
  '-pthread',
]
```

Because we also pass `-fuse-ld=lld`, on macOS this resolves to `ld64.lld`. Apple's `ld64` (and `ld64.lld`) do not implement the GNU-style `--no-undefined` flag; the macOS-equivalent is `-undefined error`, which is the default behavior anyway.

So the flag is both:
1. Unrecognized by `ld64.lld` → hard link failure.
2. Redundant on macOS → default is already "error on undefined".

## Affected jobs

Every macOS workflow on `master`, including but not limited to:
- `unit tests macos`
- `example tests macos`
- `attiny*` / `nrf*` / `rp2350*` board builds that run on macOS runners

## Proposed fix

Split the non-Windows branch by platform and skip `--no-undefined` on Darwin:

```meson
is_darwin = build_machine.system() == 'darwin'
...
if is_darwin
  # macOS: ld64.lld doesn't accept --no-undefined; default is already error-on-undefined
  runner_link_args = core_link_args + deploy_link_args + ['-pthread']
else
  # Linux: GNU/LLD style
  runner_link_args = core_link_args + deploy_link_args + ['-Wl,--no-undefined', '-pthread']
endif
```

Same treatment for the parallel `release` branch.

## Repro

```
bash compile examples --mode quick   # on macOS host
```

or just look at any recent macOS CI run on `master`.

---
Filed automatically while triaging broken builds; intent is to fix locally and open a PR.

zackees:
## Summary

- Every macOS CI job on `master` (`unit tests macos`, `example tests macos`, and macOS-host board builds) is currently failing the runner link step because `ci/meson/native/meson.build` unconditionally passes `-Wl,--no-undefined` in quick/release modes for non-Windows platforms.
- We use `-fuse-ld=lld`, which on macOS resolves to `ld64.lld`. Apple's `ld64`/`ld64.lld` does not accept the GNU-style `--no-undefined` flag.
- This PR adds an `is_darwin` check and centralizes the flag in a single `no_undefined_link_args` list (empty on macOS, `['-Wl,--no-undefined']` everywhere else), then references it from each `runner_link_args` branch.
- macOS's default linker behavior is already "error on undefined symbols", so dropping the flag preserves the safety net it was providing.

Fixes #2566

## Test plan

- [x] `bash lint` (Windows host) — clean, including `meson_linting`.
- [ ] Wait for CI to confirm macOS jobs link successfully.
- [ ] Confirm Linux and Windows runner links are unchanged (flag still emitted there).

> Note: I could not run `bash test --cpp` on this Windows box to completion because of pre-existing local cache pollution from a sibling FastLED checkout (`fastled10`) leaking through `zccache` into the PCH for example DLLs — unrelated to this change, which only touches link args.

## Summary by CodeRabbit

* **Chores**
  * Centralized linker behavior in the build system to improve cross-platform consistency.
  * Avoided applying an incompatible macOS linker flag while keeping debug builds unchanged.
  * Ensured quick and release builds apply consistent linker rules across Windows, macOS, and Linux for more reliable builds.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FastLED/FastLED/pull/2567?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/ci/meson/native/meson.build
+++ b/ci/meson/native/meson.build
@@ -191,6 +191,16 @@ example_compile_args = base_compile_args
 is_windows = build_machine.system() == 'windows'
 is_macos = build_machine.system() == 'darwin'
 
+# macOS uses ld64.lld which does not accept GNU-style --no-undefined.
+# On macOS the linker errors on undefined symbols by default, so we just
+# omit the flag there. Windows and Linux LLD accept it as expected.
+# See FastLED #2612 and PR #2567.
+if is_macos
+  no_undefined_link_args = []
+else
+  no_undefined_link_args = ['-Wl,--no-undefined']
+endif
+
 # ============================================================================
 # Detect if clang-tool-chain Python wrapper is being used
 # ============================================================================
@@ -303,10 +313,11 @@ endif
 # In release mode: static linking on Windows, dynamic on Unix
 #
 # NOTE: -lunwind is added conditionally below after checking availability.
-# NOTE: -Wl,--no-undefined is added in quick/release modes to catch missing
-# libraries early. It's NOT added in debug mode because sanitizers (ASan/UBSan)
-# have symbols resolved at runtime, not link time.
-# See: https://github.com/google/sanitizers/issues/380
+# NOTE: -Wl,--no-undefined (via no_undefined_link_args) is added in
+# quick/release modes to catch missing libraries early. It's NOT added in:
+#   - debug mode: sanitizers (ASan/UBSan) resolve symbols at runtime
+#     (https://github.com/google/sanitizers/issues/380)
+#   - macOS: ld64.lld doesn't accept the flag; error-on-undefined is default
 if build_mode == 'debug'
   # Debug mode: dynamic linking required for sanitizers
   # No --no-undefined: sanitizers resolve symbols at runtime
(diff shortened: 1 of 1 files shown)


zackees:
Upstream root-cause fix: zackees/clang-tool-chain#43 (1.5.9). The legacy `clang-tool-chain-cpp` wrapper stripped `--no-undefined` on macOS via `_translate_linker_flags_for_macos_lld`; the new `clang-tool-chain-clang-cpp` (Phase 4/6) entry point execs through `zccache_shim.exec_via_zccache` which lost that translation.

When 1.5.9 lands on PyPI, we can:
1. Bump `clang-tool-chain>=1.5.9` in pyproject.toml.
2. Revert the macOS skip in this PR.

The workaround here remains correct as a belt-and-braces defense for older clang-tool-chain installs, but the duplicate-strip is harmless once 1.5.9 is the floor (Darwin gets an empty list either way).

PR #2613, opened 2026-05-30
ci(meson): skip --allow-shlib-undefined on macOS (fixes #2612)

### LINKED ISSUE #2612: unit tests macos: ld64.lld rejects --allow-shlib-undefined (broken on every master push)
﻿## Symptom

`unit tests macos` has been red on **every** master push for at least the last 24 hours (100% failure rate across 20+ consecutive runs). Most recent: https://github.com/FastLED/FastLED/actions/runs/26694216064

Two distinct macOS-specific failures fire at link time:

```
ld64.lld: error: unknown argument '--allow-shlib-undefined'
clang++: error: linker command failed with exit code 1
ninja: build stopped: subcommand failed.
```

Origin: `ci/meson/native/meson.build` lines 267 and 286 unconditionally add `-Wl,--allow-shlib-undefined` to every non-Windows debug link:

```python
else
  dll_link_args += ['-rdynamic']
  ...
  if build_mode == 'debug'
    dll_link_args += ['-Wl,--allow-shlib-undefined']
  endif
endif
...
else
  if build_mode == 'debug'
    fastled_shared_link_args += ['-Wl,--allow-shlib-undefined']
  endif
endif
```

## Root cause

`--allow-shlib-undefined` is an **ELF/GNU-ld flag** (also supported by LLD's ELF frontend on Linux). macOS uses **Mach-O** and `ld64.lld` does not implement this flag — it errors out. The Mach-O equivalent semantic ("don't fail the link on undefined symbols, defer to runtime resolution") is `-undefined dynamic_lookup`, not `--allow-shlib-undefined`.

The comment at lines 264-265 even acknowledges the situation:

> LLD by default allows undefined symbols in shared libraries, but we add the flag explicitly on Unix for compatibility with other linkers.

But "Unix" was the wrong gate — macOS is Unix but its linker is fundamentally different from Linux's. The flag is needed on Linux + GNU-ld and harmless on Linux + LLD; it is broken on macOS + ld64.lld.

## Proposed fix

Narrow the gate from `not is_windows` to `is_linux` only. Two-line change against the existing `is_windows = build_machine.system() == 'windows'` idiom: add `is_macos = build_machine.system() == 'darwin'` and guard both `--allow-shlib-undefined` adds with `if build_mode == 'debug' and not is_macos`.

## Secondary issue surfaced by the same log

Before the link failure, the macOS log also shows a `clang-tool-chain` setup error compiling `launcher_emcc.cpp`:

```
ctc_common.h:21:10: fatal error: 'cctype' file not found
```

That is a CTC-side issue (clang-tool-chain's macOS native-tool compile is missing the C++ stdlib include path / `-stdlib=libc++`). It is **independent** of the meson `--allow-shlib-undefined` bug — the meson build proceeds past it (those errors aren't fatal at the CTC layer, just the link step is) and then dies at the link. The CTC fix belongs in `zackees/clang-tool-chain`, not here.

This issue tracks only the meson `--allow-shlib-undefined` problem. CTC `cctype` will be filed separately.

## Test plan

- [ ] After the fix, `unit tests macos` workflow goes green on the next master push.
- [ ] `unit tests linux` continues to pass (the flag is still emitted there).
- [ ] `unit tests windows` continues to pass (already gated on `not is_windows`).

zackees:
﻿## Summary

`unit tests macos` has been red on **every** master push for at least the last 24 hours (100% failure rate across 20+ consecutive runs) with:

```
ld64.lld: error: unknown argument '--allow-shlib-undefined'
clang++: error: linker command failed with exit code 1
```

## Root cause

`ci/meson/native/meson.build` adds `-Wl,--allow-shlib-undefined` to every non-Windows debug link (lines 267, 286). The flag is an **ELF/GNU-ld** flag (also supported by LLD-ELF on Linux). macOS uses **Mach-O** and `ld64.lld` does not implement it — it errors out. The Mach-O equivalent ("don't fail the link on undefined symbols, defer to runtime resolution") is `-undefined dynamic_lookup`, but Mach-O resolves undefined symbols at runtime by default so it isn't needed.

The original comment even flagged the assumption — *"we add the flag explicitly on Unix for compatibility with other linkers"* — but "Unix" was the wrong gate. macOS is Unix but its linker is fundamentally different from Linux's.

## Fix

Narrow the gate from `not is_windows` to `not is_windows and not is_macos`. Two-line change against the existing platform-detect idiom:

```diff
 is_windows = build_machine.system() == 'windows'
+is_macos   = build_machine.system() == 'darwin'
 ...
-  if build_mode == 'debug'
+  if build_mode == 'debug' and not is_macos
     dll_link_args += ['-Wl,--allow-shlib-undefined']
   endif
```

Same gate applied to `fastled_shared_link_args`. Linux continues to emit the flag (LLD-ELF accepts it, GNU-ld needs it). Windows already gated. macOS now skipped.

## Notes on the other macOS failure in the same log

The CI log also shows a `'cctype' file not found` error compiling `launcher_emcc.cpp` during clang-tool-chain native-tool install. That is a CTC-side bug (CTC's darwin install branch doesn't set `-isysroot` / `-stdlib=libc++`) and is fixed in zackees/clang-tool-chain#44 — independent of this PR.

This PR fixes only the meson `--allow-shlib-undefined` issue. Together with zackees/clang-tool-chain#44, `unit tests macos` should turn green.

## Test plan

- [x] Local Windows: `bash test --cpp fl_gfx_rgbww` — passes (`is_macos == false` so flag still emitted on non-mac).
- [ ] CI `unit tests macos` — should turn green once this lands AND zackees/clang-tool-chain#44 ships in the pinned CTC version.
- [ ] CI `unit tests linux` — should remain green (flag still emitted on Linux).
- [ ] CI `unit tests windows` — should remain green (already gated).

Fixes #2612

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed macOS build compatibility by conditionally omitting linker flags that are incompatible with Mach-O. Debug and runner builds now avoid applying ELF-only options on macOS, restoring successful macOS builds without impacting other platforms.

[Review Change Stack](https://app.coderabbit.ai/change-stack/FastLED/FastLED/pull/2613?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/ci/meson/native/meson.build
+++ b/ci/meson/native/meson.build
@@ -334,10 +334,13 @@ elif build_mode == 'quick'
   else
     # macOS and Linux: dynamic linking (glibc doesn't support static + dlopen)
     # Note: -lunwind is added conditionally below after checking availability
-    runner_link_args = core_link_args + deploy_link_args + [
-      '-Wl,--no-undefined',  # Catch missing libraries early
-      '-pthread',
-    ]
+    # macOS ld64.lld rejects `--no-undefined` (ELF-only); Mach-O defaults to
+    # erroring on undefined symbols anyway, so the flag is redundant there.
+    # See FastLED #2612.
+    runner_link_args = core_link_args + deploy_link_args + ['-pthread']
+    if not is_macos
+      runner_link_args += ['-Wl,--no-undefined']  # Catch missing libraries early
+    endif
   endif
 else
   # Release mode: static linking on Windows, dynamic on Unix
@@ -355,10 +358,11 @@ else
   else
     # macOS and Linux: dynamic linking (glibc doesn't support static + dlopen)
     # Note: -lunwind is added conditionally below after checking availability
-    runner_link_args = core_link_args + deploy_link_args + [
-      '-Wl,--no-undefined',  # Catch missing libraries early
-      '-pthread',
-    ]
+    # macOS ld64.lld rejects `--no-undefined` (ELF-only); see comment above.
+    runner_link_args = core_link_args + deploy_link_args + ['-pthread']
+    if not is_macos
+      runner_link_args += ['-Wl,--no-undefined']  # Catch missing libraries early
+    endif
   endif
 endif
 
--- a/ci/meson/native/meson.build
+++ b/ci/meson/native/meson.build
@@ -189,6 +189,7 @@ example_compile_args = base_compile_args
 
 # Detect platform for the few remaining differences
 is_windows = build_machine.system() == 'windows'
+is_macos = build_machine.system() == 'darwin'
 
 # ============================================================================
 # Detect if clang-tool-chain Python wrapper is being used
@@ -261,9 +262,13 @@ else
   dll_link_args += ['-rdynamic']
   # In debug mode with sanitizers, shared libraries need to allow undefined symbols
   # that will be provided by the sanitizer runtime when the runner loads them.
-  # LLD by default allows undefined symbols in shared libraries, but we add the
-  # flag explicitly on Unix for compatibility with other linkers.
-  if build_mode == 'debug'
+  # LLD-ELF (Linux) defaults to allowing undefined symbols in shared libraries but
+  # we set the flag explicitly for compatibility with GNU-ld. macOS's ld64.lld
+  # uses Mach-O semantics and does NOT understand `--allow-shlib-undefined`
+  # (it errors out — see FastLED #2612). On Mach-O the equivalent semantic is
+  # `-undefined dynamic_lookup`, which is not needed because Mach-O resolves
+  # undefined symbols at runtime by default.
+  if build_mode == 'debug' and not is_macos
     dll_link_args += ['-Wl,--allow-shlib-undefined']
   endif
 endif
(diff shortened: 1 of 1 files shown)


zackees:
﻿Pushed a follow-on commit (`8abdb2ff0`) that extends this PR with the matching gate for `-Wl,--no-undefined`.

The `example tests macos` log shows the same Mach-O incompatibility hits a *second* flag we emit unconditionally on non-Windows:

```
ld64.lld: error: unknown argument '--no-undefined'
clang++: error: linker command failed with exit code 1
```

`--no-undefined` is added in quick + release modes (`runner_link_args`, lines 338 and 359 of the pre-PR file). Same root cause as `--allow-shlib-undefined`: a Linux-default-on flag treated as portable.

Mach-O's `ld64.lld` defaults to erroring on undefined symbols anyway, so the flag is redundant on macOS. Gated both quick-mode and release-mode sites with `if not is_macos`.

With this commit, both macOS workflows (`unit tests macos` debug mode + `example tests macos` quick mode) should clear once the companion zackees/clang-tool-chain#44 (SDK isysroot) lands in the pinned CTC version.

Issue #2629, opened 2026-05-31
fl::audio::fft: eliminate FL_FFT_USE_ESP_DSP opt-in, auto-detect via FL_HAS_INCLUDE + scalar fallback


zackees:
﻿## Goal

Eliminate the per-sketch `FL_FFT_USE_ESP_DSP` define and the user-facing `@filter` it forces, by routing ESP-DSP through the scalar polyfill that already exists in `fl::audio::fft`. Users should never need to opt in OR out — the right backend should select automatically based on what the toolchain actually ships.

## Background

PR #2625 just landed an `@filter` workaround on `examples/AudioFftParity/AudioFftParity.ino` to keep esp32c2 (and likely esp32s2/h2/c5) green. Root cause was:

1. The example sets `#define FL_FFT_USE_ESP_DSP 1`
2. `src/fl/audio/fft/fft_backend.h:101` `#include "esp_dsp.h"` whenever that define + `FL_IS_ESP32` are both true
3. esp_dsp.h isn't in every ESP32 toolchain bundle (low-end RISC-V parts don't have it)
4. Result: hard compile error on boards where the header is missing

The `@filter` works around the symptom but the underlying design is wrong: any user-authored sketch that opts into ESP-DSP for performance will hit the same wall on a non-supported board, with no diagnostic better than "esp_dsp.h: No such file or directory."

## Proposal — eliminate the define

1. **Auto-detect esp_dsp availability** at preprocessor time:
   ```cpp
   #if defined(FL_IS_ESP32) && FL_HAS_INCLUDE("esp_dsp.h")
       #define FL_FFT_ESP_DSP_AVAILABLE 1
       #include "esp_dsp.h"
   #else
       #define FL_FFT_ESP_DSP_AVAILABLE 0
   #endif
   ```
   Drop the user-opt-in `FL_FFT_USE_ESP_DSP` entirely. `FL_HAS_INCLUDE` already exists in `fl/stl/has_include.h` and is used elsewhere in the codebase for exactly this kind of feature probe (e.g. lcd_spi gates `esp_cache.h` the same way).

2. **Scalar fallback is automatic.** The dispatch logic at the call site already has `#if FL_FFT_ESP_DSP_AVAILABLE / #else` branches — the `#else` path uses kiss_fftr or the scalar polyfill. Switching the gate from "user opted in" to "header is present" means:
   - ESP32 boards with esp_dsp → use the hardware-accelerated backend automatically.
[...]

Issue #2623, opened 2026-05-31
esp32c2: AudioFftParity fails — esp_dsp.h missing on ESP32-C2 toolchain


zackees:
﻿## Symptom

`esp32c2` workflow on master fails compiling `AudioFftParity`:

```
lib/FastLED/fl/audio/fft/fft_backend.h:101:10:
  fatal error: esp_dsp.h: No such file or directory
  101 | #include "esp_dsp.h"
FAILED: AudioFftParity
```

Reproducible: `gh run view 26697900743 --log-failed` (or any recent `esp32c2` run).

## Root cause

`examples/AudioFftParity/AudioFftParity.ino:26` deliberately enables the ESP-DSP backend for this single sketch:

```cpp
#define FL_FFT_USE_ESP_DSP 1
```

`fft_backend.h:99-107` gates the `#include "esp_dsp.h"` on `FL_FFT_USE_ESP_DSP && defined(FL_IS_ESP32)`. Both conditions are true on any ESP32 board.

But **the ESP-DSP library is not part of every ESP32 platform package**. ESP32-C2 (RISC-V, low-end) doesn't ship `esp_dsp.h` in its PlatformIO toolchain bundle. The example has no `@filter` constraint, so it tries to compile on every ESP32 board and dies on C2 (and any other ESP32 variant where esp_dsp is unavailable).

The example's own header comment (lines 16-17) lists `esp32s3` as the canonical target — suggesting the example was authored for a narrower set than "all ESP32" but never had that constraint encoded.

## Fix

Add an `@filter` to `AudioFftParity.ino` that restricts to ESP32 boards that ship `esp_dsp.h`. The conservative initial set: the boards explicitly supported by ESP-DSP upstream — `esp32`, `esp32s3`, `esp32c3`, `esp32c6`, `esp32p4`, `esp32wroom`. Excludes `esp32c2`, `esp32s2`, `esp32h2`, `esp32c5` until each is verified.

Existing audio examples already follow this pattern (`examples/Audio/Audio.ino`, `AudioInput`, `AudioReactive` use `@filter (platform is teensy) or (platform is esp32)`). This PR is the same shape, but narrower because esp_dsp's availability is per-board, not per-platform-family.

## Related (not in scope of this issue)

The same workflow run also shows a separate `RX` (Remote) link failure:
```
undefined reference to `fl::GpioIsrRx::create(int)'
```
[...]

PR #2635, opened 2026-05-31
fix(esp32): provide GpioIsrRx::create stub when FASTLED_RMT5=0

### LINKED ISSUE #2630: esp32c2: GpioIsrRx::create undefined when FASTLED_RMT5=0 (RMT-less chips)
﻿## Symptom

`esp32c2` workflow on master fails on the `RX` (`examples/Remote/Remote.ino`) example with:

```
lib/FastLED/libFastLED.a(fl.channels+_5e4c.o):
  in function `fl::shared_ptr<fl::RxDevice>
              fl::RxDevice::create<(fl::RxDeviceType)1>(int)':
fl.channels+.cpp: undefined reference to `fl::GpioIsrRx::create(int)'
collect2: error: ld returned 1 exit status
FAILED: RX
```

Reproducible from the `esp32c2` workflow log in run 26697900743.

## Root cause

`src/fl/channels/rx.cpp.hpp:99` (compiled for any `FL_IS_ESP32`) instantiates:

```cpp
template <>
fl::shared_ptr<RxDevice> RxDevice::create<RxDeviceType::ISR>(int pin) FL_NOEXCEPT {
    auto device = GpioIsrRx::create(pin);
    ...
}
```

`GpioIsrRx::create()`'s **definition** lives in `src/platforms/esp/32/drivers/gpio_isr_rx/gpio_isr_rx.cpp.hpp`, and its compilation is gated by **two** preprocessor conditions:

```cpp
#ifdef FL_IS_ESP32
#if FASTLED_RMT5                          //  ← outer guard
    #if defined(SOC_MCPWM_SUPPORTED) && SOC_MCPWM_SUPPORTED
        // MCPWM implementation of create()
    #else
        // No-op stub: create() returning nullptr
    #endif
#endif // FASTLED_RMT5                    //  ← stops here on RMT-less chips
#endif
```

On ESP32-C2 (a low-end RISC-V part), there is **no RMT peripheral** (`SOC_RMT_SUPPORTED = 0`). `feature_flags/enabled.h:111-119` then resolves:

```cpp
#if !defined(FASTLED_RMT5)
#if FASTLED_ESP32_RMT5_ONLY_PLATFORM      // false on C2
    #define FASTLED_RMT5 1
#elif ESP_IDF_VERSION >= ESP_IDF_VERSION_VAL(5, 0, 0) && FASTLED_ESP32_HAS_RMT
    #define FASTLED_RMT5 1
#else
    #define FASTLED_RMT5 0                //  ← C2 lands here
#endif
#endif
```

So on C2: `FASTLED_RMT5 = 0` → outer guard fails → **neither** the MCPWM `create()` nor the stub `create()` gets compiled → linker can't find the symbol → build dies.

The caller in `fl/channels/rx.cpp.hpp` is only gated on `FL_IS_ESP32`, not on `FASTLED_RMT5`, so it always tries to link against the symbol.

## Fix

The no-op stub is the right answer for **any** ESP32 board that lacks the hardware path — whether the reason is "no MCPWM" or "no RMT at all". Restructure the gates so the stub is the fallback for both:

```cpp
#ifdef FL_IS_ESP32

#if FASTLED_RMT5
#include "soc/soc_caps.h"
#endif

#if FASTLED_RMT5 && defined(SOC_MCPWM_SUPPORTED) && SOC_MCPWM_SUPPORTED
    // MCPWM implementation
    fl::shared_ptr<GpioIsrRx> GpioIsrRx::create(int pin) FL_NOEXCEPT {
        return GpioIsrRxMcpwm_create(pin);
    }
#else
    // No MCPWM hardware OR no RMT5 support → ISR RX not available
    fl::shared_ptr<GpioIsrRx> GpioIsrRx::create(int pin) FL_NOEXCEPT {
        (void)pin;
        return fl::shared_ptr<GpioIsrRx>();
    }
#endif

#endif // FL_IS_ESP32
```

The caller in `rx.cpp.hpp:99-104` already handles a null return — it wraps it as `DummyRxDevice("GPIO ISR RX creation failed")`. So returning null on C2 gives a clean runtime "feature not available" path instead of a link-time stop-the-build.

## Test plan

- [ ] `bash compile esp32c2 --examples Remote` succeeds (was failing).
- [ ] `bash compile esp32s3 --examples Remote` still succeeds (MCPWM path unchanged).
- [ ] `bash compile esp32c3 --examples Remote` still succeeds (RMT5 yes, MCPWM no → falls into stub).
- [ ] `bash test fl_channels --cpp` continues to pass on the host.

## Related

- #2623 — sibling esp32c2 failure (AudioFftParity `esp_dsp.h`), fixed via filter in #2625.
- #2629 — follow-up to eliminate the `FL_FFT_USE_ESP_DSP` opt-in in favor of `FL_HAS_INCLUDE` probing.

zackees:
﻿## Summary

`esp32c2` workflow on master fails linking `examples/Remote/Remote.ino`:

```
fl.channels+.cpp: undefined reference to `fl::GpioIsrRx::create(int)'
collect2: error: ld returned 1 exit status
FAILED: RX
```

## Root cause

`src/fl/channels/rx.cpp.hpp:99` is compiled for any `FL_IS_ESP32` and references `GpioIsrRx::create(pin)` unconditionally. The **definition** in `src/platforms/esp/32/drivers/gpio_isr_rx/gpio_isr_rx.cpp.hpp` was gated on `FASTLED_RMT5` outer + `SOC_MCPWM_SUPPORTED` inner:

```cpp
#if FASTLED_RMT5
    #if SOC_MCPWM_SUPPORTED
        // MCPWM impl
    #else
        // no-op stub returning null
    #endif
#endif  // ← falls off the cliff when FASTLED_RMT5 = 0
```

On ESP32-C2 (low-end RISC-V, **no RMT peripheral at all**), `feature_flags/enabled.h` resolves `FASTLED_RMT5 = 0`. Neither the MCPWM path nor the stub was compiled → symbol undefined → link error.

## Fix

Combine both gates into a single condition so the no-op stub is the fallback for *either* missing piece:

```cpp
#if FASTLED_RMT5 && defined(SOC_MCPWM_SUPPORTED) && SOC_MCPWM_SUPPORTED
    // MCPWM impl
#else
    // no-op stub returning null  (RMT-less OR MCPWM-less chips)
#endif
```

The caller (`rx.cpp.hpp:99-104`) already handles a null return — wraps it as `DummyRxDevice("GPIO ISR RX creation failed")`. Runtime semantics for chips that already worked are unchanged; new behavior on C2 is "feature unavailable at runtime" instead of "build doesn't link". That matches the file's docstring intent ("Returns nullptr on platforms without MCPWM").

`#include "soc/soc_caps.h"` is now only pulled in when `FASTLED_RMT5` is true — it's only available under the ESP-IDF tree when RMT5 is in play.

## Verification

```
$ bash compile esp32c2 --examples Remote
(in progress — verification pending)
```

(I'll merge once `bash compile esp32c2 --examples Remote` reports success, per the loop directive's authorization.)

## Related

- #2623, PR #2625 — sibling esp32c2 failure (AudioFftParity) fixed via `@filter`
- #2629 — broader follow-up to eliminate the `FL_FFT_USE_ESP_DSP` opt-in via `FL_HAS_INCLUDE` probe

Fixes #2630.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced GPIO ISR RX driver reliability and overall hardware compatibility across all ESP32 variants. Improved handling for boards with limited hardware capabilities, particularly ESP32-C2 and ESP32-C3 platforms. The driver now provides more robust and consistently stable performance, reducing potential issues during build and runtime operations across the entire ESP32 ecosystem.

### DIFF
--- a/src/platforms/esp/32/drivers/gpio_isr_rx/gpio_isr_rx.cpp.hpp
+++ b/src/platforms/esp/32/drivers/gpio_isr_rx/gpio_isr_rx.cpp.hpp
@@ -6,7 +6,9 @@
  *
  * This file redirects the GpioIsrRx::create() factory method to the
  * dual-ISR MCPWM implementation on platforms with MCPWM hardware.
- * Returns nullptr on platforms without MCPWM (e.g., ESP32-C3/C2).
+ * Returns nullptr (which the caller wraps as DummyRxDevice) on platforms
+ * without MCPWM (e.g., ESP32-C3/C2) OR without RMT5 (e.g., ESP32-C2,
+ * which lacks the RMT peripheral entirely).
  */
 
 #include "platforms/is_platform.h"
@@ -22,11 +24,19 @@
 #include "platforms/esp/32/feature_flags/enabled.h"
 // IWYU pragma: end_keep
 
+// soc_caps.h is only available under the ESP-IDF tree; only pull it in when
+// we'll actually consult SOC_MCPWM_SUPPORTED (i.e. when FASTLED_RMT5 is on).
 #if FASTLED_RMT5
-
-// Check if this SoC has MCPWM hardware
 #include "soc/soc_caps.h"  // IWYU pragma: keep
-#if defined(SOC_MCPWM_SUPPORTED) && SOC_MCPWM_SUPPORTED
+#endif
+
+// The caller in `src/fl/channels/rx.cpp.hpp` is gated only on `FL_IS_ESP32`
+// and references `GpioIsrRx::create()` unconditionally. We must therefore
+// provide a definition for every ESP32 board — even ones without RMT5 or
+// MCPWM (e.g. ESP32-C2 has neither). The MCPWM-backed path only fires
+// when both gates are positive; everything else falls into the stub.
+// See FastLED #2630.
+#if FASTLED_RMT5 && defined(SOC_MCPWM_SUPPORTED) && SOC_MCPWM_SUPPORTED
 
 // IWYU pragma: begin_keep
 #include "platforms/esp/32/drivers/gpio_isr_rx/gpio_isr_rx_mcpwm.h"
@@ -41,11 +51,13 @@ fl::shared_ptr<GpioIsrRx> GpioIsrRx::create(int pin) FL_NOEXCEPT {
 
 } // namespace fl
 
-#else // !SOC_MCPWM_SUPPORTED
+#else // !FASTLED_RMT5 OR !SOC_MCPWM_SUPPORTED
 
 namespace fl {
 
-// No MCPWM hardware on this SoC - ISR RX not supported
+// No MCPWM hardware (or no RMT5 support) on this SoC — ISR RX not
+// available. Returning null lets the caller substitute DummyRxDevice
+// at runtime rather than failing to link.
 fl::shared_ptr<GpioIsrRx> GpioIsrRx::create(int pin) FL_NOEXCEPT {
     (void)pin;
     return fl::shared_ptr<GpioIsrRx>();
@@ -53,6 +65,5 @@ fl::shared_ptr<GpioIsrRx> GpioIsrRx::create(int pin) FL_NOEXCEPT {
 
 } // namespace fl
 
-#endif // SOC_MCPWM_SUPPORTED
-#endif // FASTLED_RMT5
+#endif // FASTLED_RMT5 && SOC_MCPWM_SUPPORTED
 #endif // FL_IS_ESP32


zackees:
﻿Local verification update: `bash compile esp32c2 --examples Remote` failed at the **framework** compile step (long before reaching the linker that this PR fixes):

```
esp32-hal-gpio.c -> esp32-hal.h -> esp_sleep.h:12:10:
  fatal error: hal/touch_sensor_legacy_types.h: No such file or directory
```

This is an upstream Arduino-ESP32 SDK packaging gap — `touch_sensor_legacy_types.h` is shipped in the 3.3.7 framework cache's `esp32/c3/c5/c6/h2` subsets but **not** in the `esp32c2` subset, while `esp_sleep.h` includes it unconditionally. Verified by `find ~/.fbuild ... -name "touch_sensor_legacy_types.h"` — every other ESP32 has it, c2 doesn't.

The CI run that surfaced #2630 (run 26697900743) **did not hit** this framework error — it compiled the framework + FastLED library fine and only died at the link step on `GpioIsrRx::create`. So CI's framework cache state is different from the local one. Merging on the CI's verification rather than reproducing locally.

The fix itself is mechanically narrow:
- Only changes behavior on chips where `FASTLED_RMT5=0`, which today is just ESP32-C2 (every other ESP32 variant has RMT and therefore RMT5=1).
- For all other chips the gate evaluation is equivalent: `FASTLED_RMT5 && SOC_MCPWM_SUPPORTED` produces the same true/false as the previous nested `#if FASTLED_RMT5` outer + `SOC_MCPWM_SUPPORTED` inner.
- The new stub-fallback case for C2 returns `nullptr`, which the caller already wraps as `DummyRxDevice` at runtime.

Will file the framework SDK gap as a separate issue so users who hit it locally have a tracking link.

Issue #2653, opened 2026-05-31
nrf52840_dk / adafruit_feather_nrf52840_sense / nrf52840_supermini red: arm-gnu 15.2 LTO offset overflow


zackees:
﻿## Symptom

Three nRF52840 workflows have started failing on master with arm-gnu 15.2.rel1 LTO assembler errors:

- `nrf52840_dk` — [run 26702036353](https://github.com/FastLED/FastLED/actions/runs/26702036353)
- `adafruit_feather_nrf52840_sense` — [run 26702036295](https://github.com/FastLED/FastLED/actions/runs/26702036295)
- `nrf52840_supermini` — [run 26702036347](https://github.com/FastLED/FastLED/actions/runs/26702036347)

Identical signature on all three:

```
.lto-tmp/ccEo9QPb.s:13834: Error: offset out of range
.lto-tmp/ccEo9QPb.s:13863: Error: offset out of range
.lto-tmp/ccEo9QPb.s:13894: Error: offset out of range
...
build error: build failed: arm-none-eabi-gcc link failed:
lto-wrapper: fatal error: arm-none-eabi-gcc returned 1 exit status
❌ Compilation failed (fbuild)
```

## Root cause

Same arm-gnu-toolchain **15.2.rel1** LTO bug PR #2642 documented for `xiaoblesense_adafruit` (#2641): the LTO assembler temp file (`.lto-tmp/*.s`) contains Thumb-2 branch offsets exceeding the encodable range when the linker is asked to emit the merged firmware. The Adafruit nRF52 BSP at 1.10601.0 brings in enough core code that the LTO-merged TU exceeds the Thumb branch reach across function boundaries.

PR #2642 fixed `xiaoblesense_adafruit` by setting `build_unflags=["-flto"], build_flags=["-fno-lto"]`. PR #2644 then applied the same workaround to `xiaoblesense`. The three nRF52840 boards above use the **same Adafruit BSP + same toolchain**, so the same bug strikes them — the LTO ramp has finally crossed the threshold for the non-Seeed nRF52840 variants too.

## Fix

Apply the same `-fno-lto` workaround to the three failing nRF52840 board entries in `ci/boards.py`:

```python
build_unflags=["-flto"],
build_flags=["-fno-lto"],
```

Targets:
- `NRF52840` (`board_name="nrf52840_dk"`) — currently no build flag overrides
- `SUPERMINI_NRF52840` (`board_name="supermini_nrf52840"`) — currently no build flag overrides
[...]

Issue #2641, opened 2026-05-31
build: nRF52840 Adafruit-BSP LTO link failure — "offset out of range" in arm-gnu 15.2.1 LTO


zackees:
## Symptom

All nRF52840 boards built against the Adafruit BSP + arm-gnu-toolchain 15.2.rel1 (15.2.1) under LTO fail at link with thousands of `offset out of range` errors in the LTO-temp assembly:

\`\`\`
.../release/.lto-tmp/ccXXXXXX.s:13834: Error: offset out of range
.../release/.lto-tmp/ccXXXXXX.s:13863: Error: offset out of range
.../release/.lto-tmp/ccXXXXXX.s:13894: Error: offset out of range
lto-wrapper: fatal error: arm-none-eabi-gcc returned 1 exit status
arm-none-eabi/bin/ld: error: lto-wrapper failed
collect2: error: ld returned 1 exit status
\`\`\`

This is happening on:
- \`build_xiaoblesense_adafruit\` (post-#2636: https://github.com/FastLED/FastLED/actions/runs/26700353292)
- \`build_supermini_nrf52840\` (master: https://github.com/FastLED/FastLED/actions/runs/26699597732)
- likely \`nice_nano\` / \`nRFMicro\` and \`adafruit_feather_nrf52840_sense\` too

Toolchain path: \`/home/runner/work/_temp/fbuild-cache/toolchains/developer-gnu/e0640e442d96f78c/15.2.Rel1/arm-gnu-toolchain-15.2.rel1-x86_64-arm-none-eabi/\`.

## Root cause hypothesis

\`offset out of range\` from the assembler on LTO output usually means a 16-bit branch offset overflowed. Common triggers for arm-none-eabi 15.x + Cortex-M4 + LTO:

1. Code section bigger than the unconditional-branch reach (±16 MB for B/BL, ±20 MB for B.W) due to LTO-induced inlining — unlikely on 1 MB flash, so more likely:
2. **Long-call mismatch**: LTO merges code into a different section than the linker expects, and short-form branches are emitted for what end up as long-distance calls. Workaround: add `-mlong-calls` (or selectively `-mlong-calls -mno-long-calls` per TU), or disable LTO entirely (`-fno-lto`).
3. arm-gnu 15.2.rel1 LTO regression — worth checking the toolchain release notes.

## Proposed fix

[...]

PR #2658, opened 2026-05-31
ci(optimization-report): cap stdout dump at 1 MiB to keep GHA runner alive

zackees:
## Summary
- `-fopt-info-all` writes a per-board optimization report. Across the full example matrix (~110 sketches on nrf52840_dk + the other Adafruit-BSP boards), the report routinely exceeds **100 MB**.
- The **Optimization Report** step in `build_template.yml` dumps the entire file to the GitHub Actions log via `print(text)`. This overruns the runner's log buffer and triggers an external **runner shutdown signal**.
- Symptom: the build step succeeds, several post-build analytics steps succeed, then Optimization Report cancels at ~37s and the runner dies; downstream cleanup steps stay pending and the workflow is marked **FAILURE** despite the build being **green**.

Verified failures on master SHA `a302c1ba`:
- nrf52840_supermini [run 26705682359](https://github.com/FastLED/FastLED/actions/runs/26705682359) — build SUCCESS at 08:51:29, Optimization Report cancelled 08:55:02
- adafruit_feather_nrf52840_sense [run 26705682371](https://github.com/FastLED/FastLED/actions/runs/26705682371) — same pattern, build SUCCESS at 08:51:29, Optimization Report cancelled 08:55:02

## Fix
- Default to printing the **last 1 MiB** of the report (most-recently-emitted optimizer info, which is the most useful slice for size-driven debugging).
- `--max-bytes 0` opts back into the full dump for local use.
- Tail is snapped to the next line boundary so the output doesn't lead with a half-line.
- Prints a one-line announcement up front showing the total size and how many bytes were elided.

## Counterpart to PR #2655
#2655 added `timeout-minutes: 5` to Symbol Analysis + Optimization Report, but couldn't help when the runner shutdown arrives **faster than** the step timeout (the log-buffer overrun is faster than 5 minutes).

## Test plan
- [ ] CI passes for the affected nRF52840 workflows once this lands on master.
- [ ] Existing local `uv run ci/optimization_report.py --first --cwd /tmp/some/build` still produces useful output for normal-sized reports (the cap is only enforced when the report exceeds 1 MiB).

Refs #2641, #2653.

## Summary by CodeRabbit

* **New Features**
  * Optimization reports output is now capped at a configurable limit, displaying the tail portion when truncated
  * Added --max-bytes CLI flag and FL_OPT_REPORT_MAX_BYTES environment variable to customize the default report size (0 disables the cap)
  * Invalid or missing env values fall back to a sane default; negative values are treated as zero
  * Truncated reports include a summary showing total size, bytes displayed, and bytes omitted; CLI help text clarified for inspection use

### DIFF
--- a/ci/optimization_report.py
+++ b/ci/optimization_report.py
@@ -32,7 +32,7 @@ def _env_default_max_bytes() -> int:
     return max(value, 0)
 
 
-def parse_args() -> argparse.Namespace:
+def parse_args(argv: list[str] | None = None) -> argparse.Namespace:
     parser = argparse.ArgumentParser(
         description="Inspect the GCC -fopt-info-all optimization report for a board."
     )
@@ -48,11 +48,11 @@ def parse_args() -> argparse.Namespace:
             f"env var; set to 0 to disable the cap entirely)."
         ),
     )
-    return parser.parse_args()
+    return parser.parse_args(argv)
 
 
-def main() -> int:
-    args = parse_args()
+def main(argv: list[str] | None = None) -> int:
+    args = parse_args(argv)
     if args.cwd:
         root_build_dir = args.cwd / ".build"
     else:
--- a/ci/optimization_report.py
+++ b/ci/optimization_report.py
@@ -3,12 +3,30 @@ import sys
 from pathlib import Path
 
 
+# Cap how much of optimization_report.txt we dump to stdout. With -fopt-info-all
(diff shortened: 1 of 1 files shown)


zackees:
Addressed in commit 11c013a80: `parse_args` and `main` now accept an optional `argv: list[str] | None = None` for deterministic test injection. Behavior unchanged for CLI use (argv stays `None` → argparse defaults to `sys.argv[1:]`).

PR #2661, opened 2026-05-31
fix(ci): probe fbuild compile-many via SUB --help (unbreaks CI fast path)

### LINKED ISSUE #2660: fbuild compile-many silently disabled in CI due to wrong probe form (`fbuild help SUB` exits 2)
## Symptom

Every `build_<board>.yml` CI run on `master` logs:

> `fbuild ci and compile-many are unavailable in the installed fbuild; falling back to the legacy serial loop.`

…and then builds 100+ examples one-by-one. See e.g. uno run [78718228931](https://github.com/FastLED/FastLED/actions/runs/26709963258/job/78718228931):

```
2026-05-31T10:21:45Z Starting compilation for 1 boards with 108 examples
2026-05-31T10:21:45Z fbuild ci and compile-many are unavailable in the installed fbuild;
                     falling back to the legacy serial loop.
```

The installed fbuild on that runner (2.2.6+) actually does support both subcommands.

## Root cause

`ci/util/fbuild_runner.py:146-168`:

```python
proc = subprocess.run(
    [fbuild_exe, "help", subcommand],   # <-- wrong invocation form
    ...
)
return proc.returncode == 0
```

`fbuild help ci` parses `help` as the subcommand name and `ci` as a positional sketch path arg, then clap rejects it for missing the required `--board <BOARD>`:

```
$ fbuild help ci
error: the following required arguments were not provided:
  --board <BOARD>
Usage: fbuild.exe ci --board <BOARD> <SKETCHES>...
$ echo $?
2
```

…so `_fbuild_supports_subcommand` always returns False, `fbuild_supports_ci()` / `fbuild_supports_compile_many()` always return False, and `PioCompiler._build_with_fbuild` (`ci/compiler/pio.py:431-444`) always falls through to `_build_fbuild_sync` — the legacy per-sketch serial loop.

`fbuild <SUB> --help` is intercepted by clap before arg validation and exits 0 regardless of whether other required args are present.

## Impact

Every CI build on master serially loops over examples instead of using `compile-many`'s stage-2 worker pool. On the GitHub Actions 2-core runners that's roughly **2× build wall time** (FastLED defaults `sketch_jobs` to `min(cpus, 2)` on GitHub Actions per `fbuild_runner.py:135-143`). Recent uno run: ~3.5 min for 108 examples; with compile-many it would be closer to ~2 min.

## Fix

Swap the probe form to `<fbuild> <subcommand> --help`. PR coming next; includes a regression test (`test_fbuild_supports_subcommand_uses_sub_help_form`) that asserts the exact command line emitted, so the same drift can't silently re-disable compile-many.

zackees:
## Summary

- Every build_*.yml CI run on master has been logging `fbuild ci and compile-many are unavailable in the installed fbuild; falling back to the legacy serial loop` and silently doing per-sketch serial builds instead of using `compile-many`'s stage-2 worker pool.
- Root cause: `_fbuild_supports_subcommand` (ci/util/fbuild_runner.py:146) probed with `<fbuild> help <SUB>`, which clap parses as subcommand=help with `<SUB>` as a positional sketch arg, then rejects with exit 2 on the missing `--board`. Switch to `<fbuild> <SUB> --help`, which clap intercepts and exits 0 regardless of other required args.
- Add a regression guard so the next refactor can't quietly re-disable compile-many.

Fixes #2660. Expected CI win on GitHub Actions (2-core runners, default `sketch_jobs=min(cpus,2)=2`): roughly 2× faster per-board builds, e.g. uno's 108-example run drops from ~3.5 min to ~2 min based on the current per-sketch wall times.

## Test plan

- [x] Local probe against fbuild 2.2.6 returns True for both `ci` and `compile-many`
- [x] Inline pytest regression guard asserts the exact command line emitted
- [ ] One uno CI run with the new branch shows the green path log lines (`Starting compilation...` followed by `compile-many results:` instead of the fallback message)

## Summary by CodeRabbit

* **Bug Fixes**
  * Restored reliable fbuild subcommand detection (fixes CI compile-many fallback).

* **Tests**
  * Added a regression test to lock in the new subcommand-detection behavior.

* **New Features**
  * Enriched build output with per-sketch Flash/RAM size parsing, timing, and success/failed annotations.
  * Added a batch summary showing total wall time, parallelism factor, and ok/fail counts.

* **Chores**
  * Workflow step: added an install-cache key input keyed to project configuration.

### DIFF
--- a/.github/workflows/build_template.yml
+++ b/.github/workflows/build_template.yml
@@ -64,6 +64,11 @@ jobs:
         uses: FastLED/fbuild/.github/actions/setup@main
         with:
           cache-key-extra: ${{ steps.fbuild_board.outputs.name }}-${{ hashFiles('platformio.ini', 'library.json', 'pyproject.toml', 'uv.lock') }}
+          # Binary install cache is intentionally board-independent — every
+          # job in this repo's board matrix shares the same fbuild binary,
+          # so key the install cache solely on the fbuild pin in pyproject.
+          # The action ignores this input on older revisions (no-op).
+          install-cache-key-extra: ${{ hashFiles('pyproject.toml') }}
 
       - name: Install Platform
         if: inputs.platform != ''
--- a/ci/compiler/pio.py
+++ b/ci/compiler/pio.py
@@ -537,6 +537,11 @@ class PioCompiler(Compiler):
             self.platform_lock.release()
             return futures
 
+        import time
+
+        from ci.util.fbuild_runner import _parse_size_info_from_log
+
+        batch_start = time.monotonic()
         try:
             compile_many_result = run_batch(
                 board=self.board.board_name,
@@ -551,6 +556,9 @@ class PioCompiler(Compiler):
                 for sketch_result in compile_many_result.sketch_results
             }
 
+            success_count = 0
+            fail_count = 0
+            per_sketch_secs: list[float] = []
             for example, project_dir in staged_projects:
                 sketch_result = reported.get(project_dir.resolve())
                 if sketch_result is None:
@@ -559,6 +567,10 @@ class PioCompiler(Compiler):
                         f"{project_dir}\n\n{compile_many_result.output}"
                     )
                     success = False
+                    stage = "?"
+                    build_secs = 0.0
+                    flash_str: str | None = None
+                    ram_str: str | None = None
                 else:
                     output = sketch_result.message
                     if (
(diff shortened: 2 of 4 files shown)


zackees:
Status update: the dispatch is now gated behind \`FASTLED_USE_FBUILD_CI=1\` (default off).

The detection probe fix on this branch made compile-many reachable for the first time in CI. teensy41 and esp32s3 immediately timed out at 30 minutes because compile-many re-builds the framework from scratch in every stage-2 sketch dir — see [FastLED/fbuild#335](https://github.com/FastLED/fbuild/issues/335). On uno alone it turns a ~3.5 min serial run into ~7 min.

Until [fbuild#335](https://github.com/FastLED/fbuild/issues/335) ships a shared framework cache, compile-many is a net regression for FastLED CI, so the dispatch path is gated and the default goes back to the legacy serial loop. Everything else on this branch — the \`SUB --help\` probe fix, the normalized per-sketch output, the install-cache wiring + workflow hash — stays active and useful.

To opt back in locally: \`FASTLED_USE_FBUILD_CI=1 ./compile <board> all\`.

CI on this push should now mirror master's serial-loop times (uno ~3.5 min, etc).

Issue #342, opened 2016-09-06
NodeMCU with FastLED


beebrain:
I tried to use FashtLED and I got error many time.
I read and follow instruction. 
I got this error message 
C:\Users\Boobeejung\Documents\Arduino\libraries\FastLED-3.1.0/led_sysdefs.h:22:2: error: #error "XTENSA-architecture microcontrollers are not supported."
 #error "XTENSA-architecture microcontrollers are not supported."

I installed last version FastLED library it's version 3.1 and update package esp8266 board in arduino IDE 
in figure 

![image](https://cloud.githubusercontent.com/assets/10016227/18283546/fd4dce36-748f-11e6-97b1-260216eef0bc.png)

How to fix it??

focalintent:
You need to get the most recent version of the library's master branch on GitHub.

Issue #337, opened 2016-08-09
Instable  wifi  connection when using FastLED


mkeyno:
Hi @focalintent  ,I don’t exactly know whether this issue should be place in ESP8266 repo or FastLED, but I will open ticket in ESP8266 repo as well  and referenced to this 
However the problem is I made sketch base on @hsiboy  and @jasoncoon  sample ,which user can make arbitrary time schedule of couple of pre define effects, the effect scenario  parameters  can be set by webserver  and websocket service .  the sketch work fine for couple of minutes but after a while, Wifi stack got chocked and  wifi connection disappeared or get reset, I run some test and find out if replace ```FastLED.delay()``` function with simple delay or yield function this phenomena happened less , however I’m sure it is related  to  interfere of interrupt and wifi stack when use webserver or websocket service 
I paste the main part of sketch as follow , at first create the pointer function  variable which is updated by user   during the show 
Any help is really appreciated 
```cpp
#define FASTLED_ALLOW_INTERRUPTS 0
 //#define FASTLED_FORCE_SOFTWARE_SPI
#define FASTLED_INTERRUPT_RETRY_COUNT 0 

#include <Arduino.h>
#include <ESP8266WiFi.h>
#include <WiFiClient.h>
#include <ESP8266WebServer.h>
#include <ESP8266mDNS.h>
#include <FS.h>
#include <WebSocketsServer.h>
#include <Hash.h>
#include <EEPROM.h>
#include <ESP8266HTTPClient.h>
#include <FastLED.h>


WebSocketsServer webSocket = WebSocketsServer(81);
ESP8266WebServer server(80);
#define DATA_PIN    4                      
#define NUM_LEDS    150
#define LED_TYPE    APA104
#define COLOR_ORDER GRB//GRB GBR RGB RBG BGR BRG
#define UPDATES_PER_SECOND 1000
CRGBArray<NUM_LEDS> leds;
CRGB _color,COLORS[30],Color16[16];

uint16_t Duration[30];
byte BRIGHTNESS[30],VELOCITIES[30];
int EffectNum[30]={0},line_length=5,Effect_Index=0;
byte  _code[30],velocity=10,Brightness=200,Number_Of_Scenario=0;
 byte  Q=1,NumberofCustom=0;
 
CRGBPalette16 currentPalette=RainbowColors_p,PALLETS[10];
TBlendType    currentBlending,BELENDS[10];

//extern           CRGBPalette16 myRedWhiteBluePalette;NumberofCustom
[...]

Issue #362, opened 2016-10-26
Wemos D1 R2 resetting


gavspav:
My board is resetting. 
My sketch causes frequent resets: rst cause:2, boot mode:(3,6)
At first I thought it was Ram or Power issues but I have ruled them out.
Code works with Teensy
Apologies if it is not an issue with the library.

```
include "FastLED.h"

#define NUM_LEDS 8 //number of leds in strip length on one side
#define DATA_PIN D2//7 = second hardware spi data
#define CLOCK_PIN D3//14 = second hardware spi clock
#define COLOR_ORDER BGR
CRGB leds[NUM_LEDS];
int numberOfSlices = 8;

void PoiSonic(unsigned long time, const unsigned int array[]){
unsigned long currentTime = millis();
 while (millis()< currentTime + (time)) {

int f= numberOfSlices;
int z; //a counter
int j=NUM_LEDS;

    for (int x=0;x<f;x++){
     for(z=NUM_LEDS;z>0;z--){
       leds[z-1]=array[x+((j-z)*f)];
       
       }
    
     FastLED.show();
     delayMicroseconds(1200); //may need to increase / decrease depending on spin rate
     }       
    delayMicroseconds(1000); //may need to increase / decrease depending on spin rate
   }
 }
void setup() {

  delay(200);
    FastLED.addLeds<APA102, DATA_PIN, CLOCK_PIN, COLOR_ORDER>(leds, NUM_LEDS).setCorrection(TypicalLEDStrip);;
    FastLED.clear();
    FastLED.show();
 }


 
[...]

Issue #361, opened 2016-10-25
Teensy 3.5, static_assert(validpin(), "Invalid pin specified");


tehuster:
When i try to use pin 35/36 on the teensy 3.5 i get the error: static_assert(validpin(), "Invalid pin specified");

If i understand it correctly i need to define a pinmapping for the teensy in the fastpin.h? 
The problem is, I don't know how to do that. If you point me in the right direction I would be very gratefull. 

Thanks

jasoncoon:
Teensy 3.5 and 3.6 aren't supported by FastLED yet.  Daniel has said he has the hardware, but hasn't had the time yet: https://plus.google.com/u/0/+BrianLewis515/posts/e9PN9heiWcR

Side note, I noticed Paul has added Teensy 3.5 and 3.6 support to the OctoWS2811 library: https://github.com/PaulStoffregen/OctoWS2811/commit/d51e6a65c63e1d8b4539c7a93fc3c1c42ccdbd4b

tehuster:
Thanks for the info Jason. Guess i will just wait on Daniel or figure out how i need to pinmap it myself.

robertcedwards:
Ah shoot, that's the problem! I was having a bunch of funky issues but using 3.0.1 seems to have done the trick for our Dotstars, along with with slowing down our clock as per Paul's recommendation:
https://plus.google.com/111022735393886581158/posts/JcCkB7j66aj?sfc=true

Mad-Onion:
@robertcedwards can yout tell me what you did to get the Dotstars running on T3.5? I'm using arduino 1.6.12, teensyduino 1.3.1-beta3 and FastLED 3.1.3 but still get the error mentioned in the title.

robertcedwards:
I'm running 1.6.10 with FastLED 3.1.0, that FastLED library version was key to getting it to run properly.

Mad-Onion:
@robertcedwards thank you for the info, I'm testing Arduino 1.6.12 with FastLED 3.1.0 and WS2812Bs now and when I try to compile the blink-example I run into lots of asm-constraint and macro-expansion-errors.
Any ideas?

robertcedwards:
@Mad-Onion We had to roll back to Arduino 1.6.10 as well. There was definitely a lot of testing and multiple software setups before we got it working. I'm able to compile it currently with the IDE, so that should be correct.

kcouck:
[...]

PR #2672, opened 2026-06-01
ci: flip FASTLED_USE_FBUILD_CI to default-on (compile-many is the default)

zackees:
With [#2671](https://github.com/FastLED/FastLED/pull/2671) on master, FastLED is now on fbuild 2.2.13 — which ships the compile-many stage-2 framework-archive share ([FastLED/fbuild#337](https://github.com/FastLED/fbuild/pull/337)). That fix is what closes the gap that originally forced the gate to default-off in [#2662](https://github.com/FastLED/FastLED/pull/2662): stage-2 sketches used to rebuild the framework from scratch in their own per-sketch project dirs, which on uno took ~7 min vs ~3.5 min serial and on teensy41/esp32s3 blew past the 30 min batch timeout. With the share in place, locally that's a **~19× cold-cache speedup on uno** (200ms per stage-2 sketch vs ~3.7s prior).

## Polarity flip

| | Before | After |
|---|---|---|
| Default | legacy serial loop | compile-many (or \`fbuild ci\` when supported) |
| To select compile-many | \`FASTLED_USE_FBUILD_CI=1\` | unset / any non-falsy value |
| To select legacy serial | unset / falsy | \`FASTLED_USE_FBUILD_CI={0,false,no,off}\` |

The env var name and its truthy values (\`1\`/\`true\`/\`yes\`/\`on\`) are preserved — anyone who has the opt-in baked into a wrapper script keeps working.

## Safety / rollback

If anything goes sideways post-merge, set \`FASTLED_USE_FBUILD_CI=0\` in the affected workflow's env block — that immediately puts the path back on the legacy serial loop without touching code. The whole gate-flip is a polarity change in one method (\`PioCompiler._build_fbuild\`) plus a clarifying comment in the existing test; nothing else changes.

## Risk evaluation

The original #2662 break happened because compile-many with the framework-rebuild-per-sketch regression was unconditionally enabled — there was no escape valve. This PR has two key differences:

1. Stage-2 framework rebuild is **fixed** upstream in fbuild 2.2.13 ([#337](https://github.com/FastLED/fbuild/pull/337)); locally verified 19× speedup on uno via [#342](https://github.com/FastLED/fbuild/pull/342) (real-toolchain regression test).
2. The env var still exists as an **opt-out** — if a board-specific CI run hits something we didn't anticipate, flipping it back is a one-line workflow env change, not a code revert.

## Summary by CodeRabbit

* **Performance**
  * Batched compilation is now enabled by default for faster builds; it can be disabled via build configuration.
  * Batch execution timeout increased to 3600s and is now configurable.
  * Increased default parallelism for framework build jobs to better utilize available CPU cores.

* **Documentation**
  * Updated build configuration guidance and environment-setting instructions to reflect the new defaults.

### DIFF
--- a/ci/compiler/pio.py
+++ b/ci/compiler/pio.py
@@ -559,17 +559,29 @@ class PioCompiler(Compiler):
             self.platform_lock.release()
             return futures
 
+        import os
         import time
 
         from ci.util.fbuild_runner import _parse_size_info_from_log
 
+        # 30 min was the historical per-sketch timeout the serial path
+        # used; for the batched compile-many path the same number has to
+        # cover stage-1 framework build + stage-2 fanout of *every*
+        # sketch. On 2-core GitHub Actions runners with esp32s3 /
+        # teensy41 the framework alone outlasted 30 min in PR #2672 even
+        # with framework_jobs bumped to 2. Default to 60 min and let CI
+        # workflows tune via FASTLED_FBUILD_BATCH_TIMEOUT_SECS if a
+        # specific board needs more / less.
+        batch_timeout = float(
+            os.environ.get("FASTLED_FBUILD_BATCH_TIMEOUT_SECS", "3600")
+        )
         batch_start = time.monotonic()
         try:
             compile_many_result = run_batch(
                 board=self.board.board_name,
                 sketch_project_dirs=[project_dir for _, project_dir in staged_projects],
                 verbose=self.verbose,
-                timeout=1800,
+                timeout=batch_timeout,
                 quiet=False,
                 log_file=None,
             )
--- a/ci/compiler/pio.py
+++ b/ci/compiler/pio.py
@@ -426,15 +426,20 @@ class PioCompiler(Compiler):
     def _build_fbuild(self, examples: list[str]) -> list[Future[SketchResult]]:
         """Build examples using fbuild, preferring ``ci`` when available.
 
-        compile-many / ``fbuild ci`` are gated behind ``FASTLED_USE_FBUILD_CI``
-        because today they re-build the framework from scratch in every
-        stage-2 sketch dir (FastLED/fbuild#335). On uno that's a ~7 min run
-        vs. ~3.5 min serial; on teensy41/esp32s3 (larger framework) it
-        blows past the 30 min batch timeout — see the runs that landed on
-        master commits d2a025441e / 63e0241808 / fa147ae894. Default off
-        until fbuild#335 ships a shared framework cache; set
-        ``FASTLED_USE_FBUILD_CI=1`` to opt in (e.g. for local timing
-        experiments against the architectural fix once it lands).
+        compile-many / ``fbuild ci`` are now the default. They used to be
+        gated default-off via ``FASTLED_USE_FBUILD_CI`` because every
+        stage-2 sketch re-built the framework from scratch
+        (FastLED/fbuild#335), which on uno took a ~7 min run vs ~3.5 min
+        serial and on teensy41/esp32s3 blew past the 30 min batch
+        timeout. fbuild 2.2.13 ships the framework-archive share fix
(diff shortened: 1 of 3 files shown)


zackees:
Pushed a follow-up commit (`b71843c`) — `default_fbuild_framework_jobs` now mirrors the shape of `default_fbuild_sketch_jobs` and caps at `min(cpu_count, 2)` instead of the previous hardcoded `1`.

**Why:** the first esp32s3 / teensy41 runs on this branch timed out at the 1800s batch deadline. The log has zero `stage2` or `[stage` mentions — `fbuild ci` ran past 30 min still in stage 1. Root cause was `--framework-jobs 1` making stage-1 framework compilation single-threaded on 2-core runners. With ~150 framework files for esp32-arduino at a few seconds each on a single core, stage 1 alone outlasts the timeout before [#337](https://github.com/FastLED/fbuild/pull/337)'s framework-share seed gets to fan stage 2 out.

Cap at 2 mirrors the fbuild orchestrator's own `default_framework_jobs` policy (framework compile is memory-heavy on small runners).

Re-running CI.

zackees:
Pushed third commit `d313f48` — batch timeout default raised from 1800s → 3600s, env-tunable via `FASTLED_FBUILD_BATCH_TIMEOUT_SECS`.

**Where the previous one fell short:** even with `framework_jobs=2` (commit `b71843c`), the esp32s3 / teensy41 jobs still hit the timeout after 30 min with **zero** stage transitions logged — stage 1 alone outlasted 1800s on the 2-core GHA runners for a cold esp32-arduino framework cache. framework_jobs=2 halved the parallelism deficit but didn't recover the rest. The wrapper's 1800s timeout was sized for the legacy per-sketch serial path; the compile-many batched path now has to fit stage-1 framework + stage-2 fanout of *every* sketch into the same budget.

60 min should be plenty for any board's cold-cache batch. Workflows that get faster (warm cache after first save, or once we figure out a way to keep the daemon's framework artifacts across runs) can tune it lower.

zackees:
Closing without merge — the compile-many gate flip is a net regression on the big-framework boards.

### What we observed
Three CI attempts on this branch (with `FASTLED_FRAMEWORK_JOBS` defaulting to 1, then 2, with 1800s then 3600s wrapper timeout):
- `uno`, `wasm`, unit/example tests: PASS (small-framework boards)
- `esp32s3`: hit 30-min then 60-min timeout, never finished
- `teensy41`: same pattern

Comparison against master (serial loop, `FASTLED_USE_FBUILD_CI=0` default):
- master `esp32s3`: ~24 min consistently across last 5 runs (24/24/27/47/24)
- this PR `esp32s3`: still in_progress at 59 min → ~2.5× slower

`uno` finishes in ~2 min on this PR via compile-many vs ~5 min serial, so the framework-share seed (FastLED/fbuild#337) does work on small-framework boards. The regression is specific to ESP32-arduino and Teensy-arduino frameworks where stage-2 work dominates.

### What didn't help
- `setup-fbuild` cache hits (PR #334) — the cached binary install is instant
- `zccache` — wraps every TU but the win is shaped wrong for stage-2 of compile-many
- Raising `--framework-jobs` from 1 → 2 — there's only one framework to build at stage 1
- Raising the wrapper timeout 1800 → 3600 — just delays the same failure

### What needs to ship before re-trying this gate flip
Two follow-up fbuild issues filed:
- **FastLED/fbuild#361** — investigate the stage-2 perf regression on big-framework boards; add per-phase timing to compile-many; identify whether the bottleneck is framework re-link, daemon SQLite-pin serialization (FastLED/fbuild#341), per-sketch link contention, or seed not firing
- **FastLED/fbuild#362** — widen the `setup-fbuild` cached path to cover the per-project framework `core/` directory so the cold-key cost is paid once per workflow file change, not once per CI run

Once #361 and #362 ship, we can re-attempt this gate with confidence on all boards.

Issue #2541, opened 2026-05-29
bug(esp32p4): testSimd RPC write-timeout — Serial routes to UART0 instead of USB-Serial-JTAG


zackees:
## Symptom

`bash autoresearch esp32p4 --simd` (with or without `--no-fbuild`) reproducibly fails with:

```
   Connecting to device... ok
   Sending testSimd RPC...
SIMD TEST ERROR: Serial write error: Write timeout
```

This blocks every RPC over `--simd` on ESP32-P4: `testSimd`, `testSimdBenchmark`, and by extension `measureParlio`. Reproducible across PIE-on, PIE-off (#2538), and now `wave8ExpandBenchmark` (#2539). Forced the workaround of capturing boot-time output via `esp_rom_printf` on COM25 (see companion issue) instead of a proper RPC round-trip.

## Root cause (decoded this session)

The AutoResearch P4 build runs with:
```
CDC On Boot       : 0
Arduino USB Mode  : 1
```
With `CDC_ON_BOOT=0`, the Arduino-ESP32 core maps `Serial` to **UART0 (pins 37/38)**, not the USB-Serial-JTAG peripheral. So:

- The firmware's RPC server (which reads from `Serial`) listens on UART0.
- The host tool (`fbuild monitor` / `serial_interface.py`) opens **COM25 = the USB-Serial-JTAG**.
- The tool's writes go to USB-Serial-JTAG; the firmware never receives them → write timeout.

ESP-IDF `[I][esp32-hal-...]` core logs *do* reach COM25 (they go via the ROM console), which is why `esp_rom_printf` works as a one-way capture rail. The boot-time benchmark output from #2539 confirms this empirically.

## Proposed fix

Set both build flags in the P4 PlatformIO env (and the fbuild equivalent):
```
-D ARDUINO_USB_MODE=1
-D ARDUINO_USB_CDC_ON_BOOT=1
```
This routes `Serial` (HardwareSerial → `HWCDCSerial`) to the USB-Serial-JTAG, which is what the host tool talks to. Note from prior session memory: `feat/autoresearch-esp32p4-hwcdc` already attempted this fix but was not pushed.

After the fix:
- `bash autoresearch esp32p4 --simd` should produce the testSimd JSON results
- `measureParlio` works → we can capture **full encode_us before/after** for PRs like #2539
- Boot benchmarks can switch from `esp_rom_printf` back to `FL_PRINT` (companion issue)

## Verification plan

[...]

Issue #2540, opened 2026-05-29
perf(autoresearch): use FL_PRINT instead of esp_rom_printf in on-device benchmarks


zackees:
## Problem

The `AutoResearchWave8Expand.h` boot benchmark added in PR #2539 prints results via `esp_rom_printf`. That was chosen as a workaround so output would reach the USB-Serial-JTAG console (= COM25) without requiring the testSimd RPC, which is currently blocked by a UART0-vs-USB-JTAG routing issue (see companion issue + #2536/#2538). It worked, but it's not the right long-term choice:

- **`FL_PRINT` is much faster than `esp_rom_printf`**. `esp_rom_printf` is a ROM function not optimized for hot paths and lacks the buffering / log-level filtering / timestamping that the rest of the codebase relies on.
- **Mixes two log mechanisms in one file.** The rest of `AutoResearch.ino` (and FastLED broadly) uses `FL_PRINT`/`FL_ERROR`. The benchmark using a different rail is a project-convention divergence that will confuse future maintainers.
- **No log filtering / log level.** `esp_rom_printf` always emits regardless of `CONFIG_LOG_DEFAULT_LEVEL`. Noisier logs, no way to suppress at runtime.

## Proposal

1. Switch every benchmark output in the AutoResearch example (and any future on-device benchmark) to **`FL_PRINT`**.
2. Coupled fix: once the routing block is resolved (see the testSimd RPC issue) `FL_PRINT` reliably reaches the host capture tool, so the `esp_rom_printf` workaround becomes unnecessary.

## Touchpoints

- `examples/AutoResearch/AutoResearchWave8Expand.h` — the only current `esp_rom_printf` user introduced in #2539.
- Anywhere else this pattern gets copy-pasted before the routing fix lands.

## Why
Mentioned by @zackees during PR #2539 review: "we should use FL_PRINT everywhere because it's much much faster."

Refs #2526, #2539.

zackees:
Verified resolved on master: `grep -rn esp_rom_printf examples/AutoResearch/` returns 0 matches. The conversion was done in #2751 (chore(esp32): eliminate runtime esp_rom_printf call sites, merged). Closing as no longer reproducible.

PR #2751, opened 2026-06-04
chore(esp32): eliminate runtime esp_rom_printf call sites

### LINKED ISSUE #2699: chore(esp32): eliminate runtime esp_rom_printf call sites
## Context

`esp_rom_printf` is already discouraged by FastLED's C++ lint checker: it bypasses FastLED logging, blocks on a slow peripheral, and spreads platform-specific output into code that should use FastLED's logging/IO layer. The current source still has suppressed runtime call sites.

Current scan:

```text
rg -n esp_rom_printf\( examples/AutoResearch src/platforms/esp -g *.h -g *.hpp -g *.cpp
```

The runtime call sites are grouped in:

- `examples/AutoResearch/AutoResearchWave8Expand.h` and `examples/AutoResearch/AutoResearchParlioEncode.h`: boot-time benchmark output originally routed through ROM printf so AutoResearch output could reach USB-Serial-JTAG.
- `src/platforms/esp/32/drivers/usb_serial_jtag_esp32_idf.hpp` and `src/platforms/esp/32/detail/io_esp_jtag_or_uart.hpp`: USB-Serial-JTAG bootstrap/fallback diagnostics.
- `src/platforms/esp/32/drivers/lcd_spi/lcd_spi_peripheral_esp.cpp.hpp`: LCD SPI/clockless error diagnostics from IRAM-sensitive paths.

The lint/checker references and tests should remain; this issue is about eliminating runtime source calls rather than removing the checker.

## Proposal

Replace direct runtime `esp_rom_printf` usage with appropriate FastLED-owned mechanisms:

- Route AutoResearch benchmark output through `FL_PRINT`, `fl::print`, or an AutoResearch serial helper once USB-Serial-JTAG routing is reliable.
- Replace USB-Serial-JTAG bootstrap logs with a FastLED-owned platform output abstraction, or make the initialization path report state after the normal logging sink is available.
- For IRAM/error paths such as LCD SPI, record error state/counters in the ISR/IRAM path and emit logs from task/poll context instead of printing directly from the low-level transmit path.
- Remove `// ok esp_rom_printf` suppressions from runtime source files as each call site is replaced.

## Acceptance criteria

- `rg -n esp_rom_printf\( examples/AutoResearch src/platforms/esp -g *.h -g *.hpp -g *.cpp` finds no runtime source call sites.
- `esp_rom_printf` remains allowed only in lint/checker code, tests that verify the checker, or documentation that explains the ban.
- AutoResearch benchmark output remains visible to the host on supported ESP32 USB-Serial-JTAG boards.
- USB-Serial-JTAG bootstrap/fallback diagnostics are still available through FastLED-owned logging/IO or through explicit non-runtime state reporting.
- LCD SPI/clockless failure paths do not log from ISR/IRAM-sensitive code; failures are surfaced from task-safe context.
- The existing `EspRomPrintfChecker` continues to reject new raw `esp_rom_printf` runtime calls.

## Open questions

- Should a very early ESP32 bootstrap path have a dedicated FastLED emergency sink, or should bootstrap diagnostics be deferred until normal logging is initialized?
- Should the lint checker become stricter by disallowing `// ok esp_rom_printf` suppressions in runtime source after this cleanup lands?

## Related issues

- #2540 tracks the narrower AutoResearch benchmark cleanup.
- #2668 captured the broader `fl::print`/USB-Serial-JTAG output contract and explicitly called out avoiding `esp_rom_printf` in favor of FastLED-owned IO.

zackees:
## Summary

Removes every runtime `esp_rom_printf` site under `examples/AutoResearch` and `src/platforms/esp`, replacing each with a FastLED-owned mechanism. The `EspRomPrintfChecker` and its tests are unchanged — only the runtime offenders go.

Acceptance criterion from the issue:
```
rg -n 'esp_rom_printf\(' examples/AutoResearch src/platforms/esp -g '*.h' -g '*.hpp' -g '*.cpp'
```
returns empty on this branch.

## Per-file change

- **`examples/AutoResearch/AutoResearch{Wave8Expand,ParlioEncode}.h`** — delete `printXxxResultRom` helpers entirely. They are dead code; no caller exists anywhere in the repo. The RPC handlers JSON-serialize the result struct directly, and the boot-time ROM-printf bridge was a stopgap for #2540/#2541 that is no longer needed now that #2541 has landed.
- **`src/platforms/esp/32/drivers/usb_serial_jtag_esp32{,_idf.hpp}`** — the 16 driver-bootstrap esp_rom_printf calls ran inside the singleton constructor, where FL_PRINT would re-enter the in-progress singleton. Replaced with a `UsbSerialJtagEsp32::InitOutcome` enum recorded in a member field + `initOutcome()` / `initError()` getters. The two non-bootstrap runtime errors in `write()` / `writeln()` become FL_WARN directly (task context).
- **`src/platforms/esp/32/detail/io_esp_jtag_or_uart.hpp`** — drop the 2 backend-selection esp_rom_printf lines from the EspIO constructor; backend state is already in `mUseUsbSerialJtag`. Adds `reportInitDiagnosticsIfNeeded()` invoked at the top of `print()` / `println()` which emits a one-shot FL_PRINT summary (`EspIO: backend=USB-JTAG usb_jtag_outcome=InstalledOk err=0`) once construction has completed and FL_PRINT is safe.
- **`src/platforms/esp/32/drivers/lcd_spi/lcd_spi_peripheral_esp.{h,cpp.hpp}`** — the 3 esp_rom_printf calls lived in an FL_IRAM function that may be entered from ISR context. `LoggingInIramChecker` forbids FL_WARN inside FL_IRAM functions, so direct replacement is not possible. Latch each failure into `mLast{WaitForSlotTimeout,PanelIoRecreateError,TxColorError}` and drain via FL_WARN from `waitTransmitDone()` — task-only, since `xSemaphoreTake` would configASSERT from ISR.

## Resolution of the two open questions in the issue

The issue listed two unanswered architectural decisions. Per user direction:

1. **USB-Serial JTAG bootstrap diagnostics** → **defer via status getter** (implemented as `UsbSerialJtagEsp32::InitOutcome` + `EspIO::reportInitDiagnosticsIfNeeded`).
2. **Tighten `EspRomPrintfChecker` to forbid `// ok esp_rom_printf` in runtime source** → **out of scope for this PR**, left as a follow-up if marker creep becomes a problem.

## Test plan

- [x] `bash lint` — green (Python + C++ checkers + JS).
- [x] `bash test --cpp` — 309/309 host tests passed.
- [ ] CI: ESP32 platform compile matrix (deferred to PR CI; all changed files are ESP32-gated).
- [x] `rg -n 'esp_rom_printf\(' examples/AutoResearch src/platforms/esp -g '*.h' -g '*.hpp' -g '*.cpp'` returns empty.

Closes #2699

## Summary by CodeRabbit

* **New Features**
  * Exposed richer USB‑Serial JTAG initialization diagnostics for visibility.
  * Benchmarks updated to be RPC-only with results JSON‑serialized by the RPC handler.

* **Bug Fixes**
  * Deferred diagnostics until first print to avoid unsafe interrupt‑context logging.
  * LCD SPI now latches transmit errors in IRAM context and reports them safely from task context.
  * Replaced direct ROM console prints with structured logging.

### DIFF
--- a/examples/AutoResearch/AutoResearchParlioEncode.h
+++ b/examples/AutoResearch/AutoResearchParlioEncode.h
@@ -17,9 +17,9 @@
 ///      indicates the P4 L2 cache fully hides PSRAM latency for this access
 ///      pattern (×1.00 SRAM vs PSRAM); this rechecks on the byte-LUT version.
 ///
-/// Output: `BENCH_PARLIO_*` lines via `esp_rom_printf` -> USB-Serial-JTAG (COM25)
-/// so capture works without depending on the testSimd RPC routing (#2541).
 /// **Invocation:** RPC-only via `parlioEncodeBenchmark` in AutoResearchRemote.cpp.
+/// The result struct is JSON-serialized by the RPC handler; this header no
+/// longer prints directly.
 ///
 /// Current implementation times the production BF1 pipe4 encode path and skips
 /// PSRAM-only placements when PSRAM is not present.
@@ -38,7 +38,6 @@
 #if defined(ARDUINO_ARCH_ESP32) || defined(ESP_PLATFORM)
 #include <Arduino.h>     // micros()
 #include <esp_heap_caps.h>
-#include <esp_rom_sys.h> // esp_rom_printf -> USB-Serial-JTAG console
 #define FL_PARLIO_BENCH_ENABLED 1
 #else
 #define FL_PARLIO_BENCH_ENABLED 0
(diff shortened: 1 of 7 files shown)


zackees:
Addressing CodeRabbit's two nitpicks:

**1. `teardownLocked()` should clear latched errors (lcd_spi)** — Fixed in 311c48453b. Real session-hygiene concern; trivial 3-line fix.

**2. `mDiagnosticsReported` should be atomic (io_esp_jtag_or_uart)** — Acknowledged, not applied. CodeRabbit's own assessment was: *"the current implementation is acceptable. Only consider this refinement if strict once-only semantics are required."*

FastLED's typical runtime is a single-thread `loop()`, and the worst case of a race here is one duplicate `EspIO: backend=…` line at startup — benign. Adding `fl::atomic<bool>` would pull in atomics machinery into the boot path for a diagnostic line that is, by design, fired once per device boot. Leaving as-is.

Issue #2755, opened 2026-06-04
Normalized non-allocating reset/crash classification API for FastLED.watchdog()


zackees:
## Motivation

The unified watchdog API (#2731, landed in #2738 + follow-ups) gives a portable `ResetCause` enum and `WatchdogCrashReport` struct, but anyone who wants a *human-readable* description of what just reset the chip still has to write code like this (taken from `examples/AutoResearch/AutoResearch.ino` lines 374-392):

```cpp
const uint32_t srsr = SRC_SRSR;
Serial.print("[boot] SRC_SRSR = 0x");
Serial.println(srsr, HEX);
if (srsr & SRC_SRSR_WDOG3_RST_B)        Serial.println("[boot]   WDOG3_RST_B (RTWDOG fired)");
if (srsr & SRC_SRSR_WDOG_RST_B)         Serial.println("[boot]   WDOG_RST_B (WDOG1/2 fired)");
if (srsr & SRC_SRSR_IPP_RESET_B)        Serial.println("[boot]   POR (power-on reset)");
if (srsr & SRC_SRSR_LOCKUP_SYSRESETREQ) Serial.println("[boot]   LOCKUP_SYSRESETREQ (HardFault)");
if (srsr & SRC_SRSR_IPP_USER_RESET_B)   Serial.println("[boot]   IPP_USER_RESET_B (external pin)");
if (srsr & SRC_SRSR_JTAG_SW_RST)        Serial.println("[boot]   JTAG/SW reset");
```

That's eight lines of platform-specific bit-twiddling per sketch — not portable, not idiomatic, easy to get the bit positions wrong (which I did get wrong in this very file — the prior version had every shift off by 1–9 bits, see PR #2753 P0 fix).

There should be a **normalized, allocation-free, no-destructor classification API** that handles the cross-platform 90% case in one line, but also gives an escape hatch for the platform-specific bits when a sketch genuinely needs them.

## Design

Three tiers, layered so each tier just composes the previous one. **Zero heap allocation, zero non-trivial destructors anywhere on the hot path**.

### Tier 0 — normalized enum + zero-cost name

The existing `fl::ResetCause` enum gains a paired free function returning a static name through `fl::string_view`:

```cpp
namespace fl {

enum class ResetCause : fl::u8 {
    UNKNOWN, POWER_ON, BROWNOUT, EXTERNAL_PIN,
    WATCHDOG, SOFTWARE, LOCKUP, DEBUGGER, PANIC,
};

[...]

Issue #2731, opened 2026-06-04
Unified cross-platform Watchdog Timer API via FastLED.watchdog() singleton


zackees:
## Motivation

AutoResearch currently bricks itself temporarily when FlexIO/RMT/PARLIO peripherals wedge — the only recovery is a physical button press to enter the bootloader. We already have a partial ESP32-only watchdog (`src/platforms/esp/32/watchdog_esp32.h`, called from `examples/AutoResearch/AutoResearch.ino:315`), and a previous attempt to enable WDOG3 on Teensy 4 caused a reset loop (see `AutoResearch.ino:204-214`). Time to do this properly across the whole supported MCU matrix, behind a single uniform API.

The goal is a **god-object accessible watchdog singleton** — `FastLED.watchdog()` — that:
- Has a small, guaranteed-portable minimum API surface
- Tiered capability flags for advanced features
- Persists a crash counter + safe-mode flag across resets so the board can recover itself instead of looping into the same crash forever

## Per-platform capability matrix (from cross-MCU research)

| Family | Max timeout | Pre-reset IRQ | Window mode | Crash-cause reg | Persist (no power loss) | Persist (power cycle) | Decoded crash report |
|---|---|---|---|---|---|---|---|
| ESP32 / S2 / S3 / C3 / C6 / H2 / P4 | configurable (TWDT) | shutdown handler | no | `esp_reset_reason()` | RTC_NOINIT (8KB classic/S2/S3/C3, 4KB C6/H2, **none on P4**) → fall back to `__NOINIT_ATTR` | NVS partition | Core dump partition (opt-in) |
| Teensy 4.x (iMXRT1062) | 0.5s–128s (`WDT_T4<WDT3>`) | `.trigger` + callback | yes (`.window`) | `SRC_SRSR` | OCRAM2 retained @ `0x2027FF80` | EEPROM emul (4284B T4.1) | **Bundled `CrashReport`** |
| Teensy 3.x (K20/K66/KL26) | ms–days (LPO/bus clk) | `IRQRSTEN` | yes | `RCM_SRS0/1` | custom `.noinit` SRAM section | FlexNVM EEPROM (2KB) | manual |
| RP2040 | **8.388 s** | none (HW) | no | `VREG_AND_CHIP_RESET` + `watchdog_caused_reboot()` | `watchdog_hw->scratch[0..3]` (16B) — **scratch[4..7] reserved by SDK** | flash (LittleFS) | manual via fault handler |
[...]

PR #2756, opened 2026-06-04
feat(watchdog): non-allocating reset/crash classification API + ScopedWatchdog

zackees:
## Summary

Implements the design from #2755 — adds a normalized, **allocation-free, no-destructor** classification API on top of the existing `fl::Watchdog` so users no longer need bit-twiddling on platform-specific reset-cause registers. Also adds a zero-setup `FL_WATCHDOG_AUTOFEED()` macro that drops a RAII guard at the top of `loop()`.

## Public API surface added

```cpp
// Tier 0 — zero-cost static names, backed by string literals in flash
fl::string_view fl::resetCauseName(ResetCause c) noexcept;

// Tier 1 — bundled cause + platform raw register + describe() to caller buffer
struct fl::ResetInfo {
    ResetCause cause;
    fl::u8     subcauseId;
    fl::u32    rawRegister;
    fl::string_view causeName() const noexcept;
    fl::string_view subcauseName() const noexcept;
    fl::size describe(fl::span<char> out, bool verbose = false) const noexcept;
};
fl::ResetInfo fl::Watchdog::lastResetInfo() const noexcept;

// Tier 1.5 — RAII guard for loop()-top use
class fl::ScopedWatchdog {
public:
    ScopedWatchdog() noexcept;                          // 15 000 ms default
    explicit ScopedWatchdog(fl::u32 timeout_ms) noexcept;
    ~ScopedWatchdog() noexcept;
};

// Variadic macro — same name handles both forms
#define FL_WATCHDOG_AUTOFEED(...) \
    ::fl::ScopedWatchdog FL_WATCHDOG__CONCAT(_fl_wdt_, __LINE__) { __VA_ARGS__ }
```

Usage replaces the 8-line bit-twiddle in `AutoResearch.ino` with one line:

```cpp
void loop() {
    FL_WATCHDOG_AUTOFEED();   // feeds on construction, lazily arms on first call,
                              // prints prior reset/crash info + waits 3 s on crash
    doWork();                 // destructor feeds again at end-of-scope
}
```

`ScopedWatchdog` feeds the watchdog on **both** construction and destruction so a loop iteration starts with a fresh deadline AND the next iteration also starts fresh. The first construction per process lazily arms the WDT, prints any prior-boot reset/crash diagnostic via `fl::println`, and pauses 3 s on a crash so the developer can read the message before `loop()` repaints serial.

## Why `fl::span<char>` (caller-owned buffer) and not `fl::string` (allocating)

The user requirement was *"fl::string which doesn't allocate memory, nor destruct the memory."* The closest existing types map cleanly:
- `fl::string_view` — non-owning view, no allocation, no destructor → used for fixed enum names (Tier 0)
- `fl::span<char>` — caller-owned buffer, no allocation, no destructor → used for dynamic content via `describe(out, verbose)` (Tier 1, 2)
- This is the `snprintf`-style pattern: caller controls capacity, method returns bytes written, truncation is safe, zero allocation

## Tests

10 new test cases in `tests/fl/wdt/watchdog.cpp`:
- `fl::resetCauseName` maps every enum value to a non-empty static view
- `describe()` — terse format, verbose format, NUL termination, truncation safety, zero-sized buffer
- `lastResetInfo()` default returns `{lastResetCause(), 0, 0}`
- `ScopedWatchdog` feeds on construction and destruction
- `FL_WATCHDOG_AUTOFEED()` / `FL_WATCHDOG_AUTOFEED(timeout)` both compile

All existing watchdog tests + new tests pass on the host stub.

## Unity-build / build-system wiring

Following the existing `fl/X/` convention:
- `src/fl/wdt/watchdog.cpp.hpp` — implementations (`ResetInfo::describe`, `ScopedWatchdog` ctor/dtor)
- `src/fl/wdt/_build.cpp.hpp` — unity build aggregator + print/delay sinks
- `src/fl/build/fl.wdt+.cpp` — top-level unity TU entry
- `src/fl/_build.cpp.hpp` — wired in the new `fl/wdt/` subdir
- `ci/lint_cpp/test_unity_build.py` — registered `fl.wdt+.cpp` in `EXPECTED_BUILD_FILES`

The dispatcher (`src/platforms/watchdog.impl.cpp.hpp`) is unchanged — platform impls still live in their `.impl.hpp` files. The new TU only adds the platform-agnostic helpers.

## Phase 2 follow-ups (tracked in #2755)

- Per-platform `lastResetInfo()` extraction:
  - Teensy 4 — `SRC_SRSR` bits → `subcauseId` (0=WDOG3, 1=WDOG1/2, 2=LOCKUP, …) + raw register
  - ESP32 — `esp_reset_reason()` already structured; just lift it
  - AVR — `MCUSR` bit → subcauseId
  - … etc per #2755 subcause map
- `WatchdogCrashReport::describe()` for the Tier 2 fault frame (multi-line)
- `CrashReport::platformDetails()` Tier Max escape hatch (returns `void*` to platform-defined struct)

## References

- Design issue: #2755
- Previous watchdog PR (bare-register WDOG3 fix): #2753
- Original API design: #2731

## Summary by CodeRabbit

* **New Features**
  * Expanded watchdog API with richer reset diagnostics (cause, subcause, raw register) and a ResetInfo describe() API
  * Added RAII watchdog guard that auto-feeds, reports recoveries, detects nesting, and exposes active-scope count
  * Added convenience macro to instantiate the automatic watchdog guard

* **Tests**
  * Added comprehensive tests for reset naming, describe() formatting/truncation, auto-feeding semantics, macro usage, and nesting behavior

### DIFF
--- a/ci/lint_cpp/test_unity_build.py
+++ b/ci/lint_cpp/test_unity_build.py
@@ -80,6 +80,7 @@ EXPECTED_BUILD_FILES = {
     "fl/build/fl.test+.cpp",
     "fl/build/fl.ui+.cpp",
     "fl/build/fl.video+.cpp",
+    "fl/build/fl.wdt+.cpp",
     "fl/build/platforms+.cpp",
     "fl/build/third_party+.cpp",
 }
--- a/src/fl/_build.cpp.hpp
+++ b/src/fl/_build.cpp.hpp
@@ -27,3 +27,4 @@
 #include "fl/test/_build.cpp.hpp"
 #include "fl/ui/_build.cpp.hpp"
 #include "fl/video/_build.cpp.hpp"
+#include "fl/wdt/_build.cpp.hpp"
--- a/src/fl/build/fl.wdt+.cpp
+++ b/src/fl/build/fl.wdt+.cpp
@@ -1,0 +1,10 @@
+/// @file fl.wdt+.cpp
+/// @brief Unity build: recursive includes from src/fl/wdt/
+
+#include "platforms/new.h"
+
+// IWYU pragma: begin_keep
+#include "fl/system/arduino.h"
+// IWYU pragma: end_keep
+
+#include "fl/wdt/_build.cpp.hpp"
--- a/src/fl/wdt/_build.cpp.hpp
+++ b/src/fl/wdt/_build.cpp.hpp
@@ -1,0 +1,29 @@
+/// @file _build.cpp.hpp
+/// @brief Unity build header for fl/wdt/ directory.
+///
+/// Includes the platform-agnostic watchdog API helpers (ResetInfo::describe,
+/// ScopedWatchdog ctor/dtor, etc.) plus the per-TU print/delay sink that
+/// ScopedWatchdog's lazy first-init relies on.
+
+#include "fl/log/log.h"
+#include "fl/system/delay.h"
+
+#include "fl/wdt/watchdog.cpp.hpp"
+
+namespace fl {
+namespace platforms {
+
+// Print-sink implementation used by ScopedWatchdog's first-init path. Defined
+// here (in the unity TU) rather than in watchdog.cpp.hpp so that public
+// headers don't pull in fl/log/log.h.
+void scopedWatchdogPrintLine(fl::string_view sv) FL_NOEXCEPT {
+    FL_WARN(sv);
+}
+
+// Wraps fl::delay() — portable across stub/WASM/embedded.
+void scopedWatchdogPause3s() FL_NOEXCEPT {
+    fl::delay(3000);
+}
+
+} // namespace platforms
+} // namespace fl
--- a/src/fl/wdt/watchdog.cpp.hpp
+++ b/src/fl/wdt/watchdog.cpp.hpp
@@ -112,6 +112,17 @@ namespace platforms {
 void scopedWatchdogPrintLine(fl::string_view sv) FL_NOEXCEPT;
 void scopedWatchdogPause3s() FL_NOEXCEPT;
 
+// Counts simultaneously-alive ScopedWatchdog instances. >1 is almost always
+// a programmer error: nested guards mean the lazy-init's timeout argument
+// of the second instance is silently ignored, and timeout semantics across
+// the two scopes become ambiguous. Singleton-style accessor (function-local
+// static) so the counter has well-defined lifetime even if a guard is
+// constructed before main().
+inline int& scopedWatchdogActiveCount() FL_NOEXCEPT {
+    static int count = 0;
+    return count;
+}
+
 // First-init helper, used by ScopedWatchdog's constructor. Runs only once
 // per process. Single-threaded loop() is the canonical caller — no atomics
 // are needed for the guard byte on supported MCUs.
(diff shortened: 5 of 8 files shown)


zackees:
## Update — macro rename + nesting detection

Per review feedback on the macro name, **`FL_WATCHDOG_AUTOFEED` → `FL_WATCHDOG_AUTO`**.

### Why the rename

`FL_WATCHDOG_AUTOFEED` implied the macro just refreshes the WDT. It actually does much more: lazy first-call WDT arm, last-crash diagnostic print, 3-second pause on crash, then feeds on both construction *and* destruction. `FL_WATCHDOG_AUTO` captures "automatic management" instead of just "automatic feed".

### Naming alternatives considered

| Candidate | Family extensibility | Reads like |
|---|---|---|
| **`FL_WATCHDOG_AUTO(timeout)`** ✅ | All watchdog macros stay under one `FL_WATCHDOG_*` prefix — greppable family | "watchdog in auto mode" |
| `FL_AUTO_WATCHDOG(timeout)` | Splits family across `FL_AUTO_*` and `FL_MANUAL_*` | "the automatic watchdog" |
| `FL_WATCHDOG_GUARD(timeout)` | RAII vocab but doesn't say "auto-init" | "guard a scope" |
| `FL_WATCHDOG(timeout)` | Collision risk with capability macros `FL_WATCHDOG_HAS_*` | "watchdog this scope" |
| `FL_LOOP_WATCHDOG(timeout)` | Locks to `loop()` use only | "loop's watchdog" |

`FL_WATCHDOG_AUTO` won on family discoverability — future `FL_WATCHDOG_QUIET`, `FL_WATCHDOG_VERBOSE`, `FL_WATCHDOG_MANUAL` all stay in one grep-able prefix, and it matches FastLED's `FL_<feature>_<modifier>` convention (`FL_TEST_CASE`, `FL_CHECK_EQ`).

### Single-instance enforcement added

Also from the same review feedback: *"what happens if we have two watchdog macros in use? this can be a serious problem, i think that we should prevent this from happening."*

Two simultaneously-alive `ScopedWatchdog` instances are almost always a bug — the second one's `timeout_ms` argument is silently ignored (only the first instance gets to arm the WDT). Detection now happens at two levels:

| Misuse | Detection |
|---|---|
| Two `FL_WATCHDOG_AUTO()` on the **same source line** | Compile error — the local variable name is `_fl_wdt_<__LINE__>`, so duplicates collide |
| Two `FL_WATCHDOG_AUTO()` on different lines but **simultaneously alive** (nested scopes) | Runtime — active-scope counter > 1 triggers a **one-shot** `fl::println` warning, WDT still fed for safety |

New observability accessor `fl::ScopedWatchdog::activeScopeCount()` lets tests + diagnostics inspect the counter directly.

### Diff summary

- `src/fl/wdt/watchdog.h` — rename macro, document single-instance rule, add `activeScopeCount()` declaration
- `src/fl/wdt/watchdog.cpp.hpp` — counter increment/decrement in ctor/dtor, one-shot warn on nesting
- `tests/fl/wdt/watchdog.cpp` — new "nesting detection counter" test validates ctor inc / dtor dec

All existing watchdog tests + new tests pass on the host stub.

Issue #2773, opened 2026-06-05
Meta: ESP32-S3 binary-size reduction — Batch 1 (~26 KB) + Batch 2 (~26–50 KB) (supersedes #2771)


zackees:
## Summary

Authoritative tracking issue for binary-size reduction wins on the
ESP32-S3 NEOPIXEL Blink baseline. Replaces #2771, which had ~95 KB of
phantom savings in its first version because the map-file audit didn't
filter for `address != 0x00000000` (i.e. sections that the linker had
**already dropped** via `--gc-sections` were being counted as if they
were live). See the [closing comment on #2771][2771-correction] for the
methodology fix; this issue carries the corrected numbers as the
canonical reference.

[2771-correction]: https://github.com/FastLED/FastLED/issues/2771#issuecomment-4628404796

**Baseline:** `firmware.bin = 673,504 B` for `examples/Blink/Blink.ino`,
master `511730eb1c`, PlatformIO + IDF 5.4 + xtensa-esp-elf-gcc 14.2.0.
Source analysis: #2473 (top-down breakdown) + its [symbol audit comment][2473-audit].

[2473-audit]: https://github.com/FastLED/FastLED/issues/2473#issuecomment-4628287075

A separate fbuild issue (FastLED/fbuild#417) was filed to prevent future
recurrences of the map-file false-positive trap.

---

## What the linker IS already tree-shaking (don't chase these)

The first round of analysis on #2771 incorrectly listed these as live
bloat. Re-checking against the linked ELF (`xtensa-esp32s3-elf-nm
firmware.elf`) shows they are absent from the binary — the linker
already harvested them and put `0x00000000` placement-address tombstones
in the map file:

| Phantom bloat originally claimed | Reality |
|---|---|
| `mbedtls/x509_crt_bundle` (Mozilla CA bundle), ~68 KB | ❌ tree-shaken — section at addr `0x00000000` |
| `libmesh.a` rodata strings, ~17 KB | ❌ tree-shaken |
| `fl::third_party::huffTable`, ~8.5 KB | ❌ tree-shaken (`nm` returns nothing) |
| `fl::third_party::PLM_AUDIO_SYNTHESIS_WINDOW`, ~2 KB | ❌ tree-shaken |
| `libmbedtls_2.a` TLS strings, ~13 KB | ❌ tree-shaken |

The linker is doing its job. Do **not** open PRs to "save" any of these.

---

## Where tree-shaking genuinely fails (and why)

[...]

Issue #2473, opened 2026-05-12
ESP32-S3 binary size: 657 KB on master vs 344 KB on 3.10.3 — remaining +313 KB delta breakdown


zackees:
## Summary

Follow-up to #2420 / #2421 / #2424 / #2428. After the recent fixes landed (channels-as-default-opt-in driver registry, audio/net/video/fx/gfx eh_frame cleanup, FL_NO_UNWIND via #2423), the ESP32-S3 NEOPIXEL Blink binary is **657,072 B** — that's the structural floor of the new Channel API. The 3.10.3 baseline on the same minimal sketch was **~344 KB**, so we still carry a **+313 KB delta**. This issue tracks where that residual lives so we can decide what (if anything) to chip away at next.

## Measured on

| Item | Value |
|---|---|
| Branch | `feat/remove-legacy-driver-registry-macro` |
| Commit | `4127a0c8da` |
| Sketch | `examples/Blink/Blink.ino` (single NEOPIXEL controller, pin 3) |
| Board | `esp32-s3-devkitc-1` (Espressif ESP32-S3-DevKitC-1-N8, 8 MB flash, no PSRAM) |
| Toolchain | `xtensa-esp-elf-gcc 14.2.0` |
| Build path | PlatformIO 6.1.19 → `pio run -e esp32s3` (fbuild hangs on cold ESP32 builds — see fbuild#117) |
| `firmware.bin` | **657,072 B** |
| `firmware.elf` | 38,516,728 B (with DWARF) |
| `firmware.map` | 28,136,474 B |

## Section sizes (`xtensa-esp32s3-elf-size -A`)

| Section | Bytes | vs reporter's #2420 (`master#ea5d8ec`, 783 KB bin) |
|---|---:|---:|
| `.flash.text` | 260,444 | **−85,456** (was 345,900) |
| `.flash.rodata` | 317,064 | **−42,144** (was 359,208) |
| `.iram0.text` | 64,527 | **−15,012** (was 79,539) |
| `.dram0.data` | 13,584 | −2,532 |
| `.dram0.bss` | 11,368 | — |

## Symbol categories (`xtensa-esp32s3-elf-nm --print-size -C`)

| Category | Bytes | KB | Symbols |
|---|---:|---:|---:|
| **OTHER** (Arduino core, ESP-IDF init, debug, crypto, partition) | 130,315 | 127.3 | 1,834 |
[...]

PR #2789, opened 2026-06-05
feat(log): FASTLED_LOG_VERBOSITY release knob — 33 KB Blink savings (#2773 item 2.3)

zackees:
Implements item 2.3 from the ESP32-S3 binary-size meta tracker (#2773) — the projected single biggest remaining lever, validated on stock Blink at the projected savings range.

## What this adds

A three-level `FASTLED_LOG_VERBOSITY` knob gating `FL_WARN` / `FL_INFO` / `FL_ERROR` / `FL_PRINT` (and their `_IF`/`_ONCE`/`_EVERY`/`_FMT` siblings):

```cpp
#ifndef FASTLED_LOG_VERBOSITY
#if defined(FASTLED_NO_LOG) && FASTLED_NO_LOG
#define FASTLED_LOG_VERBOSITY 0
#else
#define FASTLED_LOG_VERBOSITY 1
#endif
#endif
```

- **`=1` (default, no change vs master)** — full stream formatting on platforms where `SKETCH_HAS_LARGE_MEMORY` allows it
- **`=0`** — each macro expands to `do { if (false) { (void)(fl::sstream_noop() << X); } } while (0)`. User stream expressions stay type-checked but the format strings dead-strip
- **`=2`** — reserved for future use; currently behaves like `=1`

`FASTLED_NO_LOG=1` is honored: it forces level 0 so existing opt-out scripts keep working with the new gate.

## Why this works

#2773's analysis pinpointed `.str1.1` (the merged rodata pool of every `FL_WARN`/`FL_LOG` format string baked into the sketch translation unit) as ~58 KB — by far the single biggest rodata contributor on stock Blink. With `if (false)` guarding the sstream chain, the optimizer drops the whole branch before `--gc-sections` ever sees the literal references — including the `__FILE__` + `__LINE__` + ` WARN: ` + user-message tokens at every call site.

## Empirical savings (ESP32-S3 Blink, PlatformIO + IDF 5.4)

| Build | `firmware.bin` |
|---|---:|
| master baseline (verbosity defaults to 1) | 654,192 B |
| `-DFASTLED_LOG_VERBOSITY=0` | **620,384 B** |

**Net: –33,808 B (–33 KB, –5.2 %).** Right in the meta tracker's 20–40 KB projection range.

## What does NOT change

- **Default behavior** — `FASTLED_LOG_VERBOSITY` unset → level 1 → byte-for-byte identical to current master. Users who haven't opted in see no change in either behavior or binary size.
- **Side-effect semantics in log calls** — `FL_WARN("…" << ++counter)` is not preserved at level 0, consistent with the documented contract of logging macros across the Arduino ecosystem (logging shouldn't have side effects). At level 1 the existing behavior is unchanged.
- **Other log macros** — `FL_DBG`, the `FASTLED_LOG_*_ENABLED` async loggers, and the `FASTLED_DBG`/`FASTLED_WARN`/`FASTLED_INFO`/`FASTLED_ERROR` upper-case variants are NOT gated by this knob (separate concerns / separate users).

## Tests

- [x] `bash test fastled_core --cpp --clean` — 1/1 passes at default verbosity
- [x] `bash compile esp32s3 --examples Blink --platformio` at default → matches master byte-for-byte
- [x] `bash compile esp32s3 --examples Blink --platformio --defines FASTLED_LOG_VERBOSITY=0` → 620,384 B (the 33 KB drop above)

## Cumulative Batch 1 + 2.3 picture

Stacking against the meta tracker:

| Cumulative state | `firmware.bin` |
|---|---:|
| #2473 reported baseline (pre-Batch-1) | 673,504 B |
| post-#2775 (item 1.1) | 673,840 B (net ~0 — see #2773 comment) |
| post-#2778 (item 1.2 follow-up #2782) | 657,680 B |
| **master today (also includes #2786 item 1.4)** | **654,192 B** |
| **this PR (item 2.3) opt-in at level 0** | **620,384 B** |

Refs: #2773 #2473

## Summary by CodeRabbit

* **New Features**
  * Introduced a new build-time logging verbosity configuration to give developers fine-grained control over which log levels are compiled into applications
  * Logging output can be completely disabled at compile time, allowing unused logging code to be stripped from the binary
  * Logging expressions remain type-checked even when logging is disabled, ensuring safer code

### DIFF
--- a/src/fl/log/log.h
+++ b/src/fl/log/log.h
@@ -38,6 +38,43 @@
 #endif
 
 // =============================================================================
+// Release Verbosity Knob (FASTLED_LOG_VERBOSITY) — #2773 item 2.3
+// =============================================================================
+//
+// FastLED's `FL_WARN` / `FL_INFO` / `FL_ERROR` / `FL_PRINT` macros each bake
+// the source file, line number, severity label, and user message string into
+// the binary at every call site. On a stock ESP32-S3 Blink build the merged
+// `.str1.1` rodata pool reaches ~58 KB — by far the single biggest rodata
+// contributor — even when no host is listening on Serial.
+//
+// `FASTLED_LOG_VERBOSITY` lets users explicitly trade diagnostic detail for
+// flash space without flipping the heavier `FASTLED_NO_LOG` switch (which is
+// all-or-nothing and also disables third-party loggers). Levels:
+//
+//   0  — Release: FL_WARN/FL_INFO/FL_ERROR/FL_PRINT expand to type-checked
+//        no-ops (`if (false) sstream_noop() << X`). Format strings and
+//        sstream operator chains disappear; the linker garbage-collects the
+//        merged string pool. Best for shipping firmware where Serial isn't
+//        used at runtime.
+//   1  — Default (current behavior): full stream formatting on platforms
+//        where SKETCH_HAS_LARGE_MEMORY allows it.
+//   2  — Verbose: reserved for future use; currently behaves like 1.
+//
+// Override at build time:
+//   build_flags = -DFASTLED_LOG_VERBOSITY=0
+//
+// When unset, the default is 1 so existing sketches do not change behavior.
+// `FASTLED_NO_LOG` is still honored and forces level 0.
+
+#ifndef FASTLED_LOG_VERBOSITY
+#if defined(FASTLED_NO_LOG) && FASTLED_NO_LOG
+#define FASTLED_LOG_VERBOSITY 0
+#else
+#define FASTLED_LOG_VERBOSITY 1
+#endif
+#endif
+
+// =============================================================================
 // Forward Declarations
 // =============================================================================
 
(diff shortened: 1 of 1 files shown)


zackees:
Closing in favor of #2791, which is strictly better:

- **Also gates `FL_DBG`** (mine only covered `FL_WARN`/`FL_INFO`/`FL_ERROR`/`FL_PRINT`). That extra coverage is the bulk of the +10 KB gap between my measured –33 KB and their measured –43 KB.
- **Cleaner intermediate `FASTLED_LOG_RUNTIME_ENABLED` macro** vs my repeated `SKETCH_HAS_LARGE_MEMORY && FASTLED_LOG_VERBOSITY > 0` condition at every gate site.
- **Pins `FASTLED_TESTING` to verbosity ≥ 1 unconditionally** so host unit tests always get the diagnostic stream regardless of how the build flag is set.

The only thing this PR has that #2791 doesn't is the `if (false) { (void)(fl::sstream_noop() << X); }` form for type-checking expressions at level 0 — marginal benefit, easy to layer on top of #2791 in a follow-up if anyone hits a missed compile error after toggling the knob.

No reason to land two competing implementations.

PR #2844, opened 2026-06-05
fix(net): convert NetworkDetector stub to C++11 nested namespace

zackees:
## Summary

FastLED targets C++11. `src/fl/net/network_detector.h:50` used C++17 nested-namespace syntax (`namespace fl::net {`), which trips `-Werror=c++17-extensions` on strict host builds.

Expand to the C++11 form `namespace fl { namespace net { ... }}` and fold the trailing single-line using-alias block into the same outer `fl` namespace block — net: -3 LOC.

Flagged by the Stage 2a LPC port agent (#2837) as a pre-existing master failure.

## Test plan
- [x] `bash lint --cpp` clean
- [x] Pre-existing test failures (example-wasm DLL loading) unrelated to this change

## Summary by CodeRabbit

* **Refactor**
  * Internal code structure reorganization with no user-facing changes.

### DIFF
--- a/src/fl/net/network_detector.h
+++ b/src/fl/net/network_detector.h
@@ -47,7 +47,8 @@
 
 #include "fl/stl/noexcept.h"
 
-namespace fl::net {
+namespace fl {
+namespace net {
 
 /// @brief Stub NetworkDetector for platforms without the real implementation.
 ///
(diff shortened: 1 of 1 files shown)


zackees:
Closing — the C++17 nested-namespace blocker on master is already resolved (master HEAD has `namespace fl { namespace net { ... }}` form). Likely landed via another PR between this branch being cut and now. The remaining diff would be cosmetic (collapsing two adjacent `namespace fl {}` blocks into one), not worth a separate PR.

PR #2988, opened 2026-06-08
lpc8xx: fix build errors on bare-metal. WIP example for LPC845.

phatpaul:
Trying to get a minimal example working on LPC845 bare-metal (no arduino).

I added a bunch of CMSIS files to the example to get it to compile. Let me know if any files are not needed and I'll remove from the PR.

PR #2989, opened 2026-06-08
lpc: move bare-metal support into platform code

zackees:
## Summary
- move the LPC-specific FastLED runtime/time/IO glue into `src/platforms/arm/lpc` instead of keeping board-support files under an example
- wire LPC into the ARM platform build aggregation and use the shared no-op pin map path
- qualify FastLED's internal yield call as `fl::yield()` so Arduino global `yield()` stubs do not collide
- keep Arduino-owned typedefs and timing out of the bare-metal LPC fallback path

This intentionally does not bring over the misplaced `examples/BlinkLPC845` include tree or generated build-info JSON from external PR #2988. Those fbuild-owned pieces are in the companion PR: FastLED/fbuild#512.

## Validation
- `git diff --check master..HEAD`
- with scratch-only `examples/Blink/platformio.ini`: `..\fbuild\target\debug\fbuild.exe ci --board lpc845 examples\Blink\Blink.ino --lib . --quick -v`
- with scratch-only `examples/Blink/platformio.ini`: `..\fbuild\target\debug\fbuild.exe ci --board lpc804 examples\Blink\Blink.ino --lib . --quick -v`

## Summary by CodeRabbit

* **New Features**
  * Added support for ARM LPC microcontroller platform with timing, I/O operations, and runtime management

* **Bug Fixes**
  * Fixed delay implementation to use correct function namespace resolution

* **Refactor**
  * Enhanced platform architecture to integrate new ARM LPC platform support

### DIFF
--- a/src/FastLED.cpp.hpp
+++ b/src/FastLED.cpp.hpp
@@ -13,6 +13,7 @@
 #include "fl/channels/driver.h"  // for IChannelDriver
 #include "fl/channels/detail/wait_spin_budget.h"  // for tiered-wait spin-budget setters (#2818)
 #include "fl/system/delay.h"  // for delayMicroseconds
+#include "fl/system/yield.h"  // for fl::yield
 #include "fl/system/sketch_macros.h"
 #if SKETCH_HAS_LARGE_MEMORY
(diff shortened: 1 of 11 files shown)


zackees:
@coderabbitai review

PR #2992, opened 2026-06-09
feat: add ISR-backed channel poll wakeups

### LINKED ISSUE #2991: fix(esp32-s3): verify and finish #2420 FastLED.show() timing regression
## Context

#2421 mixed two separate regressions from #2420: ESP32-S3 `FastLED.show()` timing drift and binary-size growth. The binary-size work has its own history and follow-ups; this issue intentionally tracks only the timing regression.

Original timing report from #2420:

- Board: ESP32-S3
- LEDs: 35 WS2812B / NeoPixel LEDs on pin 4
- Pacing: `FastLED.show()` every 5 ms via `millisDelay` from `powerbroker2/SafeString`
- FastLED settings: `FastLED.setMaxRefreshRate(0)`, `setCpuFrequencyMhz(160)`
- Expected sequence time: ~2500 ms
- Observed on `master#ea5d8ec`: commonly ~2600 ms, with random spikes above 3000 ms
- Later retest on `master#787de450d`: 2600-2800 ms remained, although the long-running drift toward 3000 ms was no longer present

Since then, #2815, #2817, #2827, and #2828 landed wait-loop improvements. There is not yet a focused hardware confirmation that the #2420 ESP32-S3 timing case is fixed on current `master`, and current source still has one suspicious gap: `IChannelDriver::waitDone()` exists, but no ESP32 driver overrides it yet.

Binary-size numbers, PlatformIO exception-frame behavior, and lean-build work are out of scope for this issue.

## Reproduction Case

Use the original #2420 sketch and measure one full fade sequence. Prefer serial `micros()`/`millis()` instrumentation over stopwatch timing so the result is machine-readable.

```cpp
#include <Arduino.h>
#include <millisDelay.h>
#include <FastLED.h>

uint8_t i_post_fade = 255;
#define LED_PIN 4
#define FAST_LED_UPDATE_MS 5
#define NUM_LEDS 35 // 35-LED NeoPixel ring
CRGB leds[NUM_LEDS];
uint8_t i_fast_led_delay = FAST_LED_UPDATE_MS;
millisDelay ms_fast_led;
millisDelay ms_delay_post;
uint32_t cycle_start_ms = 0;

void setup() {
  FastLED.addLeds<NEOPIXEL, LED_PIN>(leds, NUM_LEDS).setCorrection(TypicalLEDStrip);
  FastLED.setMaxRefreshRate(0);
  FastLED.show();
  setCpuFrequencyMhz(160);
  Serial.begin(115200);
  ms_delay_post.start(0);
  ms_fast_led.start(i_fast_led_delay);
  cycle_start_ms = millis();
}

void updateLEDs() {
  if (ms_fast_led.justFinished()) {
    FastLED.show();
    ms_fast_led.start(i_fast_led_delay);
  }
}

void loop() {
  if (i_post_fade > 0 && ms_delay_post.justFinished()) {
    const uint8_t i_ring_start = 0;
    const uint8_t i_ring_end = 34;
    const uint8_t i_ring_divisor = 7;
    uint8_t i_ring_counter = ((255 - i_post_fade) / i_ring_divisor) + i_ring_start;

    if (i_ring_counter <= i_ring_end) {
      leds[i_ring_counter] = CRGB::Red;
    }

    if (i_ring_counter - 1 >= i_ring_start && i_ring_counter - 1 <= i_ring_end) {
      leds[i_ring_counter - 1] = CRGB::Black;
    }

    i_post_fade--;

    if (i_post_fade == 0) {
      for (uint8_t i = i_ring_start; i <= i_ring_end; i++) {
        leds[i] = CRGB::Black;
      }

      updateLEDs();
      const uint32_t elapsed = millis() - cycle_start_ms;
      Serial.printf("cycle_ms=%lu\n", (unsigned long)elapsed);
      delay(1000);
      i_post_fade = 255;
      cycle_start_ms = millis();
      ms_delay_post.start(225);
      return;
    } else {
      ms_delay_post.start(5);
    }
  }

  updateLEDs();
}
```

Original PlatformIO setup from #2420:

```ini
[env:esp32s3]
platform = https://github.com/pioarduino/platform-espressif32/releases/download/55.03.38-1/platform-espressif32.zip
board = esp32s3custom
framework = arduino
lib_deps =
  https://github.com/FastLED/FastLED.git#<commit-under-test>
  powerbroker2/SafeString@^4.1.42
build_flags =
  -D "CORE_DEBUG_LEVEL=3"
  -D "LOG_LOCAL_LEVEL=ESP_LOG_INFO"
  -D "FASTLED_WS2812_T1=375"
  -D "FASTLED_WS2812_T2=375"
  -D "FASTLED_WS2812_T3=500"
```

For repository automation, an equivalent `bash autoresearch esp32s3 ...` / `examples/AutoResearch` timing mode would be better than manual PlatformIO testing.

## Likely Code Areas

Current `master` has the #2815 wait-loop scaffolding, but the driver-specific completion wait still appears incomplete:

- `src/fl/channels/manager.cpp.hpp:471-520` has the single-driver fast path. It delegates to `onlyBusy->waitDone(timeoutMs)` when exactly one driver is busy.
- `src/fl/channels/driver.h:151-152` defaults `waitDone()` back to `waitForReady(timeoutMs)`.
- A current tree search only finds the default and the manager call site for `waitDone(`, so ESP32-S3 RMT5 still falls back to the generic wait path instead of an ISR-exact wait.
- `src/fl/channels/driver.cpp.hpp:12-60` and `src/fl/channels/manager.cpp.hpp:375-435` are the generic tiered wait loops: instant check, bounded spin, then `NetworkDetector`-gated yield.
- If `fl::NetworkDetector::isAnyNetworkActive()` returns true unexpectedly, the wait still calls `task::run(250, task::ExecFlags::SYSTEM)`, which reaches `pumpCoroutines(1000)` in `src/fl/task/executor.cpp.hpp:135-140` and `vTaskDelay(ticks)` in `src/platforms/esp/32/coroutine_esp32.impl.hpp:62-87`.
- RMT5 already has an underlying completion primitive: `IRMT5Peripheral::waitAllDone()` / `Rmt5PeripheralESPImpl::waitAllDone()` wraps `rmt_tx_wait_all_done()` (`src/platforms/esp/32/drivers/rmt/rmt_5/rmt5_peripheral_esp.cpp.hpp`). That is the likely primitive for an RMT5 `waitDone()` override.
- `src/platforms/esp/32/drivers/rmt/rmt_5/channel_driver_rmt.cpp.hpp:214-231` currently calls `waitForReady()` before beginning a new transmission. That path should be checked alongside the manager's `onBeginFrame()` / `onEndFrame()` waits.

## Proposal

1. Re-run the #2420 reproduction on current `master` with serial timing output.
2. If current `master` still drifts beyond tolerance, implement the ESP32-S3/RMT5 `waitDone()` override around the existing RMT5 `waitAllDone()` completion primitive.
3. Confirm the no-WiFi/no-BT case avoids the `vTaskDelay(>=1 tick)` path entirely.
4. Keep the #2254 invariant: when WiFi/lwIP/BT is active, FastLED must still yield enough for network tasks.
5. Add a regression test or hardware-harness mode so this does not regress silently.

## Acceptance Criteria

- The #2420 35-LED ESP32-S3 reproduction reports a 2500 ms cycle within a tight tolerance on current `master` after the fix. Target: +/-10 ms if the hardware harness is stable; otherwise document the measured jitter floor and use that as the threshold.
- `FastLED.show()` for a single RMT5 strip with no WiFi/BT active has no visible >=1 FreeRTOS tick floor. For 35 WS2812B LEDs, wall time should be close to the ~1.05 ms bitstream duration plus small overhead.
- RMT5 either overrides `IChannelDriver::waitDone()` with an ISR/IDF completion wait, or the issue documents why the generic tiered wait is sufficient with current measurements.
- WiFi/lwIP behavior from #2254 is not regressed when a network is active.
- Binary-size work remains out of scope.

## Open Questions

- Does the original #2420 sketch still reproduce on current `master` after #2817/#2827/#2828?
- Is the RMT5 `waitAllDone()` wrapper safe to call from the `waitDone()` fast path for every RMT5 channel state, including queued/multi-channel cases?
- Is `ChannelManager::onBeginFrame()->waitForReady()` still necessary for the single RMT5 path, or can readiness be enforced lazily at enqueue/show time without adding frame-start latency?

## Related Issues

- #2420 - original bug report and reproduction
- #2421 - mixed timing + binary-size tracker; closed, but useful for the first timing diagnosis
- #2815 - closed umbrella for replacing the >=1 tick wait-loop behavior
- #2819 - closed Phase 2 issue that added `waitDone()` scaffolding; current tree still needs driver-specific overrides
- #2254 - WiFi/lwIP starvation rationale for deep yields

zackees:
## Summary

Fixes #2991.

This adds `IChannelDriver::PollNeededCallback` and `setPollNeededCallback()` so a driver ISR can wake the channel manager when driver state changed and `poll()` should run again. The callback is intentionally not a done callback: a driver may signal that a chunk, batch, or transfer changed state while more inter-batch work remains on the main thread.

`PollNeededCallback` is now an ISR-safe function/context handle, with `PollNeededCallbackSlot` providing atomic task-to-ISR publication for drivers that invoke the callback from interrupt context. This avoids storing or mutating `fl::function` on the ISR boundary.

`ChannelManager` owns the single poll-needed wait primitive. The manager talks only to `platforms::ChannelPollSignal`, so the channel-manager code has no ESP-specific semaphore guards or drain loops. On ESP32, the platform implementation uses `fl::binary_semaphore`; drivers only receive the callback. If no driver invokes the callback, the manager still wakes on bounded 1 ms timeout slices and polls, so no separate capability gate or per-driver semaphore is needed.

The ESP32 `fl::binary_semaphore` release path is now ISR-aware: it detects ISR context, uses `xSemaphoreGiveFromISR()`, and calls `portYIELD_FROM_ISR()` when a higher-priority task should run. That lets ESP wake the main loop task at the next scheduling opportunity while keeping the channel manager code normalized around a platform wait abstraction.

The PR also adds an AutoResearch `--tight-timing` metric for `FastLED.show(); FastLED.wait()` overhead, enables ESP32-S3 USB CDC for AutoResearch RPC, temporarily raises ESP32-S3/C6 Arduino loop stack size for the deeper AutoResearch driver setup path, and makes the ESP watchdog subscribe/feed the loop task so hung autotest firmware resets instead of wedging the board.

## Verification

- `git diff --check`
- touched-file C++ lint for channel manager/driver changes, platform poll signal dispatch, ESP RMT/I2S/LCD/PARLIO callback changes, ESP semaphore, ESP watchdog, shared atomic, and AutoResearch C++ changes
- `uv run python -m compileall ci/autoresearch/phases.py`
- `uv run pytest ci/tests/test_no_include_after_namespace.py -q`
- `uv run python -m ci.lint_cpp.run_all_checkers src/platforms/esp/32/watchdog_esp32.impl.hpp`
- `uv run python -m compileall ci/boards.py ci/autoresearch/args.py ci/autoresearch/phases.py ci/autoresearch/context.py`
- `uv run python -m ci.autoresearch --help`
- `bash test watchdog`
- `bash compile esp32s3 --examples AutoResearch --platformio`
- `bash compile esp32c6 --examples AutoResearch --platformio`
- Hardware: ESP32-C6 PARLIO tight timing passed, max overhead 1266 us with 2000 us limit
- Hardware: ESP32-C6 RMT tight timing passed, max overhead 296 us with 2000 us limit
- Hardware: ESP32-S3 RMT tight timing passed, max overhead 516 us with 2000 us limit
- Hardware: ESP32-S3 I2S tight timing passed, max overhead 1503 us with 2000 us limit
- Hardware: ESP32-C6 deliberate AutoResearch hang reset via watchdog and resumed RPC after reboot

## Notes

The wait-state bus should already be resolved before this path; `Bus::AUTO` must not be represented in any future bus-mask wait state. If a bus mask is added later, use a resolved bus value only.

The `-DARDUINO_LOOP_STACK_SIZE=16384` override is temporary. Follow-up issue #2993 tracks investigating the ESP AutoResearch stack growth and reverting back to stock Arduino loop stack settings.

## Release Notes

* **New Features**
  * AutoResearch now includes a "Tight Timing Metric" option to measure LED timing overhead and performance characteristics across iterations.

* **Bug Fixes**
  * Enhanced ESP32 watchdog robustness with explicit loop task monitoring and recovery mechanisms.

* **Chores**
  * Updated ESP32 board build configurations for improved USB and memory management.
  * Improved internal driver coordination for better responsiveness and ISR safety.

### DIFF
--- a/ci/autoresearch/args.py
+++ b/ci/autoresearch/args.py
@@ -92,6 +92,11 @@ class Args:
     # See issues #2254 and #2288 (ESP32-S3 SPI second-frame degradation).
     frames: int | None
 
+    # Tight timing metric (#2991)
+    tight_timing: bool
+    tight_timing_iterations: int
+    tight_timing_max_overhead_us: int
+
     @staticmethod
     def parse_args(argv: list[str] | None = None) -> "Args":
         """Parse command-line arguments and return Args dataclass instance."""
@@ -497,6 +502,30 @@ See Also:
             "See issues #2254, #2288.",
         )
 
+        timing_group = parser.add_argument_group(
+            "Tight Timing Metric",
+            "Measure show()+wait() latency against expected LED wire time.",
+        )
+        timing_group.add_argument(
+            "--tight-timing",
+            action="store_true",
+            help="Add the tight timing metric to each runSingleTest response and fail if it exceeds the overhead budget.",
+        )
+        timing_group.add_argument(
+            "--tight-timing-iterations",
+            type=int,
+            default=8,
+            metavar="N",
+            help="Number of steady-state show()+wait() samples for --tight-timing (default: 8).",
+        )
+        timing_group.add_argument(
+            "--tight-timing-max-overhead-us",
+            type=int,
+            default=2000,
+            metavar="US",
+            help="Maximum allowed max(show()+wait()-wire_time) overhead for --tight-timing (default: 2000us).",
+        )
+
         parsed = parser.parse_args(argv)
 
         if parsed.use_fbuild or parsed.no_fbuild:
@@ -554,4 +583,7 @@ See Also:
             parallel=parsed.parallel,
             decode=parsed.decode,
             frames=parsed.frames,
+            tight_timing=parsed.tight_timing,
+            tight_timing_iterations=parsed.tight_timing_iterations,
+            tight_timing_max_overhead_us=parsed.tight_timing_max_overhead_us,
         )
--- a/ci/autoresearch/context.py
+++ b/ci/autoresearch/context.py
@@ -258,6 +258,28 @@ def display_pattern_details(result: dict[str, Any]) -> None:
     print("-" * 62)
 
 
+def display_tight_timing(result: dict[str, Any]) -> None:
+    """Display tight timing metrics from runSingleTest responses."""
+    metric = result.get("tightTiming")
+    if not isinstance(metric, dict):
+        return
+
+    if not metric.get("supported", False):
+        reason = metric.get("message", "unsupported")
+        print(f"   Tight timing: unsupported ({reason})")
+        return
+
+    status = "PASS" if metric.get("passed", False) else "FAIL"
+    max_overhead = metric.get("max_overhead_us", "?")
+    max_allowed = metric.get("max_allowed_overhead_us", "?")
+    expected = metric.get("expected_wire_us", "?")
+    avg_total = metric.get("avg_total_us", "?")
+    print(
+        f"   Tight timing: {status} max overhead {max_overhead}us "
+        f"(limit {max_allowed}us, wire {expected}us, avg total {avg_total}us)"
+    )
+
+
 def print_run_summary(ctx: RunContext) -> None:
     """Print the compact configuration header."""
     print("\u2500" * 60)
--- a/ci/autoresearch/phases.py
+++ b/ci/autoresearch/phases.py
@@ -26,6 +26,7 @@ from ci.autoresearch.context import (
     QuietContext,
     RunContext,
     display_pattern_details,
+    display_tight_timing,
 )
 from ci.autoresearch.gpio import run_gpio_pretest, run_pin_discovery
 from ci.debug_attached import run_cpp_lint
@@ -530,6 +531,21 @@ def _parse_args_and_build_commands(args: Args) -> RunContext | int:
                         frame_count = 1
                     if frame_count > 1:
                         test_config["frameCount"] = frame_count
+                    if args.tight_timing:
+                        if (
+                            args.tight_timing_iterations < 1
+                            or args.tight_timing_iterations > 64
+                        ):
+                            print("Error: --tight-timing-iterations must be in [1, 64]")
+                            return 1
+                        if args.tight_timing_max_overhead_us < 1:
+                            print("Error: --tight-timing-max-overhead-us must be >= 1")
+                            return 1
+                        test_config["tightTiming"] = True
+                        test_config["tightTimingIterations"] = args.tight_timing_iterations
+                        test_config["tightTimingMaxOverheadUs"] = (
+                            args.tight_timing_max_overhead_us
+                        )
                     rpc_command = {"method": "runSingleTest", "params": test_config}
                     rpc_commands_list.append(rpc_command)
 
@@ -1441,6 +1457,8 @@ async def _run_rpc_tests(ctx: RunContext, qctx: QuietContext) -> int:
                             )
                         print(f"   Duration: {duration_str}")
 
+                    display_tight_timing(test_data)
+
                     if "drivers" in test_data and isinstance(
                         test_data["drivers"], list
                     ):
@@ -1474,6 +1492,8 @@ async def _run_rpc_tests(ctx: RunContext, qctx: QuietContext) -> int:
                         print(f"   Lane: {failure.get('lane', 'unknown')}")
                         print(f"   Expected: {failure.get('expected', 'unknown')}")
                         print(f"   Actual: {failure.get('actual', 'unknown')}")
+
+                    display_tight_timing(test_data)
 
                     if "patterns" in test_data:
                         display_pattern_details(test_data)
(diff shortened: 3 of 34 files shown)


zackees:
I still see task sleep in there, it should be fixed. Please see the commit. It's unclear if the wait is happening in the channel driver of the channel manager.

PR #568, opened 2018-02-15
stm32duino port (Maple, Blue Pill etc.)

kolyshkin:
This mostly builds on previous work for SparkCore STM32103 (Cortex M3),
and patches from Evey Quirk @evq.

Unfortunately, stm32duino distribution does not include a header
with definitions for CoreDebug registers and flags etc.
so they were added to a newly created cm3_regs.h from
https://github.com/ARM-software/CMSIS/blob/master/CMSIS/Include/core_cm3.h

Tested on my Blue Pill, works fine

Issue #26, opened 2014-02-10
LPD8806 low-bit handling


kriegsman:
The LPD8806 is a 7-bit-per channel strip, but people program it in 8-bit values.

We need to somehow make the low bit 'visible' so channel values of "1" are not black.

PR #2996, opened 2026-06-11
WIP on LPC845 port

phatpaul:
WIP on LPC845 platform. Still working on the platform, have not even tried the FastLED code yet.  But it builds and runs without hardfault now. Serial doesn't work yet.

BTW, Claude copied the linker script into the example project because we had some issue with path resolution.

### DIFF
--- a/examples/BlinkLPC845Arduino/Blink.ino
+++ b/examples/BlinkLPC845Arduino/Blink.ino
@@ -1,0 +1,26 @@
+/// @file    Blink.ino
+/// @brief   Blink the first LED of an LED strip
+/// @example Blink.ino
+
+#include <Arduino.h>
+
+void setup() {
+    // Test Serial initialization
+    Serial.begin(115200);
+    Serial.println("\n==============================================");
+    Serial.println("   BLINK.INO - Simple LED Blink Example");
+    Serial.println("==============================================\n");
+    Serial.flush();
+}
+
+void loop() {
+    static uint32_t count = 0;
+    count++;
+
+    // Test Serial.print (uses Print::write virtual function)
+    Serial.print("Loop: ");
+    Serial.println(count);
+
+    // Test delay (requires SysTick)
+    delay(1000);
+}
--- a/examples/BlinkLPC845Arduino/SerialTest.ino
+++ b/examples/BlinkLPC845Arduino/SerialTest.ino
@@ -1,0 +1,28 @@
+/// @file    SerialTest.ino
+/// @brief   Minimal serial test for LPC845
+
+#include <Arduino.h>
+
+void setup() {
+    // Initialize serial with a delay to let USB enumerate
+    delay(2000);
+    Serial.begin(115200);
+    delay(500);
+    
+    // Send startup message
+    Serial.println("\n\n========== LPC845 SERIAL TEST ==========");
+    Serial.println("Board: LPC845-BRK");
+    Serial.println("Test: Basic serial communication");
+    Serial.println("========================================\n");
+}
+
+void loop() {
+    static uint32_t counter = 0;
+    
+    Serial.print("Loop #");
+    Serial.print(counter++);
+    Serial.print(" - millis: ");
+    Serial.println(millis());
+    
+    delay(1000);
+}
--- a/examples/BlinkLPC845Arduino/lpc845_flash.ld
+++ b/examples/BlinkLPC845Arduino/lpc845_flash.ld
@@ -1,0 +1,79 @@
+/* Combined linker script for LPC845 - avoids relative INCLUDE issues */
+
+MEMORY
+{
+  FLASH (rx)  : ORIGIN = 0x00000000, LENGTH = 64K
+  RAM   (rwx) : ORIGIN = 0x10000000, LENGTH = 16K
+}
+
+_Min_Heap_Size = 0x400;
+_Min_Stack_Size = 0x800;
+
+/* Stack pointer goes at top of RAM */
+_vStackTop = ORIGIN(RAM) + LENGTH(RAM);
+
+/* Legacy symbol for compatibility */
+_estack = _vStackTop;
+
+ENTRY(Reset_Handler)
+
+SECTIONS
+{
+  .isr_vector :
+  {
+    . = ALIGN(4);
+    KEEP(*(.isr_vector))  /* Vector table from startup C file */
+    . = ALIGN(4);
+  } > FLASH
+
+  .crp 0x000002FC :
+  {
+    KEEP(*(.crp))
+  } > FLASH = 0xFF
+
+  .text :
+  {
+    KEEP(*(.after_vectors))
+    *(.text*)
+    *(.rodata*)
+    KEEP(*(.init))
+    KEEP(*(.fini))
+    . = ALIGN(4);
+    KEEP(*(.preinit_array*))
+    KEEP(*(.init_array*))
+    KEEP(*(.fini_array*))
+    . = ALIGN(4);
+    _etext = .;
+  } > FLASH
+
+  _sidata = LOADADDR(.data);
+
+  .data : AT(_etext)
+  {
+    . = ALIGN(4);
+    _sdata = .;
+    *(.data*)
+    . = ALIGN(4);
+    _edata = .;
+  } > RAM
+
+  .bss (NOLOAD) :
+  {
+    . = ALIGN(4);
+    _sbss = .;
+    *(.bss*)
+    *(COMMON)
+    . = ALIGN(4);
+    _ebss = .;
+  } > RAM
+
+  .heap_stack (NOLOAD) :
+  {
+    . = ALIGN(8);
+    PROVIDE(end = .);
+    PROVIDE(_end = .);
+    . = . + _Min_Heap_Size;
+    . = . + _Min_Stack_Size;
+    . = ALIGN(8);
+  } > RAM
+}
--- a/examples/BlinkLPC845Arduino/platformio.ini
+++ b/examples/BlinkLPC845Arduino/platformio.ini
@@ -1,0 +1,44 @@
+[platformio]
+default_envs = lpc845brk
+
+[env]
+# pinning the platform to my fork for now until the PR is merged, but this can be removed once that happens
+#platform = https://github.com/phatpaul/platform-nxplpc-arduino.git
+# Using local development copies - switch to GitHub URLs for production/CI
+platform = symlink:///home/pabbott/repos/platform-nxplpc-arduino
+# pinning the framework to my fork for now until the PR is merged, but this can be removed once that happens
+#platform_packages = framework-arduino-lpc8xx @ https://github.com/phatpaul/ArduinoCore-LPC8xx.git
+# Using local development copies - switch to GitHub URLs for production/CI
+platform_packages = framework-arduino-lpc8xx @ symlink:///home/pabbott/repos/ArduinoCore-LPC8xx
+
+framework = arduino
+upload_protocol = cmsis-dap
+debug_tool = cmsis-dap
+lib_extra_dirs = ../..
+lib_deps = fastled=symlink://../../
+
+[env:lpc845brk]
+board = lpc845brk
+board_build.ldscript = lpc845_flash.ld
+upload_protocol = cmsis-dap
+debug_tool = cmsis-dap
+debug_init_break = tbreak setup
+build_type = debug
+build_flags =
+    -I${platformio.packages_dir}/framework-arduino-lpc8xx/variants/lpc845brk
+    -I${platformio.packages_dir}/framework-arduino-lpc8xx/variants/lpc845
+    -I${platformio.packages_dir}/framework-arduino-lpc8xx/CMSIS
+    -std=gnu++14
+    -DFASTLED_ALL_PINS_VALID
+    -DFASTLED_M0_USE_C_IMPLEMENTATION
+    -Og
+    -g3
+build_unflags = -std=gnu++11
+
+[env:lpcxpresso804]
+board = lpcxpresso804
+board_build.ldscript = ${platformio.packages_dir}/framework-arduino-lpc8xx/linker_scripts/gcc/lpc804_flash.ld
+
+[env:lpcxpresso845max]
+board = lpcxpresso845max
+board_build.ldscript = ${platformio.packages_dir}/framework-arduino-lpc8xx/linker_scripts/gcc/lpc845_flash.ld
--- a/examples/BlinkLPC845Arduino/src/Blink.ino
+++ b/examples/BlinkLPC845Arduino/src/Blink.ino
@@ -1,0 +1,1 @@
+/home/pabbott/repos/FastLED/examples/Blink/Blink.ino
--- a/examples/BlinkLPC845CMSIS/platformio.ini
+++ b/examples/BlinkLPC845CMSIS/platformio.ini
@@ -1,0 +1,17 @@
+[platformio]
+default_envs = lpc845
+
+
+[env:lpc845]
+platform = nxplpc
+board = lpc845
+framework = cmsis
+upload_protocol = cmsis-dap
+debug_tool = cmsis-dap
+build_type = debug
+build_flags =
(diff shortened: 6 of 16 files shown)


zackees:
Superseded by #2997, which rebases the core `src/` changes from this PR onto current `master`.

Since this PR was opened, `master` has merged most of the LPC port (#2989/#2992/#2995), so #2997 carries only what's still additive:
- the Arduino `Serial` I/O path in `io_lpc.cpp.hpp` (now `#if defined(ARDUINO)`-gated so it doesn't regress the new bare-metal stub path)
- the LPC `pin.h` dispatch to `pin_noop.hpp`
- the `__cxa_pure_virtual` ARDUINO guard in `FastLED.cpp.hpp`

The example sketches were dropped per project policy. Original authorship is preserved via `Co-authored-by`. Please direct further review to #2997 — this one can be closed.

zackees:
@phatpaul — dug into the *"Claude copied the linker script into the example project because we had some issue with path resolution"* note, since we want to fix the root cause rather than carry a copied `.ld` around.

**I'm fairly sure this is a GNU LD `INCLUDE`-resolution issue, not an OS path-length thing.** (From your `platformio.ini` you're on Linux — `/home/pabbott/...` via `symlink://` — so Windows `MAX_PATH` / 260-char limits aren't in play.)

Here's what I found. The stock `lpc845_flash.ld` in the framework (`ArduinoCore-LPC8xx/linker_scripts/gcc/lpc845_flash.ld`) ends with:

```
INCLUDE linker_scripts/gcc/lpc8xx_common.ld
```

That `INCLUDE` path is **relative to the framework root**. The catch is that GNU `ld` resolves `INCLUDE` paths against:
1. the current working directory, and
2. any `-L` search directories,

**not** relative to the directory of the script that contains the `INCLUDE`. When PlatformIO links via `board_build.ldscript = ${platformio.packages_dir}/.../lpc845_flash.ld`, the main script loads fine (absolute path), but the build CWD is `.pio/build/<env>/` and the framework's `linker_scripts/gcc` dir isn't on ld's `-L` path — so `lpc8xx_common.ld` can't be found and the link dies with something like `cannot open linker script file linker_scripts/gcc/lpc8xx_common.ld: No such file or directory`. Your "combined" script (which inlines the common fragment, per its own header comment) sidesteps this entirely.

If that matches what you saw, a couple of cleaner fixes instead of copying the script:
- Add the framework root to the linker search path so the relative `INCLUDE` resolves:
  `build_flags = -Wl,-L${platformio.packages_dir}/framework-arduino-lpc8xx`
- Or, in the framework, change the `INCLUDE` to a bare filename (`INCLUDE lpc8xx_common.ld`) and add `-Wl,-L${platformio.packages_dir}/framework-arduino-lpc8xx/linker_scripts/gcc`.

**Could you confirm a few things so we pick the right fix?**
1. The exact linker error you hit when you *didn't* copy the script (the `cannot open linker script file ...` line).
2. Whether you'd tried referencing it as `board_build.ldscript = ${platformio.packages_dir}/framework-arduino-lpc8xx/linker_scripts/gcc/lpc845_flash.ld` (like your `lpcxpresso845max` env) before copying it locally.
3. Your OS / toolchain (just to be 100% sure it's not platform-specific).

FYI the rebased core changes are now in #2997 (the example sketches + this `.ld` were dropped there), so this is purely about understanding the framework integration for the next iteration.

zackees:
Update: the linker-script path-resolution root cause is now **fixed upstream** in the PlatformIO platform.

- Investigation + repro plan: zackees/platform-nxplpc-arduino#3
- Fix (merged): zackees/platform-nxplpc-arduino#4

The builder now appends the framework directory to the linker search path (`-L`), so GNU ld can resolve the framework scripts' root-relative `INCLUDE linker_scripts/gcc/lpc8xx_common.ld`. @phatpaul — once you pull the updated platform, you should be able to drop the copied/"combined" `lpc845_flash.ld` and point `board_build.ldscript` straight at the stock script:

```ini
board_build.ldscript = ${platformio.packages_dir}/framework-arduino-lpc8xx/linker_scripts/gcc/lpc845_flash.ld
```

(same form your `lpcxpresso845max` env already uses). If you still hit anything, the questions in my earlier comment about the exact error + your toolchain would help us confirm. The rebased FastLED core changes already landed in #2997.

zackees:
Final follow-up: linker-script handling is now **fully baked into the platform backend** — zackees/platform-nxplpc-arduino#5 (merged).

The builder now selects the framework linker script from `build.mcu` automatically, so on the latest platform you don't need `board_build.ldscript` **at all** for the stock boards. A minimal env is enough:

```ini
[env:lpc845brk]
platform = https://github.com/zackees/platform-nxplpc-arduino.git
framework = arduino
board = lpc845brk   ; lpc845_flash.ld selected automatically
```

Verified by building Blink with an empty `ldscript` everywhere:
- `lpc845brk` → links, RAM 16K / Flash 64K (lpc845_flash.ld)
- `lpcxpresso804` → links, RAM 4K / Flash 32K (lpc804_flash.ld)

An explicit `board_build.ldscript` still wins if you ever need a custom layout. No copied/"combined" `.ld` required anymore.

zackees:
@phatpaul — tagging you directly on the note above: you can now drop `board_build.ldscript` entirely on the latest platform. No copied/combined `.ld` needed anymore.

zackees:
I received the board, everything plugs in and works as expected. The auto research agent is engaged and bring up is now being done by opus 4.8.


Bringup schedule:
  1. printf
  2. wdt
  3. led drivers

zackees:
Closing — superseded by #2997 (merged earlier today).

@phatpaul thanks for the WIP; your `src/` work landed via #2997 with original authorship preserved (`Co-authored-by: Paul Abbott <phatpaul@gmail.com>`). What changed in the rebase vs. what you had here:

- **`io_lpc.cpp.hpp`** — your Arduino `Serial`-backed I/O implementation kept, gated under `#if defined(ARDUINO)` so it doesn't regress the bare-metal path that landed in #2989.
- **`pin.h` + `FastLED.cpp.hpp`** — your changes kept verbatim.
- **`led_sysdefs_arm_lpc.h`, `spi_arm_lpc.h`, `clockless_arm_lpc.h`, `m0clockless_c.h`** — most of these edits had already been absorbed via #2989 / #2992 / #2995 in a more complete form, so they came out clean against master.
- **`examples/BlinkLPC845Arduino/` + `examples/BlinkLPC845CMSIS/`** — dropped per project policy (kept out of the example/CI matrix). If you want those upstreamed they fit better in [`zackees/ArduinoCore-LPC8xx`](https://github.com/zackees/ArduinoCore-LPC8xx)`/examples/` or [`zackees/platform-nxplpc-arduino`](https://github.com/zackees/platform-nxplpc-arduino).
- **The `lpc845_flash.ld` workaround you added** — no longer needed. [`platform-nxplpc-arduino#5`](https://github.com/zackees/platform-nxplpc-arduino/pull/5) (merged today) bakes per-MCU linker script defaults into the platform builder, so a minimal env is enough:
  ```ini
  [env:lpc845brk]
  platform = https://github.com/zackees/platform-nxplpc-arduino.git
  framework = arduino
  board = lpc845brk   ; lpc845_flash.ld selected automatically
  ```
  An explicit `board_build.ldscript` still wins if you ever need a custom layout. Verified by building `Blink` with empty `ldscript` everywhere: `lpc845brk` → links, RAM 16K / Flash 64K (`lpc845_flash.ld`); `lpcxpresso804` → links, RAM 4K / Flash 32K (`lpc804_flash.ld`).

## Bringup status on LPC845-BRK

**JSON-RPC bringup works.** The crash you were chasing was three stacked bugs landed today; here's the chain with concrete validation per fix.

1. **Boot-ROM vector checksum** ([fbuild#556](https://github.com/FastLED/fbuild/pull/556)) — the LPC8xx boot ROM only runs an image whose first 8 vector words sum to zero (mod 2³²); the word at offset `0x1C` is reserved for the two's complement. `lpc21isp` patches this during ISP flashing, but raw SWD programmers don't. The linker script now computes `_isr_vector_checksum` and the startup table emits it at `0x1C`, so SWD-flashed images boot without external patching. *Validated:* readelf of `firmware.elf` shows the checksum slot zeros the running sum; image boots from cold via SWD without any post-link patcher.

2. **C++ global constructors weren't running** ([fbuild#569](https://github.com/FastLED/fbuild/pull/569)) — `Reset_Handler` called `SystemInit` then jumped straight to `main`, never invoking `__libc_init_array`, so `.init_array` never ran. Polymorphic globals — notably `fl`'s default `memory_resource` singleton — were left with uninitialised vtable pointers; the first virtual dispatch (the first heap allocation on the JSON-RPC path) loaded a garbage vtable from RAM and HardFaulted, which trapped in the weak `Default_Handler` infinite loop. Plain `Serial.println` / `FL_WARN` survived because they touch no constructor-initialised global. *Fix:* added `bl __libc_init_array` between `SystemInit` and `main`, and provided no-op `_init`/`_fini` weak stubs that `__libc_init_array`/`__libc_fini_array` reference (normally from `crti.o`/`crtn.o`, absent under `-nostartfiles`). Applied to both LPC845 and LPC804 startup. *Validated:* PC samples land in `fl::variant<fl::json>` / `main` / `memcpy` (xPSR IPSR=0, Thread mode, no fault) where they previously pinned in `Default_Handler`.

3. **Heap base inside live `.bss`** ([fbuild#568](https://github.com/FastLED/fbuild/pull/568)) — the LPC845 linker script declared the heap base via `PROVIDE(end = .);`. An LTO partition object emitted its own strong `end` symbol inside `.bss`, which silently suppresses the `PROVIDE` and left the heap base ~340 bytes **below** `_ebss`, overlapping live `.bss` (including newlib `_sbrk`'s own `heap_end` state). The first heap-growing `malloc` then corrupted `.bss` and the core HardFaulted — breaking any C++ that allocated (`fl::vector`, `fl::string`, `fl::json`, the JSON-RPC stack). *Fix:* switched `end` / `_end` / `__heap_start` / `__HeapBase` to forced assignments so the script definition overrides the object symbol and the heap base lands at the true top of `.bss`. *Validated:* heap base now `== _ebss` (`0x10001608` on this build); first `malloc` grows above `.bss` instead of stomping it.

Rolled-in supporting work from the same commit train:
- `HardwareSerial` USART TX/RX implementation + header.
- `operator new` / `new[]` / `delete` backed by `malloc` (`new_delete.cpp`) wired into `ARDUINO_STUB_ASSETS` — then refined upstream in [`ArduinoCore-LPC8xx#25`](https://github.com/zackees/ArduinoCore-LPC8xx/pull/25) (drop `unsigned long` overloads, 32-bit ABI) and [`#26`](https://github.com/zackees/ArduinoCore-LPC8xx/pull/26) (collapse `operator delete` variants to a single `free` thunk).
- `startup_lpc845.S` vector / init fixes.
- `f_cpu` 30 MHz → **24 MHz** to match the actual `FRO_DIRECT` clock that `SystemInit` configures. [`platform-nxplpc-arduino#6`](https://github.com/zackees/platform-nxplpc-arduino/pull/6) carries the matching `board.f_cpu` value for the platform package; without this, every `delay(1000)` was ~20% off.

After all of the above, on the same image:
- Cold boot → constructors run → `loop()` enters the RPC dispatcher.
- Thread mode steady; no `Default_Handler` traps under load.
- `Serial.println` / `FL_WARN` paths still alive (they were the survivors when only #569 was outstanding, so they're a useful regression guard).

## Currently testing

Serial in + out under the JSON-RPC stack on real hardware (the USART path from #568 is in place; now driving it from the application side and confirming RX framing + flow under streaming load). Next bringup steps from here are `printf`, watchdog, then the LED drivers.

Thanks again for the head start — it's why the linker-script work on the platform side got prioritised.

zackees:
@phatpaul — a minimal frameworkless `main.cpp` demo for the LPC8xx bringup matrix, per the convo above. No Arduino layer; entry is `int main()` invoked by `Reset_Handler` after `SystemInit` + `__libc_init_array` (the chain that landed today in [fbuild#569](https://github.com/FastLED/fbuild/pull/569)).

**Heads-up on clockless:** FastLED's WS2812 / SK6812 / etc. clockless drivers (`src/platforms/arm/lpc/clockless_arm_lpc.h`) have **NOT** been ported to the bare-metal LPC8xx path yet — they still key on the Arduino timing macros. So this demo intentionally limits itself to `FastPin<>` + the SysTick-backed `fl::platforms::*` timing (which the bare-metal port already provides via `platform_time_lpc.cpp.hpp`). That's enough to validate the boot path + GPIO + timer without depending on the clockless work.

Once `src/platforms/arm/lpc/io_lpc.cpp.hpp` grows a USART backend on the `!defined(ARDUINO)` side (currently all no-op stubs at lines 18–66), the `fl::println()` calls light up on VCOM automatically — no code change in the sketch.

```cpp
// main.cpp — frameworkless FastLED blink demo for NXP LPC8xx.
//
// Pair with a CMSIS / framework=none build against the LPC8xx
// linker scripts. Entry is `int main()`, invoked by Reset_Handler
// after SystemInit + __libc_init_array (FastLED/fbuild#569).
// No Arduino layer involved.

#include "fl/log/log.h"                          // fl::println
#include "platforms/io.h"                        // fl::platforms::begin
#include "platforms/time_platform.h"             // fl::platforms::delay, millis
#include "platforms/arm/lpc/fastpin_arm_lpc.h"   // fl::FastPin<>

namespace {
// PIO0_15 on the LPC845-BRK / LPCXpresso845-MAX breakout header.
// Wire an external LED + series resistor here.
// (FastPin only covers PORT0 today, so the on-board RGB LED on
//  PORT1 isn't reachable yet — see fastpin_arm_lpc.h.)
constexpr unsigned LED_PIN  = 15;
constexpr unsigned BLINK_MS = 250;   // 2 Hz toggle
}  // namespace

extern "C" int main(void) {
    fl::platforms::begin(115200);
    fl::println("[fastled-lpc] boot — frameworkless main()");

    fl::FastPin<LED_PIN>::setOutput();
    fl::println("[fastled-lpc] GPIO ready");

    fl::u32 toggles = 0;
    fl::u32 last_log_ms = fl::platforms::millis();

    for (;;) {
        fl::FastPin<LED_PIN>::hi();
        fl::platforms::delay(BLINK_MS);
        fl::FastPin<LED_PIN>::lo();
        fl::platforms::delay(BLINK_MS);
        ++toggles;

        const fl::u32 now = fl::platforms::millis();
        if ((now - last_log_ms) >= 2000) {
            fl::println("[fastled-lpc] heartbeat");
            last_log_ms = now;
        }
    }
    return 0;
}
```

## What this exercises (bare-metal)
- `Reset_Handler` → `SystemInit` → `__libc_init_array` → `main()` — proves the [#569](https://github.com/FastLED/fbuild/pull/569) startup chain.
- `fl::FastPin<15>::setOutput / hi / lo` — proves the modern-LPC GPIO controller mapping (`DIRSET[0] / SET[0] / CLR[0]` at `0xA0000000`) from `fastpin_arm_lpc.h`.
- `fl::platforms::millis()` / `delay()` — proves the SysTick init + ISR path from `platform_time_lpc.cpp.hpp` (lazy `init_systick`, `g_millis` ticking out of `SysTick_Handler`).
- `fl::println("…")` — currently a no-op via the bare-metal `io_lpc.cpp.hpp` stub; will start emitting the moment that file's USART backend lands.

## What it deliberately does NOT exercise
- `FastLED.addLeds<WS2812, …>` / `FastLED.show()` — would pull in `clockless_arm_lpc.h`, which needs the Arduino-time port noted above.
- `Arduino.h`, `Serial.*`, `setup()` / `loop()` — none of these are referenced anywhere in the sketch.
- `examples/Blink/Blink.ino` — that example assumes Arduino; this is its bare-metal sibling.

If you want a follow-up demo wiring the WS2812 path, the porting work is `clockless_arm_lpc.h` swapping its time source from the Arduino macros to `fl::platforms::micros()` / a tight `__NOP()` loop calibrated against `F_CPU=24000000` — that can land independently and this `main.cpp` picks it up via one line: `FastLED.addLeds<WS2812, 15, GRB>(leds, NUM)`.

Issue #3000, opened 2026-06-13
Rename SKETCH_HAS_*_MEMORY macros to FL_PLATFORM_HAS_*_MEMORY


zackees:
## Summary

The memory-tier classification macros are named `SKETCH_HAS_*_MEMORY`, but they describe **platform** memory capacity, not anything about a user sketch. The name is misleading. Rename the family to `FL_PLATFORM_HAS_*_MEMORY` (preferred) or `FL_HAS_*_MEMORY`, while keeping backward-compatible aliases for external user code.

## Macros to rename

Defined in `src/fl/system/sketch_macros.h`:

| Current | Proposed |
|---------|----------|
| `SKETCH_HAS_TINY_MEMORY`  | `FL_PLATFORM_HAS_TINY_MEMORY` |
| `SKETCH_HAS_LARGE_MEMORY` | `FL_PLATFORM_HAS_LARGE_MEMORY` |
| `SKETCH_HAS_HUGE_MEMORY`  | `FL_PLATFORM_HAS_HUGE_MEMORY` |
| `SKETCH_HAS_TINY_MEMORY_OVERRIDDEN`  | `FL_PLATFORM_HAS_TINY_MEMORY_OVERRIDDEN` |
| `SKETCH_HAS_LARGE_MEMORY_OVERRIDDEN` | `FL_PLATFORM_HAS_LARGE_MEMORY_OVERRIDDEN` |
| `SKETCH_HAS_HUGE_MEMORY_OVERRIDDEN`  | `FL_PLATFORM_HAS_HUGE_MEMORY_OVERRIDDEN` |

Existing backward-compat aliases that also reference these:
- `SKETCH_HAS_LOTS_OF_MEMORY` -> `SKETCH_HAS_LARGE_MEMORY`
- `SKETCH_HAS_VERY_LARGE_MEMORY` -> `SKETCH_HAS_HUGE_MEMORY`

## Naming decision

Recommend `FL_PLATFORM_HAS_LARGE_MEMORY` over `FL_HAS_LARGE_MEMORY` because the tier is a property of the target platform/MCU, and the `FL_PLATFORM_` prefix matches other platform-detection macros. Open to `FL_HAS_*` if shorter is preferred — needs a maintainer call before implementation.

## Backward compatibility (required)

[...]

PR #3027, opened 2026-06-14
feat(json): gate float out of variant on FL_JSON_HAS_FLOAT=0 (#3022 phase 1)

### LINKED ISSUE #3022: fl::json: custom 64-bit packed tagged-number type (replaces variant<i64, float, double> in #3002)
## Motivation

Spinning out of #3002 (drop libgcc soft-FP bloat in `fl::json` on Cortex-M0+ / no-FPU). The current proposal in #3002 is `fl::variant<fixed-point types>` for the numeric slot of the JSON value variant. After more thought that design wastes the 64-bit storage budget:

- `fl::variant<...>` carries a separate `u8` discriminator outside the payload bytes, so the *effective* JSON-value slot is `max(alternative_size) + 1B tag + padding` — minimum 9 B, usually padded to 16 B on the alignment of the largest alternative.
- All existing `fl::fixed_point<>` types are **32-bit only** (`s0x32`, `s16x16`, `u24x8`, ...). To represent the integer ranges JSON-RPC traffic in (microsecond timestamps, LED counts, file offsets, money in fixed-point cents) we need *at least* 60 effective bits of payload.
- `fl::variant<i64, fl::sfixed_integer<...>, fl::ufixed_integer<...>>` would either truncate the integer range (each alternative is 32 b) or add even more discriminator + padding overhead.

## Proposal

A purpose-built **packed 64-bit tagged number** that lives in *exactly one* 64-bit slot of `json_value::variant_t`, replacing both the `i64` and `float` alternatives currently there:

```cpp
namespace fl {

struct json_number {
    // 64 bits total.
    //   Bits [63:61]  tag  (3 bits, 8 representations)
    //   Bits [60:0]   payload (61 bits)
    fl::u64 raw;

    enum class Tag : fl::u8 {
        SIGNED_INT      = 0,   // payload = sign-extended 61-bit signed integer
        UNSIGNED_INT    = 1,   // payload = 61-bit unsigned integer
        FIXED_Q30_31    = 2,   // 1 sign bit + 30 int bits + 30 frac bits  (covers ~±10^9 with ~10^-9 resolution)
        FIXED_Q15_46    = 3,   // 1 sign bit + 14 int bits + 46 frac bits  (covers ±16 384, sub-ppm fractional precision — ratios / brightness)
        FIXED_UQ32_29   = 4,   // 32 int bits + 29 frac bits (unsigned)    (covers timestamps with sub-ns fractions if needed)
        // Reserved tags 5..7 for future expansion (e.g. signed bigint, IEEE-half on host, NaN sentinel for `error` numbers from parse failures, ...).
        RESERVED_5      = 5,
        RESERVED_6      = 6,
        RESERVED_7      = 7,
    };

    constexpr Tag tag() const FL_NOEXCEPT { return static_cast<Tag>(raw >> 61); }
    constexpr fl::u64 payload_bits() const FL_NOEXCEPT { return raw & 0x1FFF'FFFF'FFFF'FFFFULL; }

    // Construction (only constexpr-safe int paths; fixed-point paths must
    // call the explicit factory functions below to make the loss-of-precision
    // intent visible at the call site).
    static constexpr json_number from_i64(fl::i64 v) FL_NOEXCEPT;
    static constexpr json_number from_u64(fl::u64 v) FL_NOEXCEPT;
    static json_number from_q30_31(fl::i32 int_part, fl::u32 frac_part) FL_NOEXCEPT;
    static json_number from_double_lossy(double v) FL_NOEXCEPT;   // host only — selects best Tag
    static json_number from_float_lossy(float v) FL_NOEXCEPT;     // host only — same

    // Extraction (always defined; integer extraction saturates on FP tags).
    constexpr fl::i64    to_i64() const FL_NOEXCEPT;
    constexpr fl::u64    to_u64() const FL_NOEXCEPT;
    // The to_float / to_double paths are gated behind FL_PLATFORM_HAS_LARGE_MEMORY
    // — on Low-memory targets they compile-error to keep the soft-FP cascade
    // permanently absent from the link. Callers that need a float on those
    // platforms must explicitly use the fixed-point Q-format extractors
    // below (`to_q30_31()` etc.) so the loss-of-precision intent is visible.
    float   to_float()  const FL_NOEXCEPT;
    double  to_double() const FL_NOEXCEPT;

    // Q-format extractors (always available — pure integer math).
    fl::pair<fl::i32, fl::u32> to_q30_31() const FL_NOEXCEPT;
    // ... one per Q-format Tag.
};

}  // namespace fl
```

## Why this fits the `fl::json` use case

1. **No IEEE 754 on the wire.** Every conversion between tags is integer arithmetic on `raw` (sign-extend, bit-shift, mask). The soft-FP cascade we identified during LPC845 bring-up (`__aeabi_dadd`/`__aeabi_dmul` chains via `_floatdisf.o` / `_fixunssfdi.o`) is structurally absent from this path.
2. **JSON-RPC numbers fit comfortably.** Protocol IDs, LED counts, GPIO indices, millis timestamps, money/cents/ppm fractions — every realistic JSON-RPC payload type maps cleanly to one of the four Tags above. The `SIGNED_INT` / `UNSIGNED_INT` tags alone (61 bits) cover the full int64 range with one bit of headroom.
3. **One slot in `json_value::variant_t` instead of two.** Today the variant carries `i64` and `float` as separate alternatives, eating two variant-tag positions and forcing every type-conversion visitor to enumerate both. After this lands the variant carries one `json_number` alternative; the type-conversion visitors fan out across Tags inline (one switch on `n.tag()`) instead of across variant alternatives.
4. **The 8-bit variant discriminator stops growing.** `fl::variant` discriminates between alternatives with a `u8` tag; we're removing one alternative (`float`) and replacing the `i64` slot with `json_number` — net minus-one alternative, free space for future variant alternatives (e.g. `fl::vector<u16>` audio buckets if those land).
5. **Host stays on `double` via the template parameter from #3002.** When `FL_PLATFORM_HAS_LARGE_MEMORY=1`, the `json_value`'s number slot can stay on `double` so the host IEEE 754 round-trip tests do not lose precision. The template choice gates `json_number` to `double` at compile time.

## Why this is NOT covered by existing `fl::fixed_point<>`

The existing `fl::fixed_point<IntBits, FracBits, Sign>` templates (`src/fl/math/fixed_point/`) are deliberately **32-bit only** — see `src/fl/math/fixed_point.h:18-20`:

> Available types (IntBits.FracBits):
>   Signed:   s0x32, s4x12, s8x8, s8x24, s12x4, s16x16, s24x8
>   Unsigned: u0x32, u4x12, u8x8, u8x24, u12x4, u16x16, u24x8

`json_number` needs >=60 effective payload bits to represent int64 ranges. The right move is a **new sibling type** specific to the JSON number slot, not a 64-bit extension of `fl::fixed_point<>` (which has its own carefully-tuned SIMD / promotion / saturation behavior that does not transfer to a tagged-union representation).

## Out of scope

- General-purpose use outside `fl::json`. If `json_number` proves useful elsewhere it can promote to a public utility later — but the design constraint here is "fits one 64-bit slot of `fl::json`'s variant", not "is the canonical FastLED 64-bit number".
- IEEE 754 storage. If a host-side caller really needs full `double` precision, they use the existing host config (`FL_PLATFORM_HAS_LARGE_MEMORY=1`, `json_value` stays on `double`).
- NaN / Inf representation in the embedded path. JSON itself does not allow NaN / Inf in the standard, and the embedded path is fixed-point; any host-side `from_double_lossy(NaN)` must throw a parse error (or map to a `RESERVED_*` sentinel — design decision deferred).

## Acceptance criteria

1. `sizeof(fl::json_number) == 8` and `alignof(fl::json_number) == 8` on every supported platform.
2. Construction + extraction round-trips for every Tag (host unit tests under `tests/fl/stl/json/json_number.cpp`):
   - i64 -> from_i64 -> to_i64: bitwise identity within the 61-bit range, saturating outside.
   - u64 -> from_u64 -> to_u64: same.
   - q30_31 -> from_q30_31 -> to_q30_31: bitwise identity.
   - q15_46 / uq32_29: same.
3. `to_double()` / `to_float()` symbol-strip check: when `FL_PLATFORM_HAS_LARGE_MEMORY=0` and a TU references either method, the build fails at compile time with a static_assert pointing the caller at the Q-format extractors. (Prevents the host-only path from accidentally linking into an embedded target.)
4. `fl::json`'s variant_t is updated: `i64` and `float` alternatives are removed, replaced by a single `fl::json_number` alternative (on embedded). Host config keeps `double` separately.
5. The JSON parser's number path uses hand-rolled integer parsers feeding `json_number::from_*`, never `strtod` / `strtof`.
6. The full #3002 success criterion #1 (zero `__aeabi_d*` / `__aeabi_f*` symbols in the LPC845 ELF) passes via `arm-none-eabi-nm -C firmware.elf | grep -E '(aeabi_d|aeabi_f|__[gle]q?[ds]f|scalbn|lround|fmod|atan|acos)'` producing empty output.

## Related

- #3002 — parent issue (`fl::json`: replace double with templated fixed-point variant). This issue is the *concrete number-type design* that lands inside #3002.
- #3000 — `FL_PLATFORM_HAS_*_MEMORY` canonical names (already merged via #3019); used as the compile-time switch for `to_double()`/`to_float()` gating above.
- `src/fl/math/fixed_point/` — existing 32-bit fixed-point templates; sibling, not replaced.

Filed via Claude Code per the request in the design discussion on #3002.

zackees:
## Summary

Phase 1 of #3022. Eliminates ~13 KB of libgcc soft-FP cascade from the LPC845 / no-FPU ARM `fl::json` code path by gating the `float` and `fl::vector<float>` alternatives out of `json_value::variant_t` when `FL_JSON_HAS_FLOAT=0` (defaults to `FL_PLATFORM_HAS_LARGE_MEMORY`, the canonical name from #3019 / #3000).

Closes #3022 (phase 1 — the structural gate). The `fl::json_number` packed-tagged-union design from the issue body lands in phase 2; see the in-flight design-clarification comment for the parser-truncation placeholder.

## Verification (LPC845-BRK, `examples/AutoResearchLpc/AutoResearchLpc.ino`)

| Metric | Before | After | Δ |
|--------|--------|-------|---|
| `text` | 64 204 B | 51 280 B | **−12 924 B** |
| Flash usage | 64.20 / 64.00 KB (~98 %) | 50.44 / 64.00 KB (78.8 %) | **+13.6 KB margin** |
| `__aeabi_d*` / `__aeabi_f*` symbols | dadd, dmul, d2f, fadd, fsub, fmul, fdiv, fcmp*, f2iz, __eqsf2 | **none** | — |

Verification command (now documented in `CLAUDE.md`):

```
fbuild build -e lpc845brk
arm-none-eabi-nm -C .fbuild/build/lpc845brk/release/firmware.elf \
  | grep -E '(aeabi_d|aeabi_f|__[gle]q?[ds]f|scalbn|lround|fmod|atan|acos)'
# (empty output — goal met)
```

Or via fbuild's wrapper that auto-resolves the cross `nm`/`c++filt` from `build_info.json`:

```
fbuild symbols .fbuild/build/lpc845brk/release/firmware.elf
```

## What changed

- **`FL_JSON_HAS_FLOAT`** macro in `fl/stl/json/types.h`, defaulting to `FL_PLATFORM_HAS_LARGE_MEMORY`. Users can force `-DFL_JSON_HAS_FLOAT=1` on a Low-memory part to re-enable floats (paying the soft-FP cost).
- **Variant gating:** `float` and `fl::vector<float>` drop out of `variant_t` on `FL_JSON_HAS_FLOAT=0`. All `visit_fn<float, Visitor>` instantiations vanish, breaking the chain that pulled the unused `operator()(const float&)` visitor overloads (and their FP arithmetic) into the link.
- **Parser gating** in `fl/stl/json.cpp.hpp`: `JsonTokenizer::scan_array_lookahead`, `JsonBuilder::parse_float_array`, `JsonToken::ARRAY_FLOAT`, `JsonToken::NUMBER` (float path), `canBeRepresentedAsFloat(double)`, `optimize_array`'s `ALL_FLOATS` arm, `classify_array`'s `has_float` check. `fl::parseFloat` is no longer reachable.
- **Public constructors preserved.** `json(float)` / `json(double)` / `operator=` / `set(key, float)` stay available; on `FL_JSON_HAS_FLOAT=0` the library takes over the conversion by truncating to `i64` storage. Cost: a single `__aeabi_f2iz` / `__aeabi_d2iz` (~64 B) helper at the call site, gc-sectioned away when not used.
- **`as_float()` / `as_double()` bodies unchanged.** With `float` gone from the variant, the visitor naturally dispatches through the `i64` arm via `static_cast<FloatType>(value)`. Sketches that actually call `as_float()` pay one `__aeabi_l2f` helper; sketches that don't (like `AutoResearchLpc.ino`) pay zero.
- **ScreenMap:** `ParseJson` / `toJson` use float coordinates and a float diameter, so they extend the existing `FASTLED_NO_JSON` early-bail to also bail on `!FL_JSON_HAS_FLOAT`. A 16 KB-SRAM part has no business deserializing 2D map metadata.
- **CLAUDE.md** gains a short pointer to `fbuild symbols <elf>` and the `build_info.json` toolchain-path schema, so future agents stop hunting for `arm-none-eabi-nm` under PlatformIO package dirs.

## Test plan

- [x] `fbuild build -e lpc845brk` succeeds on `examples/AutoResearchLpc/AutoResearchLpc.ino`
- [x] `arm-none-eabi-nm -C firmware.elf | grep -E '(aeabi_d|aeabi_f|...)'` emits **zero matches** (was 14 symbols before)
- [x] Flash usage drops 64.20 KB → 50.44 KB (≥ 13.6 KB margin, exceeds the issue's `≥ 4 KB margin` acceptance criterion)
- [x] `bash test --cpp` JSON suite passes on host (host keeps `FL_JSON_HAS_FLOAT=1`)
- [x] `bash lint --cpp` clean

## Phase 2 (separate PR)

- `fl::json_number` packed 64-bit tagged union with `Q30.31` / `Q15.46` / `UQ32.29` Tags per the issue body
- Replace the placeholder integer-prefix-only parse for fractional JSON numbers (currently truncates `"3.14"` to `3` — flagged with a `TODO(#3022 phase 2)` comment) with a Q-format integer parse so the library preserves fractional precision without ever invoking IEEE-754
- `to_q30_31()` / `to_qN_M()` extractors
- `sizeof(fl::json_number) == 8 && alignof(fl::json_number) == 8` round-trip tests

Closes #3022

## Summary by CodeRabbit

* **Documentation**
  * Added a new “Key Commands” guide for inspecting already-built embedded ELF files with `fbuild symbols`, including example usage and where cross-tool paths are recorded in build info outputs.

* **Chores**
  * JSON floating-point support is now conditionally enabled via `FL_JSON_HAS_FLOAT`, affecting JSON parsing/serialization and UI components that rely on float-backed JSON types.

* **Bug Fixes**
  * Builds without float JSON support now consistently use the same early-return/warn behavior for screen map JSON parsing/serialization as other float-disabled configurations.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -34,6 +34,15 @@
 - `bash profile <function>` — Performance profiling
 - `bash bloat <board>` — Per-symbol flash/RAM bloat report (see `agents/docs/binary-size-analysis.md`)
 
+**Inspecting an already-built embedded ELF (e.g. for soft-FP / libgcc / symbol audits):**
+
+Do NOT spend time hunting for `arm-none-eabi-nm` / `xtensa-esp32s3-elf-objdump` / etc. across PlatformIO package dirs. fbuild already records the right cross-tools for the board.
+
+- `fbuild symbols <PATH-to-elf-or-build-dir>` — runs `nm --print-size --size-sort -S | c++filt` with the board's correct cross-tools, auto-locating `build_info.json` for the toolchain path resolution. Add `--top N` for a longer per-archive / per-symbol table; `--json <path>` for machine output.
+- Toolchain paths live in `build_info.json` (or `build_info_<env>.json`) under: `nm_path`, `cppfilt_path`, `objdump_path`, `readelf_path`, `size_path`, `objcopy_path`, `ar_path`, `cc_path`, `cxx_path`. Read these instead of guessing the toolchain version under `~/.platformio/packages/` or `~/.fbuild/prod/cache/toolchains/`.
+- `bash compile <board> --build-info <out.json>` emits the `build_info.json` next to the build artifact even if one is not already present.
+- LPC845 example: after `fbuild build -e lpc845brk`, run `fbuild symbols .fbuild/build/lpc845brk/release/firmware.elf` — it picks up the correct `arm-none-eabi-nm` from `build_info.json` automatically.
+
 **NEVER use:** `uv run python test.py` — use `bash test` or `uv run test.py`
 **FORBIDDEN:** `--no-fingerprint` (use `bash test --clean`), bare `pio`/`platformio`, bare `meson`/`ninja`/`clang++`
 
--- a/src/fl/math/screenmap.cpp.hpp
+++ b/src/fl/math/screenmap.cpp.hpp
@@ -98,7 +98,12 @@ ScreenMap ScreenMap::DefaultStrip(int numLeds, float cm_between_leds,
 bool ScreenMap::ParseJson(const char *jsonStrScreenMap,
                           fl::flat_map<string, ScreenMap> *segmentMaps, string *err) {
 
-#if FASTLED_NO_JSON
+#if FASTLED_NO_JSON || !FL_JSON_HAS_FLOAT
+    // FL_JSON_HAS_FLOAT==0 (FastLED #3022): the `fl::json` variant carries
+    // no float alternative on Low-memory targets. ScreenMap deserializes
+    // float coordinates and a float diameter, so it has no business
+    // running on a build that intentionally excised the IEEE-754 path —
+    // refuse the same way FASTLED_NO_JSON does.
     FL_UNUSED(jsonStrScreenMap);
     FL_UNUSED(segmentMaps);
     FL_UNUSED(err);
@@ -258,7 +263,11 @@ bool ScreenMap::ParseJson(const char *jsonStrScreenMap,
 void ScreenMap::toJson(const fl::flat_map<string, ScreenMap> &segmentMaps,
                        fl::json *doc) {
 
-#if FASTLED_NO_JSON
+#if FASTLED_NO_JSON || !FL_JSON_HAS_FLOAT
+    // See ParseJson() above — FL_JSON_HAS_FLOAT==0 (FastLED #3022) has no
+    // float alternative in `fl::json::variant_t`, so writing float
+    // coordinates / diameter through the json API is structurally
+    // unavailable.
     FL_WARN("ScreenMap::toJson called with FASTLED_NO_JSON");
     return;
 #else
--- a/src/fl/stl/json.cpp.hpp
+++ b/src/fl/stl/json.cpp.hpp
@@ -35,34 +35,42 @@ namespace fl {
 
 
 // Helper function to check if a double can be reasonably represented as a float
-// Used for debug logging - may appear unused in release builds
+// Used for debug logging - may appear unused in release builds.
+//
+// Gated behind FL_JSON_HAS_FLOAT — the body uses `double` arithmetic which
+// pulls in `__aeabi_dadd` / `__aeabi_dmul` (~3 KB of soft-FP on no-FPU
+// ARM). On FL_JSON_HAS_FLOAT==0 (LPC845 etc.) the variant never carries a
+// float alternative, so this helper has no remaining callers; omitting it
+// keeps libgcc's soft-double cascade out of the link. See FastLED #3022.
+#if FL_JSON_HAS_FLOAT
 FL_DISABLE_WARNING_PUSH
 FL_DISABLE_WARNING(unused-function)
 static bool canBeRepresentedAsFloat(double value) {
 
-    
+
     auto isnan = [](double value) -> bool {
         return value != value;
     };
-    
+
     // Check for special values
     if (isnan(value)) {
         return true; // These can be represented as float
     }
-    
+
     // Check if the value is within reasonable float range
     // Reject values that are clearly beyond float precision (beyond 2^24 for integers)
     // or outside the float range
     if (fl::abs(value) > 16777216.0) { // 2^24 - beyond which floats lose integer precision
         return false;
     }
-    
+
     // For values within reasonable range, allow conversion even with minor precision loss
     // This handles cases like 300000.14159 which should be convertible to float
     // even though it loses some precision
     return true;
 }
-FL_DISABLE_WARNING_POP    
+FL_DISABLE_WARNING_POP
+#endif  // FL_JSON_HAS_FLOAT
 
 
 json_value& get_null_json_value() {
@@ -218,12 +226,26 @@ private:
 
                 if (is_float) {
                     has_float = true;
+#if FL_JSON_HAS_FLOAT
                     // Parse float value to check if it's beyond integer precision
                     // Note: fl::parseFloat may have precision loss for very large numbers
                     float f_val = fl::parseFloat(&mInput[num_start], mPos - num_start);
                     if (fl::abs(f_val) > 16777216.0f) {
                         has_float_beyond_precision = true;
                     }
+#else
+                    // FL_JSON_HAS_FLOAT==0: parser still detects fractional
+                    // syntax (for the type-classification logic below) but
+                    // never calls `fl::parseFloat`, since that helper would
+                    // pull in libgcc's soft-FP cascade. The corresponding
+                    // ARRAY_FLOAT branch in `JsonBuilder` is also gated out,
+                    // so this routes to the LBRACKET slow path where each
+                    // number flows through the phase-1 placeholder
+                    // integer-prefix parse (see `JsonToken::NUMBER` below
+                    // for the long-term Q-format plan tracked in FastLED
+                    // #3022).
+                    has_float_beyond_precision = true;  // forces LBRACKET slow path
+#endif
                 } else {
                     has_int = true;
                     // Parse actual value to get accurate range
@@ -676,6 +698,7 @@ ArrayType classify_array(const json_array& arr) {
                 all_numeric = false;
                 break;
             }
+#if FL_JSON_HAS_FLOAT
         } else if (elem->is_float()) {
             has_float = true;
             auto val = elem->as_float();
@@ -688,6 +711,7 @@ ArrayType classify_array(const json_array& arr) {
             if (fl::abs(f) > 16777216.0f) {
                 has_float_beyond_precision = true;
             }
+#endif
         } else {
             all_numeric = false;
             break;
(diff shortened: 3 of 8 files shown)


zackees:
Note from review: #3029 tracks the gap exposed by this PR — the FastLED-internal float→i64 truncation fallback in `json.h` (`json(float)` / `json(double)` / `operator=` / `set(key, float)`) violates the mandate that the library must not pull in FP ops on Low-memory targets. Same for the UI-control `.cpp.hpp` files (slider, number_field, json_console) which call `json::set(key, float)` from library code. Phase 1.5 PR will remove the truncation fallback (hard compile error pointing callers at `fl::fixed_point<>`) and gate the UI-control TUs the same way ScreenMap is already gated here. Manual `as_fixed_point<>()` / `set(key, fixed_point)` API is the destination — design sketched in #3029 §B.

zackees:
Superseded by phase-2 work: integer-only IEEE 754 codec replacement.

The bit-twiddling codec (`fl::ieee754_string` — string ↔ IEEE 754 bits via integer math, plus integer-only `int ↔ float_bits` helpers in the JSON visitors) removes the libgcc soft-FP cascade structurally — without dropping the `float` or `fl::vector<float>` alternatives from `json_value::variant_t`. On a Low-memory target the LPC845 link still shows zero `__aeabi_d*` / `__aeabi_f*` symbols, just as this PR's gate achieved, but JSON I/O for `fl::vector<float>` keeps working (round-trips via the integer-only codec).

Closing in favor of the codec PR. The design conclusion from #3029 — that `fl::vector<float>` *storage* in the variant is structurally soft-FP-free; only the parser / serializer / scalar conversion routines pull libgcc — is what motivated the move from gating to codec replacement.

Related: #3022 (parent), #3029 (gap-tracking).

Issue #3208, opened 2026-06-17
[meta] LPC / LPC8xx current burn-down


zackees:
﻿?## Purpose

Current burn-down tracker for open LPC / LPC8xx work in FastLED. This issue intentionally excludes unrelated search hits and adjacent non-LPC metas unless they directly block LPC closure.

## Open LPC issue inventory

| Issue | Status | Scope | Next action |
|---|---|---|---|
| #3002 | open | JSON numeric representation / soft-FP bloat on embedded; LPC8xx was one motivating target | Re-read with #3022/#3029; close if superseded, otherwise finish remaining fixed-point replacement work. |
| #3022 | open | Packed tagged-number replacement for `variant<i64, float, double>` | Re-read current implementation; verify whether #3038/#3027 already delivered enough to close or whether code remains. |
| #3029 | closed | Remaining FastLED-internal float callers / missing fixed-point extractor API leaking soft-FP into low-memory links | Fixed by #3211; merged 2026-06-18. JSON/fixed-point conversion now uses integer IEEE bit construction on low-memory paths; issue closed with upgrade note. |
| #3039 | closed | On-device LPC845 verification of `fl::ieee754_string` codec | Fixed by #3213; merged 2026-06-18. LPC845-BRK `--ieee754` run passed 35 checks / 0 failures on COM10; issue closed with upgrade note. |
| #3046 | closed | Late CodeRabbit finding sweep for LPC bring-up PRs #3041-#3045 | Re-checked all referenced PR review/comment surfaces; no CodeRabbit follow-up remained. Closed completed. |
| #3102 | open | LPC845-BRK hardware bench validation meta: deploy, VCOM sidestep, SCT-RX phases | Execute bench checklist, file/fix failures, then close. |
| #3128 | closed | Retire `FASTLED_LPC_RX_SCT_WS2812` manual opt-in; derive from platform predicate | Fixed by #3209; merged 2026-06-17. |

## Working order

1. Close administrative/no-code items first: #3046.
2. Resolve low-risk code cleanup: #3128.
3. Re-read and collapse JSON/soft-FP cluster: #3002, #3022, #3029, #3039.
4. Run the hardware-heavy #3102 bench last, using any fixes from the earlier items.

## Definition of done

- [ ] Every issue listed above is either closed with evidence or has a merged fix PR linked from it.
- [ ] LPC845-BRK AutoResearch still builds, flashes, avoids HardFault, and returns `echo` over COM10 after the final code changes.
[...]

PR #3211, opened 2026-06-18
fix(json): use integer float conversion for low-memory JSON

### LINKED ISSUE #3029: fl::json: FastLED-internal float callers + missing manual fixed-point extractor API leak soft-FP into Low-memory link (gap from #3022 phase 1 / #3027)
## Context

#3027 (phase 1 of #3022) lands the structural gate that drops `float` / `fl::vector<float>` from `fl::json_value::variant_t` on `FL_JSON_HAS_FLOAT=0`. That eliminates the soft-FP cascade from `JsonTokenizer::parse`, `JsonBuilder::on_token`, `parseFloat`, and all `variant::visit_fn<float, ...>` instantiations — verified on `lpc845brk` AutoResearchLpc.ino: zero `__aeabi_d*` / `__aeabi_f*` symbols, flash 64.20 KB → 50.44 KB (~13.6 KB saved).

But during PR review the principle was tightened: **the FastLED library must never pull in floating-point ops on Low-memory targets, regardless of which sketch is linked**. The current PR violates that in two ways:

1. **`json(float)` / `json(double)` / `operator=(float)` / `set(key, float)` truncate-to-i64 fallback** in `src/fl/stl/json.h` invokes `static_cast<i64>(f)` → pulls `__aeabi_f2iz` / `__aeabi_d2iz` (~64 B each) into the link at every call site, even when the user only ever wanted integer storage. The library is doing the float conversion itself; the user did not opt in.

2. **FastLED-internal UI controls call `json::set(key, float)` from library code** — that is, the soft-FP pull is anchored *inside the library*, not inside the user's sketch. The library hits the FP path even when the user's sketch is pure integer JSON-RPC, as long as the link pulls one of the offending TUs in.

The mandate per the design discussion: "If the user uses float that's their problem, FastLED lib should not pull in floating ops when memory is low. If the sketch does then that's beyond our control."

### A. FastLED-internal callers that pass `float` into `fl::json`

| File | Line(s) | Pattern |
|------|---------|---------|
| `src/platforms/shared/ui/json/slider.h` | 20–21, 27–31, 40 | `JsonSliderImpl` API takes float for value/min/max/step. |
| `src/platforms/shared/ui/json/slider.cpp.hpp` | 18–21, 41–46, 52, 62–65 | Stores `float mMin/mMax/mValue/mStep`; calls `json.set("value", mValue)` etc. and consumes `json | 0.0f`. |
| `src/platforms/shared/ui/json/number_field.h` | 20, 26–27, 35, 37–40 | Same pattern as slider: float in API + storage. |
| `src/platforms/shared/ui/json/number_field.cpp.hpp` | 30–32 | `json.set("value"/"min"/"max", float)`. |
| `src/platforms/shared/ui/json/json_console.cpp.hpp` | 190–199, 216, 263 | `strtod` + float storage + `doc.set(idStr, value)` with a float. |
| `src/fl/math/screenmap.cpp.hpp` | 209, 290–304 | `as_float()` for diameter; `fl::json(static_cast<double>(coord))` per element. Already gated in #3027 behind `!FL_JSON_HAS_FLOAT` — works as a model for the other consumers. |

### B. Missing manual fixed-point ↔ json API

`fl::fixed_point<IntBits, FracBits, Sign>` already exists at `src/fl/math/fixed_point/` (32-bit Q-format types: `s0x32`, `s16x16`, `u24x8`, etc.) but has **no `fl::json` integration**. Today there is no way to round-trip a fractional JSON number on a Low-memory target other than through `as_float()` (which dispatches the variant's `i64` arm via `static_cast<float>(value)` and pulls `__aeabi_l2f` at the call site).

To honor the mandate, `fl::json` needs these manual extractor / setter overloads:

```cpp
// In fl::json_value (extractors — pure-integer, no FP):
template<int IntBits, int FracBits, typename Sign>
fl::optional<fl::fixed_point<IntBits, FracBits, Sign>> as_fixed_point() const FL_NOEXCEPT;

// Q-format pair-style accessors per the #3022 design body:
fl::pair<fl::i32, fl::u32> to_q30_31() const FL_NOEXCEPT;
fl::pair<fl::i32, fl::u32> to_q15_46() const FL_NOEXCEPT;
fl::pair<fl::u32, fl::u32> to_uq32_29() const FL_NOEXCEPT;

// In fl::json (setter overloads that store as packed i64, no FP):
template<int IntBits, int FracBits, typename Sign>
json& operator=(fl::fixed_point<IntBits, FracBits, Sign> v) FL_NOEXCEPT;

template<int IntBits, int FracBits, typename Sign>
void set(const fl::string& key, fl::fixed_point<IntBits, FracBits, Sign> v) FL_NOEXCEPT;
```

These should land alongside the `fl::json_number` packed-tagged-union design from the parent issue (#3022), since the storage layout is the same (61-bit payload + 3-bit tag in one `i64` variant slot).

### C. Phase-1 design defect: `json(float)` / `json(double)` truncation fallback in #3027

The fallback added in #3027 `src/fl/stl/json.h`:

```cpp
#else  // FL_JSON_HAS_FLOAT == 0
    json(float f) FL_NOEXCEPT : mValue(fl::make_shared<json_value>(static_cast<i64>(f))) {}
    json(double d) FL_NOEXCEPT : mValue(fl::make_shared<json_value>(static_cast<i64>(d))) {}
#endif
```

does the conversion *inside the library*. The user did not ask for truncation; the library decided. `__aeabi_f2iz` / `__aeabi_d2iz` get pulled in at every call site. Per the mandate this should be a hard compile error directing the caller to the manual fixed-point setter — or, equivalently, the overload should not exist.

Same applies to:
- `json& operator=(float)` / `json& operator=(double)` in `json.h`
- `void set(const fl::string& key, float)` / `void set(..., double)` in `json.h`

## Proposal — "phase 1.5" PR (closes this issue)

1. **Remove the truncation fallback in `json.h`.** Gate `json(float/double)`, `operator=(float/double)`, `set(key, float/double)` behind `#if FL_JSON_HAS_FLOAT` only. No `#else` branch. Caller code that tries to pass a float on a Low-memory build hits a compile error that directs them to `fl::fixed_point<>` + the new `set(key, fixed_point)` overload (B above). The error message should name the migration path.

2. **Gate FastLED-internal UI-control `.cpp.hpp` files behind `#if FL_JSON_HAS_FLOAT`.** `slider.cpp.hpp`, `number_field.cpp.hpp`, the float branch of `json_console.cpp.hpp`. On `FL_JSON_HAS_FLOAT=0` the entire UI-slider TU drops out of the library archive — sketches that link nothing else from `platforms/shared/ui/json/` see zero impact, sketches that *try* to use a slider fail at link with `undefined reference`, mirroring the existing `FASTLED_NO_JSON` ergonomics. ScreenMap already follows this pattern in #3027.

   *Future / phase 2:* add fixed-point variants — `JsonSliderImplFixed<Q16x16>` etc. — so on `FL_JSON_HAS_FLOAT=0` users still have a working slider, just one whose value type is `fl::s16x16` instead of `float`. Tracked separately under #3022 once the json_number Q-format extractors land.

3. **Land the json↔fixed_point extractor / setter API (B above)** as a sibling PR or bundled into the phase-2 `fl::json_number` PR for #3022. The extractor / setter design is independent of the packed-tag union; it can land first and the packed-tag union promotes to use it later.

## Verification (for the closing PR)

- `bash compile lpc845brk --examples AutoResearchLpc` succeeds with `FL_JSON_HAS_FLOAT=0` (same as today).
- `arm-none-eabi-nm -C firmware.elf | grep -E '(aeabi_d|aeabi_f|f2iz|d2iz|__l2f|__floatdisf)'` stays empty (currently passes the `aeabi_d` / `aeabi_f` check but `__aeabi_f2iz` / `__aeabi_d2iz` are reachable from the truncation fallback when any caller links).
- A pure-integer LPC sketch that does not use UI controls or ScreenMap continues to compile (this is the `AutoResearchLpc.ino` baseline).
- A pure-integer LPC sketch that *does* try to write `json["x"] = 3.14f` fails at compile time with a clear error message pointing to `fl::s16x16` + the manual setter.

## Related

- #3022 (parent — `fl::json_number` packed type)
- #3027 (in-flight phase 1 — structural variant gate; introduces the truncation fallback this issue is asking to remove)
- #3019 / #3000 (`FL_PLATFORM_HAS_*_MEMORY` canonical names; the `FL_JSON_HAS_FLOAT` default tracks these)
- `src/fl/math/fixed_point/` (existing 32-bit Q-format types — destination for the manual extractor / setter API)

🤖 Filed via [Claude Code](https://claude.com/claude-code) as the gap-tracking issue spun out of #3027 review.

zackees:
﻿Fixes #3029.
Updates #3208.

## Summary
- Keep `fl::json` float/double entry points available on low-memory targets while avoiding floating-point arithmetic in the conversion path.
- Convert 64-bit `double` inputs to JSON's single-precision float storage with integer IEEE-754 bit handling, including NaN/inf preservation and overflow saturation to infinity.
- Convert FastLED fixed-point values to JSON float storage with integer bit construction, including saturation when the fixed-point magnitude is too large for single precision.
- Keep direct `float` inputs as existing float storage, avoid unnecessary `float -> double` promotion in `ScreenMap::toJson`, and continue gating JSON UI implementation files that do real floating-point parsing/arithmetic behind `FL_JSON_FLOAT_ENABLED`.

## Validation
- `bash test --cpp tests/fl/stl/json.cpp`
- `bash compile uno --examples Blink --backend fbuild`
- `bash compile lpc845brk --examples AutoResearch --backend fbuild` passed: flash 62.47 KB / 64 KB, RAM 10.99 KB / 16 KB.

## Upgrade Notes
No user-facing upgrade is required for existing JSON float callers. Low-memory builds should continue to use the same `fl::json(float)`, `fl::json(double)`, and fixed-point JSON APIs; the conversion is now handled without depending on target floating-point conversion helpers.

## Release Notes

* **New Features**
  * Added fixed-point numeric type support for JSON values, including constructors, setters, and `as_fixed_point<FP>()` extraction.

* **Improvements**
  * Enhanced JSON float handling with IEEE-754 bit-accurate conversion.
  * Updated JSON geometry (v2) to emit `x`/`y`, `diameter`, and an integer `version`.
  * UI JSON number fields and sliders now compile only when float JSON support is enabled.

* **Bug Fixes**
  * Adjusted JSON-related warning behavior for builds without JSON support.

* **Tests**
  * Added regression coverage for fixed-point round-trips and IEEE-754 conversions.

### DIFF
--- a/src/fl/math/screenmap.cpp.hpp
+++ b/src/fl/math/screenmap.cpp.hpp
@@ -370,7 +370,7 @@ void ScreenMap::toJson(const fl::flat_map<string, ScreenMap> &segmentMaps,
                        fl::json *doc) {
 
 #if FASTLED_NO_JSON
-    FL_WARN_F("ScreenMap::toJson called with FASTLED_NO_JSON");
+    FL_WARN("ScreenMap::toJson called with FASTLED_NO_JSON");
     return;
 #else
     if (!doc) {
@@ -412,11 +412,11 @@ void ScreenMap::toJson(const fl::flat_map<string, ScreenMap> &segmentMaps,
 
         fl::json xArray = fl::json::array();
         for (u16 i = 0; i < segment.getLength(); i++) {
-            xArray.push_back(fl::json(static_cast<double>(segment[i].x)));
+            xArray.push_back(fl::json(segment[i].x));
         }
         fl::json yArray = fl::json::array();
         for (u16 i = 0; i < segment.getLength(); i++) {
-            yArray.push_back(fl::json(static_cast<double>(segment[i].y)));
+            yArray.push_back(fl::json(segment[i].y));
         }
 
         fl::json groupObj = fl::json::object();
@@ -429,13 +429,13 @@ void ScreenMap::toJson(const fl::flat_map<string, ScreenMap> &segmentMaps,
         segmentObj.set("group", fl::json(fl::string(name)));
         segmentObj.set("x", xArray);
         segmentObj.set("y", yArray);
-        segmentObj.set("diameter", fl::json(static_cast<double>(diameter)));
+        segmentObj.set("diameter", fl::json(diameter));
         segmentsArr.push_back(segmentObj);
 
         idx++;
     }
 
-    doc->set("version", fl::json(static_cast<double>(2)));
+    doc->set("version", fl::json(i64(2)));
     doc->set("groups", groupsObj);
     doc->set("segments", segmentsArr);
 
--- a/src/fl/stl/json.h
+++ b/src/fl/stl/json.h
@@ -123,9 +123,142 @@
 
 // Implementation details - users should not rely on these directly
 #include "fl/stl/json/types.h"
+#include "fl/stl/bit_cast.h"
 #include "fl/stl/noexcept.h"
+#include "fl/stl/static_assert.h"
 
 namespace fl {
+
+namespace detail {
+
+template <typename T, typename = void>
+struct is_fl_fixed_point : fl::false_type {};
+
+template <typename T>
+struct is_fl_fixed_point<T, decltype(
+    static_cast<void>(T::INT_BITS),
+    static_cast<void>(T::FRAC_BITS),
+    static_cast<void>(fl::declval<const T&>().raw()),
+    static_cast<void>(T::from_raw(fl::declval<const T&>().raw())),
+    void()
+)> : fl::true_type {};
+
+inline u32 round_shift_right_u64(u64 value, int shift) FL_NOEXCEPT {
+    if (shift <= 0) {
+        return static_cast<u32>(value);
+    }
+    if (shift >= 64) {
+        return 0;
+    }
+    const u64 half = u64(1) << (shift - 1);
+    const u64 mask = (u64(1) << shift) - 1;
+    const u64 truncated = value >> shift;
+    const u64 remainder = value & mask;
+    const bool round_up = remainder > half || (remainder == half && (truncated & 1));
+    return static_cast<u32>(truncated + (round_up ? 1 : 0));
+}
+
+inline u32 ieee754_bits_from_scaled_u64(
+    u32 sign, u64 magnitude, int binary_exp) FL_NOEXCEPT {
+    if (magnitude == 0) {
+        return sign;
+    }
+
+    const int top_bit = json_floor_log2_u64(magnitude);
+    int exponent = top_bit + binary_exp;
+    if (exponent > 127) {
+        return sign | 0x7F800000u;
+    }
+
+    if (exponent >= -126) {
+        const int shift = top_bit - 23;
+        u32 significand = 0;
+        if (shift > 0) {
+            significand = round_shift_right_u64(magnitude, shift);
+        } else {
+            significand = static_cast<u32>(magnitude << -shift);
+        }
+        if (significand >= 0x01000000u) {
+            significand >>= 1;
+            ++exponent;
+            if (exponent > 127) {
+                return sign | 0x7F800000u;
+            }
+        }
+        return sign | (static_cast<u32>(exponent + 127) << 23) |
+               (significand & 0x007FFFFFu);
+    }
+
+    const int subnormal_shift = -(binary_exp + 149);
+    u32 mantissa = 0;
+    if (subnormal_shift > 0) {
+        mantissa = round_shift_right_u64(magnitude, subnormal_shift);
+    } else {
+        const int left_shift = -subnormal_shift;
+        if (left_shift >= 64 || magnitude > (u64(-1) >> left_shift)) {
+            return sign | 0x7F800000u;
+        }
+        const u64 shifted = magnitude << left_shift;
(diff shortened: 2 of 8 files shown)


zackees:
﻿Latest update on the remaining CodeRabbit requested-change threads:

- The implementation-heavy IEEE conversion helpers were moved out of `types.h` into `types.impl.hpp` in `02e6a65`.
- The float/double API gating and raw fixed-point storage comments are superseded by the updated requirement for this PR: keep `fl::json(float)`, `fl::json(double)`, and fixed-point JSON entry points available on low-memory targets, but route fixed-point/double conversion through integer IEEE bit construction instead of target floating-point conversion helpers.
- Added coverage for oversized scaled fixed-point magnitude saturation and double NaN/overflow conversion.

Local validation for `02e6a65`:
- `bash test --cpp tests/fl/stl/json.cpp`
- `bash compile uno --examples Blink --backend fbuild`
- `bash compile lpc845brk --examples AutoResearch --backend fbuild`

Issue #3218, opened 2026-06-18
refactor(platform/lpc): replace sparse GPIO register padding with safer offset-based accessors


zackees:
Follow-up to #3212.

## Background

PR #3212 fixes a real bug in `src/platforms/arm/lpc/fastpin_arm_lpc.h`: the fallback `FL_LPC_GPIO_Type` mirror did not include the large holes in the modern LPC GPIO register map, so fields such as `W`, `DIR`, `PIN`, `SET`, `CLR`, `NOT`, `DIRSET`, and `DIRCLR` landed at the wrong byte offsets.

That PR corrects the immediate issue by adding explicit padding so the fallback mirror matches the NXP/CMSIS `GPIO_Type` layout:

- `B` at `0x0000`
- `W` at `0x1000`
- `DIR` at `0x2000`
- `PIN` at `0x2100`
- `SET` at `0x2200`
- `CLR` at `0x2280`
- `NOT` at `0x2300`
- `DIRSET` at `0x2380`
- `DIRCLR` at `0x2400`
- `DIRNOT` at `0x2480`

The padding does not allocate SRAM because this is a typed view over a fixed hardware address, not an instantiated object. However, it is visually bulky and easy to get wrong again.

## Better fix

Replace the sparse fallback struct with one of these safer representations:

1. Direct offset accessors/constants for the small register subset FastLED actually uses, e.g. `PIN`, `SET`, `CLR`, `NOT`, `DIRSET`, `DIRCLR`; or
2. Keep the CMSIS-style struct mirror, but add compile-time `offsetof` assertions for every register FastLED touches.

Direct offsets are probably the cleaner long-term fit for FastLED because the LPC FastPin code only needs a handful of registers and the controller map has large sparse gaps.

## Acceptance criteria

- [ ] FastPin LPC fallback no longer depends on unverified sparse padding arithmetic, or every used padded-field offset is protected by `static_assert(offsetof(...))`.
- [ ] Register offsets used by `FastPin<>::setOutput`, `setInput`, `hi`, `lo`, `toggle`, `hival`, `loval`, `port`, `sport`, and `cport` are covered.
- [ ] The implementation still defers to the SDK/CMSIS `LPC_GPIO` / `GPIO` definition when available, or documents why the fallback address constants are authoritative.
- [ ] A host-side compile/layout test covers the fallback path if practical.
[...]

Issue #3124, opened 2026-06-16
chore(autoresearch): rename perf + test flags to consistent --perf-XX / --test-XX prefixes


zackees:
Naming convention housekeeping for the `bash autoresearch` flag surface.

## What

All performance-related autoresearch flags should use the prefix `--perf-XX`, and all test-related flags should use `--test-XX`. Currently the just-added `--wave2d-perf` (suffix-style) doesn't match this; further perf and test flags added in the future should land under the canonical prefix from day one.

## Proposed renames

- `--wave2d-perf` → `--perf-wave2d`
- *(other perf flags, when they land)* → `--perf-XX`
- *(other test flags, when they land)* → `--test-XX`

## Scope

- `ci/autoresearch/args.py` — flag declarations + help strings.
- `ci/autoresearch/phases.py` — dispatch checks.
- `ci/tests/test_autoresearch_phases.py` — `_make_args` field name + tests.
- Any docs that reference the old names (`agents/docs/hardware-autoresearch.md` etc.).

## Why deferred from #3122 A1

Caught while wiring up the initial `--wave2d-perf` flag in A1 of meta #3122. Renaming retroactively before the flag ships avoids a deprecation cycle; doing it now in a separate focused PR keeps A1's diff scoped to the actual feature.

## Acceptance

- All perf flags use `--perf-XX` prefix.
- All test flags use `--test-XX` prefix.
- Existing scripts/docs invoking the old names are updated.
- A deprecated-alias path is optional; since these flags are new (none merged yet beyond the just-added `--wave2d-perf`), straight rename is fine.

## Cross-references

- Parent rundown meta: #3122
- The flag this issue retroactively renames: A1 of #3122, currently in flight.

PR #3223, opened 2026-06-18
Fix LPC AutoResearch SCT RX setup

zackees:
﻿## Summary

- enable the LPC SCT RX backend in the `lpc845brk` build so AutoResearch links the real capture implementation
- configure LPC845 SCT input pads through IOCON/SWM/INPUTMUX, set the SCT clock divider, and avoid resetting SWM so UART routing survives
- fix fbuild firmware artifact lookup for AutoResearch and move LPC loopback defaults off true-open-drain P0_10/P0_11 to P0_14/P0_15
- cover the fbuild path and LPC loopback command defaults with focused tests

## Context

Follow-up to #3218 / #3221 while validating AutoResearch LPC pin-toggle bring-up. The previous build could flash and echo, but the SCT RX backend was not compiled into the library TU and the test defaults used LPC845 true-open-drain I2C pins.

## Verification

- `uv run python -m pytest ci/tests/test_autoresearch_phases.py ci/tests/test_fbuild_default_selection.py -q`
  - 60 passed
- `python -m py_compile ci/autoresearch/build_driver.py ci/autoresearch/phases.py ci/autoresearch/test_lpc_pin_toggle_rx.py ci/autoresearch/test_lpc_ws2812_loopback.py ci/tests/test_autoresearch_phases.py ci/tests/test_fbuild_default_selection.py`
- `git diff --check`
- `uv run ci/ci-compile.py lpc845brk --examples AutoResearch --fbuild --max-failures 1`
  - succeeded; flags include `-DFASTLED_LPC_RX_SCT=1`; firmware size 63.12 KB / 64 KB
- `uv run python -m ci.autoresearch lpc845brk --pin-toggle-rx --upload-port COM10 --quiet --skip-schema --timeout 120s`
  - build, pyOCD flash, reset, and RPC echo succeeded
  - pin-toggle capture still failed with 0 edges at 1 kHz, 10 kHz, and 100 kHz because this attached board has no detected TX?RX electrical loopback; the runner now asks for P0_14 ? P0_15

## Remaining hardware gate

The AutoResearch pin-toggle validation still needs a physical jumper from P0_14 to P0_15, or equivalent wiring, on the attached LPC845-BRK. Earlier probing found no built-in connected GPIO pair, and same-pin SCT input is not a valid loopback.

### DIFF
--- a/ci/autoresearch/args.py
+++ b/ci/autoresearch/args.py
@@ -361,7 +361,7 @@ See Also:
             action="store_true",
             help="LPC845-only: bit-bang TX → SCT-RX pin-toggle loopback "
             "(closes Phase 1 of FastLED #3021). Requires a jumper from "
-            "--tx-pin (default P0_10) to --rx-pin (default P0_11).",
+            "--tx-pin (default P0_14) to --rx-pin (default P0_15).",
         )
         lpc_group.add_argument(
             "--ws2812-loopback",
--- a/ci/autoresearch/build_driver.py
+++ b/ci/autoresearch/build_driver.py
@@ -7,6 +7,40 @@ from pathlib import Path
 from typing import IO, Protocol, runtime_checkable
 
 
+def fbuild_firmware_path_candidates(build_dir: Path, environment: str) -> list[Path]:
+    """Return known fbuild firmware.bin artifact locations for a project."""
+    return [
+        build_dir / ".fbuild" / "build" / "release" / "firmware.bin",
+        build_dir / ".fbuild" / "build" / environment / "release" / "firmware.bin",
+        build_dir
+        / ".build"
+        / "pio"
+        / environment
+        / ".fbuild"
+        / "build"
+        / "release"
+        / "firmware.bin",
+        build_dir
+        / ".build"
+        / "pio"
+        / environment
+        / ".fbuild"
+        / "build"
+        / environment
+        / "release"
+        / "firmware.bin",
+    ]
+
+
+def find_fbuild_firmware_path(build_dir: Path, environment: str) -> Path:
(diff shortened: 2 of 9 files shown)


zackees:
Closing this draft as stale. The engineering here is real and worth keeping findable, so recording precisely what it contains and what reviving it would take.

**What's valuable and hard to re-derive:** the LPC845 IOCON per-pin offset table, the SCT clock configuration (`SCTCLKSEL`/`SCTCLKDIV`), the INPUTMUX routing for SCT input 0, and the decision to stop resetting SWM so UART routing survives. That is silicon-level detail that cost real bench time.

**Why it can't just be rebased and merged:**

1. It uses `FL_NOEXCEPT` in 25 places. That macro was renamed to `FL_NO_EXCEPT` and now has **zero** occurrences anywhere in `src/` — so the branch cannot compile against master as-is.
2. `MERGEABLE=CONFLICTING`, `drift-check` failing, untouched since 2026-06-19.
3. It touches `ci/autoresearch/args.py` and `phases.py`, which have since been restructured — including the `--perf-XX` / `--test-XX` flag rename in #3124 / #3742.
4. Most importantly: these are **register-level writes to SCT/IOCON/INPUTMUX**. They can only be validated on an LPC845 bench. Rebasing and merging unvalidated register configuration into the driver would be worse than leaving it out — a silently wrong pad config fails as "capture just doesn't work", which is exactly the class of bug that cost the bench time in the first place.

**To revive:** branch from current master, port the register work from `581cce2`, rename `FL_NOEXCEPT` → `FL_NO_EXCEPT`, re-apply against the current autoresearch flag surface, then validate on an LPC845-BRK with the loopback jumper before un-drafting. The commit stays reachable on this closed PR.

Refs #3218 / #3221.

PR #3275, opened 2026-06-19
fix(ci/boards): set PARLIO TX ISR IRAM flags on all PARLIO-capable boards (#3271)

### LINKED ISSUE #3271: esp32: CONFIG_PARLIO_TX_ISR_HANDLER_IN_IRAM=1 should be a boards.py default on all PARLIO-capable boards
## Summary

`CONFIG_PARLIO_TX_ISR_HANDLER_IN_IRAM=1` and `CONFIG_PARLIO_TX_ISR_CACHE_SAFE=1` are required for correct PARLIO TX ISR behavior — without them the ISR sits in flash, takes cache-miss stalls during execution, and violates LED protocol timing. The driver source explicitly `#warning`s when the flag is missing (`src/platforms/esp/32/drivers/parlio/parlio_peripheral_esp.cpp.hpp:46-48`).

Today these flags live **only** in the root `platformio.ini` under `[env:esp32c6]` (lines 76-77). They are not present in `ci/boards.py` for any board. CI builds get them by accident, via the root-merge in `PioCompiler._init_platformio_build` (ci/compiler/pio.py:155) — and only for `esp32c6`. Every other PARLIO-capable board compiles without them today.

## Affected boards in `ci/boards.py`

PARLIO support is gated on `SOC_PARLIO_SUPPORTED` (`src/platforms/esp/32/feature_flags/enabled.h:135`). Boards in `ci/boards.py` that should set the flags:

- `esp32c6` (`ESP32_C6_DEVKITC_1`, boards.py:801) — currently missing in boards.py, masked by root merge
- `esp32s3` (`ESP32_S3_DEVKITC_1`, boards.py:815) — missing, no mask, compiles with `#warning` today
- `esp32h2` (`ESP32_H2`, boards.py:843) — missing
- `esp32p4` (`ESP32_P4`, boards.py:850) — missing
- `esp32c5` (boards.py:795) — missing
- Anywhere else `SOC_PARLIO_SUPPORTED` is true

## Why this matters now

Discovered while auditing the root `platformio.ini` → CI merge coupling (see parent issue / discussion of PR #2794 / #2664 / #3268). The planned cleanup is to **sever root `platformio.ini` from the CI compile path entirely** and make `ci/boards.py` the sole source of truth. If we do that without porting these PARLIO flags first, every `esp32c6` build silently loses ISR-in-IRAM placement and becomes a runtime-timing regression with only a compile-time `#warning` as signal.

This issue should land **before** the root-merge sever, so the migration is safe.

## Proposed fix

Add to `build_flags` on every PARLIO-capable board definition in `ci/boards.py`:

```python
"-DCONFIG_PARLIO_TX_ISR_HANDLER_IN_IRAM=1",
"-DCONFIG_PARLIO_TX_ISR_CACHE_SAFE=1",
```

Preferred: extract a shared `ESP32_PARLIO_BUILD_FLAGS` constant (or a small mixin / helper) so the next PARLIO-capable board added doesn't quietly miss them. Alternative: emit them unconditionally from `Board.to_platformio_ini()` when the platform is ESP32 and `SOC_PARLIO_SUPPORTED` would be true — but the macro is per-target, so a board-side allowlist is simpler than a platform-side derivation.

## Verification

- After porting, remove the duplicate lines from `[env:esp32c6]` in root `platformio.ini` to confirm the flags now come from `boards.py`.
- Compile each PARLIO board and confirm the `#warning` in `parlio_peripheral_esp.cpp.hpp` no longer fires.
- Optional: add a `static_assert` next to the `#warning` so missing flags are a hard error on PARLIO-supporting targets (ratchet from warn → error).

## Labels / refs

- area: build, area: esp32, parlio
- Blocks: parent root-`platformio.ini` decoupling work (to be filed)
- Discovered via: PR #3268 follow-up audit

davidcreated:
## Summary

Fixes #3271.

`CONFIG_PARLIO_TX_ISR_HANDLER_IN_IRAM=1` and `CONFIG_PARLIO_TX_ISR_CACHE_SAFE=1` are required for correct PARLIO TX ISR behavior. Without them the ISR sits in flash, takes cache-miss stalls during execution, and violates LED protocol timing — the driver only signals this with a compile-time `#warning` (`src/platforms/esp/32/drivers/parlio/parlio_peripheral_esp.cpp.hpp:46-48`).

Today these flags live **only** in the root `platformio.ini` `[env:esp32c6]` block. CI gets them by accident via the root-merge in `PioCompiler._init_platformio_build` — and only for `esp32c6`. Every other PARLIO-capable board (`esp32s3`, `esp32h2`, `esp32p4`, `esp32c5`) compiles without them today.

## Change

- Extract a shared `ESP32_PARLIO_BUILD_FLAGS` constant in `ci/boards.py`.
- Apply it to the `build_flags` of every PARLIO-capable board: `esp32c5`, `esp32c6`, `esp32s3`, `esp32h2`, `esp32p4`.

This makes `ci/boards.py` a self-sufficient source of truth ahead of the planned root-`platformio.ini` sever, and ensures the next PARLIO board added cannot silently miss the flags (just spread the shared constant).

## Scope note

The root `platformio.ini` `[env:esp32c6]` flags are intentionally **left in place**. That env is still used for standalone manual `pio run` (which does not go through `boards.py`), and the root-merge removal it depends on is tracked as separate follow-up work in #3271. Removing them now would regress standalone dev builds before the sever lands. The duplicate `-D` on the c6 CI build is harmless.

## Verification

Rendered each board's `to_platformio_ini()` and confirmed both flags are present:

```
esp32c5      IRAM=True CACHE_SAFE=True
esp32c6      IRAM=True CACHE_SAFE=True
esp32s3      IRAM=True CACHE_SAFE=True
esp32h2      IRAM=True CACHE_SAFE=True
esp32p4      IRAM=True CACHE_SAFE=True
```

`python3 -m py_compile ci/boards.py` passes.

## Release Notes

* **Chores**
  * Standardized PARLIO-related compiler flags across all PARLIO-capable ESP32 board variants (C5, C6, S3, H2, and P4) to ensure consistent build behavior.
  * Exposed a shared, reusable set of build flags for PARLIO so each board applies the same ISR placement and cache-safety settings.

### DIFF
--- a/ci/boards.py
+++ b/ci/boards.py
@@ -808,19 +808,6 @@ STM32H747XI_GIGA = Board(
 #     platform=ESP32_IDF_5_1_PIOARDUINO,
 # )
 
-# Required on every PARLIO-capable ESP32 target (C5/C6/H2/P4/S3). Without these,
-# the PARLIO TX ISR is placed in flash and may take cache-miss stalls during
-# execution — long enough to violate LED protocol timing. The driver emits a
-# #warning at compile time when CONFIG_PARLIO_TX_ISR_HANDLER_IN_IRAM is missing
-# (src/platforms/esp/32/drivers/parlio/parlio_peripheral_esp.cpp.hpp:46-48).
-# Historically these lived only in root platformio.ini under [env:esp32c6] and
-# reached CI only via the implicit root-merge — leaving every other PARLIO
-# board (S3, H2, P4, C5) compiling without them. See #3271.
-_ESP32_PARLIO_BUILD_FLAGS = [
-    "-DCONFIG_PARLIO_TX_ISR_HANDLER_IN_IRAM=1",
-    "-DCONFIG_PARLIO_TX_ISR_CACHE_SAFE=1",
-]
-
 # ESP32-C2: Use Arduino framework only (not "arduino, espidf")
 # The dual framework mode causes PlatformIO to use ESP-IDF's component-based build system,
 # which does not automatically discover and compile .cpp files in example subdirectories
@@ -853,7 +840,6 @@ ESP32_C5_DEVKITC_1 = Board(
     real_board_name="esp32-c5-devkitc-1",
     platform=ESP32_IDF_5_5_1_PIOARDUINO,
     build_flags=[*ESP32_PARLIO_BUILD_FLAGS],
-    build_flags=list(_ESP32_PARLIO_BUILD_FLAGS),
 )
 
 ESP32_C6_DEVKITC_1 = Board(
@@ -863,7 +849,6 @@ ESP32_C6_DEVKITC_1 = Board(
     board_build_flash_size="4MB",  # ESP32-C6FH4 actual flash size confirmed by esptool
     board_partitions="huge_app.csv",
     build_flags=[
-        *_ESP32_PARLIO_BUILD_FLAGS,
         "-funwind-tables",  # Better stack traces for RISC-V crash decoding
         "-DARDUINO_USB_MODE=1",  # Select HWCDC (USB-Serial/JTAG) over OTG
         "-DARDUINO_USB_CDC_ON_BOOT=1",  # Route Serial to HWCDC. Safe with setTxTimeoutMs(0); see #2668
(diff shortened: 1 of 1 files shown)


davidcreated:
Addressed in 1947d37 — added the explicit `list[str]` annotation to `ESP32_PARLIO_BUILD_FLAGS`. Thanks @coderabbitai.

zackees:
Thank you for the contribution @davidcreated! Closing as **superseded** — your fix was correct in intent and just lost the race to a parallel effort by an hour or so. The PARLIO IRAM lockdown has now landed on master across a sequence of PRs:

- **PR #3276** — initial PARLIO IRAM flags in \`ci/boards.py\` (closed #3271)
- **PR #3299** — documented the correct PARLIO-capable chipset list (C5, C6, H2, P4 — and explicitly **not** S3) in the channels README
- **PR #3306** — refactored to a \`parlio_capable\` field on the \`Board\` dataclass with auto-emit from \`Board.to_platformio_ini()\`, removed S3 from the PARLIO set (both your PR and #3276 had incorrectly included S3 since it has no PARLIO peripheral — see #3304)

Net result on master: every PARLIO-capable board (C5/C6/H2/P4) gets the IRAM + cache-safe flags automatically, S3 correctly does not, and a snapshot test (\`ci/tests/test_parlio_capable_boards.py\`) guards against future drift.

Really appreciate you taking the time to file this — same fix, just routed through a different shipping cycle.

Issue #3325, opened 2026-06-20
[#3300 bisect] Find the commit that broke LPC845-BRK JSON-RPC echo — bisect Arduino-core LPC8xx


zackees:
## Goal

Find the commit (in FastLED, fbuild, or `zackees/ArduinoCore-LPC8xx`) that broke LPC845-BRK JSON-RPC echo, by bisection.

## What we know

**Last known-working baseline:** **2026-06-13 ~17:40 PT** -- FastLED commit `0968d55f28` (`feat(autoresearch): wire LPC845-BRK bring-up harness into bash autoresearch`). Its body claims direct verification:

> "the three hardware capabilities (Serial.println, FastLED log pipeline, JSON-RPC echo) all work in isolation on the device -- proven by the JSON-RPC round-trip `TX params:[4242] -> RX result:4242` plus FL_DBG emission `src/fl/remote/remote.cpp.hpp(151): Stored request ID for echo` reaching COM10"

**Current broken state:** **2026-06-19 / 20** -- direct PowerShell `[System.IO.Ports.SerialPort]::Open('COM20')` returns **0 bytes in 6 seconds** after a clean pyocd reset. The device transmits nothing.

### Phase 1: Confirm FastLED-side refactors did not regress the RPC dispatch path

Test ESP32-C6 (board already attached on `COM9`):

\`\`\`bash
bash autoresearch esp32c6
\`\`\`

- **If echo round-trips** -> FastLED's RPC refactor (#3238, #3258, #3294) is sound across platforms. Bug is **LPC-specific**, almost certainly in `zackees/ArduinoCore-LPC8xx`. Skip to Phase 2.
- **If echo fails on ESP32-C6 too** -> regression is in FastLED's cross-platform RPC layer. Bisect `master` of FastLED instead. Suspect window: \`0968d55f28..HEAD\` on FastLED.

### Phase 2: Bisect the LPC Arduino-core

The framework is referenced by URL only (no pinned commit) in `zackees/platform-nxplpc-arduino`'s `platform.json`:

\`\`\`json
"framework-arduino-lpc8xx": {
    "type": "framework",
    "optional": true,
    "version": "https://github.com/zackees/ArduinoCore-LPC8xx.git"
}
\`\`\`

[...]

Issue #3246, opened 2026-06-18
[Type-Erasure] RPC dispatch: RuntimeRpcBinding -- replace TypedRpcBinding<Sig> with non-template + per-Sig trampoline (#3244 Tier 1A)


zackees:
Parent: #3244 (Phase 3 burn-down).

## Goal

Replace `fl::TypedRpcBinding<Sig>::invokeWithReturn` with a non-template `RuntimeRpcBinding` plus a tiny per-Sig trampoline. Collapse the ~2.1 KB per-Sig template instantiation cost (observed on LPC845-BRK as the `TypedRpcBinding<fl::string(int,int,int,int)>::invokeWithReturn` symbol at 2,142 B) down to a ~50-150 B per-Sig trampoline + a one-time ~1 KB shared dispatch body.

This is the **Tier 3H deferral** from meta #3224 and the **Tier 1A** carry-forward from meta #3235.

## Reproduction (baseline)

```
bash bloat lpc845brk --example AutoResearch --top 10
# Look for the line:
#   2,142 B  fl::TypedRpcBinding<fl::string (int, int, int, int)>::invokeWithReturn
```

That ~2 KB is per registered method. AutoResearch's LowMemory mode registers ~3-4 methods; bigger sketches register 10+, where the per-Sig tax scales linearly.

## Acceptance

- [ ] PR opens `RuntimeRpcBinding` (non-template, in `fl/remote/rpc/`) that takes:
  - `void* user_fn` (type-erased pointer to the callable)
  - `RpcTrampolineFn` (per-Sig 50-150 B trampoline that argument-extracts and calls the user fn)
  - `const RpcTypeId* arg_types` + `RpcTypeId result_type` (runtime arg-type table)
  And implements `ErasedInvoker::invoke(const json&)` as a single signature-agnostic decode/encode/dispatch body.
- [ ] Per-Sig trampoline is generated at the `Rpc::bind(name, lambda)` site (no source change to existing call sites).
- [ ] `Rpc::registerMethod<Sig>` compile-time forks on `FL_PLATFORM_HAS_LARGE_MEMORY`: Low-memory uses `RuntimeRpcBinding`, Large-memory may keep `TypedRpcBinding<Sig>` if benchmarks show speed regression.
- [ ] Measured savings on LPC845-BRK / AutoResearch: ≥ 1.5 KB net flash drop (target is ~6 KB across the 4 registered methods, but per-Sig accounting must show at least 1.5 KB).
[...]

Issue #3300, opened 2026-06-19
[BLOCKER] LPC845-BRK AutoResearch bring-up echo returns empty RX (suspected HardFault) — blocks #3102


zackees:
## Symptom

On LPC845-BRK with the AutoResearch low-memory firmware (flash 53.61 KB / 64 KB = 83.8%, RAM 9.03 KB / 16 KB = 56.4%), the bring-up phase observes:

- Build + flash + reset succeed (pyOCD CMSIS-DAP).
- Device boots and emits log lines back over USB-VCOM (we see `src/fl/remote/remote.cpp.hpp(151): INFO: Stored request ID for rpc.discover`).
- `rpc.discover` returns a schema response (malformed under current pydantic model but data IS present -- list-of-lists with method names like `'testSimd'`).
- **`echo` RPC request returns empty:** `TX: {"jsonrpc":"2.0","method":"echo","params":[4242],"id":1}` → `RX: b''`.
- Phase fails: `BRING-UP TEST FAILED echo_ok=False remote_ok=False warn_ok=False log_ok=False`.

This is a **blocker** for the #3102 bench-validation pass. Two of four validation phases (build/flash/reset and link-symbol soft-FP check) are green; the bring-up echo silence prevents validating the USB-VCOM RPC round-trip and gates the SCT-RX phases.

## What we've already ruled out

1. **Method not registered.** `echo` IS bound in `examples/AutoResearch/AutoResearchLowMemory.h:158` (`remote.bind("echo", [](int v) -> int { ... return v; })`). The dispatcher also explicitly handles missing-method with `-32601 "Method not found"` + `FL_WARN_F` log at `src/fl/remote/rpc/rpc.cpp.hpp:121`. Silence (not an error response) means the request never reached the dispatcher.

2. **Jumper interference.** The LPC845-BRK USB-VCOM bridge is hardwired to **P0_25 (U0_TXD) / P0_24 (U0_RXD)** (per `variants/lpc845/variant.h`). The autoresearch default loopback jumper is **P0_10 → P0_11** -- general-purpose GPIO, not USART pins. Confirmed jumper-on-loopback-pins cannot break host↔device USART.

3. **Wrong COM port.** Confirmed COM20 IS the LPC845-BRK USB-VCOM port (NOT COM25, which is an unrelated ESP32-P4). Discovery responses come back on COM20, proving the path is alive.

## Hypothesis (per user instinct)

> "in the past the chip not responding means it's crashed"

[...]

PR #3337, opened 2026-06-20
test(remote): wire-format echo bind canary (FastLED#3300 / #3325)

zackees:
## Summary

Adds a host unit test that mirrors `AutoResearchLowMemory.h:158`'s exact bind pattern (`remote.bind(\"echo\", [](int v) -> int { return v; })`) and drives the full `pull → handler → processRpc → mRpc.handle → sink → push` pipeline using the literal JSON-RPC wire bytes the device receives.

**The test PASSES on master.** That is itself the finding: the bug seen on LPC845-BRK and ESP32-C6 (\`{\"error\":{\"code\":-32601,\"message\":\"Method not found: echo\"}}\`) does NOT reproduce in the host bind+dispatch chain. The regression must live in one of three places host doesn't model:

1. **Device-side serial source / parse** — `createSerialRequestSource()` uses real serial I/O. If the line accumulator hands the dispatcher a request whose \`method\` field differs from \`\"echo\"\` (extra char, BOM, surrogate, etc.), lookup fails. Host bypasses this.
2. **Static-init / ODR linkage** — `static fl::Remote remote(...)` inside an inline function in `AutoResearchLowMemory.h`, with `g_low_memory_remote` in an anonymous namespace per-TU. If the setup() and loop() functions resolve to different TUs' `g_low_memory_remote`, bind() and dispatch could touch different Remote instances. Host uses a directly-constructed Remote.
3. **Build-define divergence** — host has `-DSKETCH_HAS_LARGE_MEMORY=1` so `FL_PLATFORM_HAS_LARGE_MEMORY=1`. On LPC845-BRK the same flag is _also_ 1, but only because `FL_IS_ARM_LPC_845` mis-detects (ArduinoCore-LPC8xx's `variant.h` defines `VARIANT_BOARD_LPC845BRK` but not the `LPC845*` macros our `is_lpc.h` keys off of — see #3325 update). The TypedInvoker path then runs on a 16 KB part. If `fl::make_shared<TypedInvoker<int(int)>>(fn)` returns null shared_ptr under OOM and `mRegistry[key] = fl::move(entry)` silently no-ops on a full `vector_inlined<Entry, 8>`, lookup misses. Not yet proven.

## Why land this anyway

Even though it currently passes, the canary anchors the bind+dispatch invariant against future regressions. The next time someone refactors `RuntimeRpcBinding` / `TypedInvoker` (e.g. another #3246-style change) and accidentally breaks the registration key, this test fails immediately instead of being discovered weeks later via a silent device.

The test is tiny (~50 LOC), runs in 0.03s, and uses zero new infrastructure (reuses the existing \`TestIO\` fixture + \`makeRequest\` helper).

## Test plan

- [x] \`bash test remote\` — passes (the new case + 100+ existing remote tests, total ~0.22s test execution).
- [ ] CI cross-platform.

## Refs

- FastLED#3300 — LPC845-BRK bring-up echo silence (root).
- FastLED#3325 — ACLPC bisection plan, re-scoped after this test confirmed the bug is FastLED-side, not Arduino-core. See [#3325 update comment](https://github.com/FastLED/FastLED/issues/3325#issuecomment-4757727203).
- FastLED#3336 — agents/hooks: forbid PowerShell SerialPort (the methodology bug that masked the real symptom for two sessions).

## Summary by CodeRabbit

* **Tests**
  * Added regression test for JSON-RPC echo functionality to validate proper request parsing and response handling, ensuring method resolution reliability.

### DIFF
--- a/tests/fl/remote/remote.cpp
+++ b/tests/fl/remote/remote.cpp
@@ -1713,4 +1713,53 @@ FL_TEST_CASE("Remote: Async RPC - Sync vs Async behavior difference") {
         FL_REQUIRE(ackResult["acknowledged"].as_bool().value() == true);
     }
 }
+
+// =============================================================================
+// FastLED #3300 / #3325 — AutoResearch echo bind regression reproducer
+// =============================================================================
+//
+// Mirrors AutoResearchLowMemory.h:158 — `remote.bind("echo", [](int v) -> int
+// { return v; })` — and the exact JSON-RPC wire request the LPC845-BRK and
+// ESP32-C6 both see ("RESULT: Method not found: echo"). If this test fails
+// on host we have a fast reproducer for the binding regression; if it
+// passes, the bug is in the device-side serial source / parse / wiring,
+// not in bind/dispatch.
+
+FL_TEST_CASE("Remote: AutoResearch echo via parsed wire JSON (issue #3300)") {
+    TestIO io;
+    fl::Remote remote(
+        [&io]() { return io.pullRequest(); },
+        [&io](const fl::json& r) { io.pushResponse(r); }
+    );
+
+    // Exact AutoResearch binding (AutoResearchLowMemory.h:158).
+    remote.bind("echo", [](int v) -> int { return v; });
+
+    // Exact wire bytes the device receives.
+    const char* wireRequest =
+        "{\"jsonrpc\":\"2.0\",\"id\":1,\"method\":\"echo\",\"params\":[4242]}";
+
+    fl::optional<fl::json> parsed = fl::json::parse(wireRequest);
+    FL_REQUIRE(parsed.has_value());
+    FL_REQUIRE(parsed->contains("method"));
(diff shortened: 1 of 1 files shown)


zackees:
Closing as a completed investigation rather than a change to merge.

This PR's own finding is that **the test passes on master** — the `{"error":{"code":-32601,"message":"Method not found: echo"}}` failure seen on LPC845-BRK and ESP32-C6 does not reproduce in the host bind+dispatch chain. That is a useful negative result (it rules out the host-modelled `pull → handler → processRpc → mRpc.handle → sink → push` path and points at the three device-side layers host doesn't model), but it is not a fix, and the branch has been red since 2026-06-20.

Merging a canary that passes for the wrong reason would add CI surface without protecting anything. The diagnostic value lives in the writeup, which stays on this closed PR and is linked from #3300 / #3325.

If the wire-format bug resurfaces, reopen and rebase — the test file is a good starting point once there's a device-side reproduction to anchor it.

PR #3619, opened 2026-07-12
The PIO assembly definitions are incorrect for two of the MOV operations: INVERT and REVERSE

eelstretching:
According to [the PIO datasheet](https://pip-assets.raspberrypi.com/categories/1214-rp2350/documents/RP-008373-DS-2-rp2350-datasheet.pdf), PIO_MOV_OP_INVERT should be 0b01 (not 0b00) and PIO_MOV_OP_REVERSE should be 0b10 (not 0b00).

This is a small commit that fixes this.

Discovered while trying to fill out the parallel version of the controllers for the Pi Pico.

fbuild's a little broken for me on MacOS for the moment, but 

`uv run ci/ci-compile.py rp2350 all --backend platformio
`
builds everything with no issues.

## Summary by CodeRabbit

- **Bug Fixes**
  - Corrected PIO move-operation encodings for invert and reverse operations.
  - Ensures these operations execute with their intended behavior on supported RP platforms.


zackees:
thanks for the contribution! fbuild for mac is updating now and should be fixed today.

Issue #3716, opened 2026-07-22
esp32: bare-IDF fl::platforms time impl — drop Arduino millis()/micros()/delayMicroseconds() (blocks all framework=espidf builds)


zackees:
Part of #3715.

# Problem

`fl::platforms::{delay, delayMicroseconds, millis, micros}` for ESP32 is implemented only in `src/platforms/esp/32/platform_time_esp32.cpp.hpp`, and the whole file is gated:

```cpp
#if defined(FL_IS_ESP32) && defined(ARDUINO) && !defined(FASTLED_STUB_IMPL)   // line 11
```

Inside, it calls Arduino globals: `::delayMicroseconds()` (line 39), `::millis()` (line 44), `::micros()` (line 48). The generic fallback `src/platforms/arduino/platform_time.cpp.hpp` is *also* gated on `defined(ARDUINO)` (and excludes `FL_IS_ESP32` anyway).

**Consequence: a bare ESP-IDF build has no time implementation at all** — `fl::platforms::millis` etc. are undefined symbols at link time. This is the single hardest blocker for any `framework=espidf` build and must land before the CI sub-issue can go green.

Additional stray Arduino call: `src/platforms/esp/32/clockless_block_esp32.h:90` calls unqualified `delayMicroseconds(...)`.

# Fix

Replace the Arduino calls with IDF-native equivalents so the impl compiles under both Arduino and bare IDF, and drop the `defined(ARDUINO)` term from the guard:

| Arduino | IDF-native |
|---|---|
| `::millis()` | `esp_timer_get_time() / 1000` |
| `::micros()` | `esp_timer_get_time()` |
| `::delayMicroseconds(us)` | `esp_rom_delay_us(us)` (IDF 4+) / `ets_delay_us(us)` (IDF 3.x) |
| `::delay(ms)` | already `vTaskDelay(pdMS_TO_TICKS(ms))` — keep (FreeRTOS, framework-agnostic) |

The existing ms/us split in `delayMicroseconds` (yield via `vTaskDelay` for whole ms, busy-wait the remainder) should be preserved.

`clockless_block_esp32.h:90` → route through `fl::platforms::delayMicroseconds` (or `esp_rom_delay_us` directly).

# IDF compat notes (3.x/4.x/5.x/6.x)

- `esp_timer_get_time()` exists in IDF 3.x+ (`esp_timer.h`); on 3.x the header lives at the same include path — verify against the `esp32dev_idf33` CI leg.
[...]

Issue #3715, opened 2026-07-22
Meta: decouple src/platforms/esp from the Arduino core — go straight to Espressif (ESP-IDF) headers, with IDF 3.x-6.x compat + per-IDF compile-test CI


zackees:
# Goal

Take `src/platforms/esp/**` fully off the Arduino dependency path: every ESP32/ESP8266 code path should compile against **Espressif headers directly** (ESP-IDF drivers / ROM / SoC headers, or NONOS-SDK on 8266), with the Arduino core reduced to an optional, opt-in convenience layer — never a requirement.

# Current state (audit of `src/platforms/esp/**`, 2026-07-22)

The good news: the tree is already *mostly* IDF-native. `led_sysdefs_esp32.h` / `led_sysdefs_esp8266.h`, all of `32/drivers/*` (rmt_5, spi, i2s, i2s_spi, parlio, lcd_cam, lcd_spi, uart), `net/`, `ota/`, `ble`, `interrupts/`, and the `pin_*_native*` implementations already use ESP-IDF / SDK headers directly. **Those need no migration.**

The remaining Arduino-core coupling is concentrated in a handful of legacy files:

| Area | Files | Arduino dependency |
|---|---|---|
| Timing | `32/platform_time_esp32.cpp.hpp`, `32/clockless_block_esp32.h` | `::millis()`, `::micros()`, `::delayMicroseconds()`; **no bare-IDF time impl exists at all** — a pure-IDF build fails to link `fl::platforms::millis` etc. |
| Cycle counter | `32/core/clock_cycles.h`, `32/drivers/rmt/rmt_4/channel_driver_rmt4.h` | `<esp32-hal.h>` fallback branches |
| SPI (ESP32) | `32/core/fastspi_esp32_arduino.h` | `<SPI.h>`, `SPIClass`, `SPISettings` (an IDF backend already exists in `fastspi_esp32_idf.h`) |
| PWM / CLED | `32/drivers/cled.cpp.hpp` | `ledcAttach` / `ledcAttachPin` / `ledcSetup` / `ledcWrite` + `ESP_ARDUINO_VERSION_MAJOR` fork |
| Serial IO | `32/io_esp.cpp.hpp` | probes `FL_HAS_INCLUDE(<Arduino.h>)` and can route to `platforms/arduino/io_arduino.hpp` (an IDF path already exists in `io_esp_idf.hpp`) |
[...]

PR #3731, opened 2026-07-22
feat(esp8266): qualify delayMicroseconds; document scope decision on the rest

### LINKED ISSUE #3723: esp8266: migrate off Arduino core — pgmspace, SPI.h, HardwareSerial Serial1, delayMicroseconds
Part of #3715.

# Problem

The ESP8266 subtree retains the heaviest Arduino coupling in the ESP family. Note the target here is the **NONOS SDK / bare-metal registers**, not ESP-IDF (8266 never got IDF proper; RTOS-SDK is out of scope):

| file:line | dependency | native replacement |
|---|---|---|
| `8266/progmem_esp8266.h:9` | `#include <pgmspace.h>` (Arduino core) | flash is memory-mapped on ESP8266 — `FL_PGM_READ_*` become plain (aligned) pointer reads; `PROGMEM` → section attribute or no-op |
| `8266/fastspi_esp8266.h:7` | `#include <SPI.h>`, Arduino `SPIClass` | direct HSPI register programming via SDK `spi_register.h` — or explicitly document 8266 hardware-SPI as Arduino-only |
| `8266/fastled_esp8266_uart.h:31` | `extern HardwareSerial Serial1` | direct UART1 FIFO writes (`USF(1)`) via `esp8266_peri.h` registers |
| `8266/fastled_esp8266_uart.h:99,108,131,161` | `Serial1.flush()/begin()/write()` | UART1 register sequence (baud via `USD(1)`, conf via `USC0(1)`) |
| `8266/fastled_esp8266_uart.h:32,134,167` + `8266/clockless_esp8266.h:67` + `8266/clockless_block_esp8266.h:59` | `delayMicroseconds(...)` | `ets_delay_us(...)` (ROM/SDK) |
| `8266/led_sysdefs_esp8266.h:36` | `#ifdef ARDUINO_ESP8266_NODEMCU` pin-order fork | keep — board *identification* macro, allowed exception per the dispatch-convention sub-issue |
| `8266/ldf_headers.h:16` | `<SPI.h>` LDF hint | remove alongside the fastspi migration |

Already native (no work needed): `8266/pin_esp8266_native.hpp` (register-level GPIO via `esp8266_peri.h`/`osapi.h` — SDK headers, not Arduino wiring), `8266/io_esp_idf.hpp` (ROM `ets_putc`), `8266/fastpin_esp8266.h` (SDK `GPO`/`GP16O` registers).

# Priority

Lower than the ESP32 sub-issues — ESP8266 is a legacy platform and its Arduino core is the only actively maintained runtime most users have. Recommended scope cut: do the cheap wins (`pgmspace`, `delayMicroseconds`→`ets_delay_us`), and take a deliberate documented decision on UART/SPI (migrate vs declare Arduino-only) rather than defaulting into a half-migration.

# Acceptance criteria

- [ ] `progmem_esp8266.h` has no Arduino include; PROGMEM reads verified on-device (flash-mapped read alignment on 8266 traps on unaligned byte reads — the existing `FL_PGM_READ_BYTE_NEAR` shift/mask logic must be preserved).
- [ ] No bare `delayMicroseconds` calls remain under `src/platforms/esp/8266/`.
- [ ] Decision recorded (in this issue) for UART1 + hardware-SPI: native registers or documented Arduino-only.
- [ ] ESP8266 CI leg green.

zackees:
Closes #3723. Part of meta #3715 (ESP Arduino-decoupling).

## Change
Small, safe cleanup: `clockless_esp8266.h` and `clockless_block_esp8266.h` call unqualified `delayMicroseconds(...)` inside `namespace fl { ... }` — this already resolves to the `fl::` facade via unqualified lookup (same situation as `clockless_block_esp32.h` before #3716), but is now explicit: `fl::delayMicroseconds(...)`.

## Scope decision (documented per the issue's own guidance)
#3723's acceptance criteria explicitly called for "a decision recorded... for UART1 + hardware-SPI: native registers or documented Arduino-only" rather than a rushed half-migration. After looking at each remaining item, here's the call:

- **`fl::delayMicroseconds()` still resolves to Arduino's `::delayMicroseconds()`** on ESP8266 (no ESP8266-specific `platform_time` file exists; it falls through to the generic `platforms/arduino/platform_time.cpp.hpp`). A native `ets_delay_us()`-based microsecond delay is plausible and safe, but ESP8266's millisecond-scale `delay()` needs software-watchdog-aware chunking to avoid resets — I can't validate that across delay durations without extensive on-device testing. **Documented follow-up, not migrated.**
- **`pgmspace.h`**: NOT migrated. ESP8266 flash reads are memory-mapped and alignment-sensitive (see this file's own `FL_ALIGN_PROGMEM` comment) — exactly the kind of change that risks silent data corruption if gotten subtly wrong, and the existing Arduino implementation is the only one in this codebase that's actually hardware-validated. **Kept Arduino-only.**
- **`fastspi_esp8266.h` (`<SPI.h>`) and `fastled_esp8266_uart.h` (`HardwareSerial Serial1`)**: NOT migrated. Both are real protocol implementations with no existing native-SDK counterpart in this codebase to promote (unlike ESP32, which already had a complete `fastspi_esp32_idf.h`). A register-level rewrite is a real, separate effort deserving its own issue with dedicated hardware bring-up — not something to rush inside a low-priority sub-issue of a larger meta. **Kept Arduino-only.**

## Test plan
- [x] `bash lint --cpp` — clean
- [x] `bash test --cpp` — 355/355 pass
- [ ] `bash compile esp8266 --examples Blink` — real ESP8266 hardware attached (NodeMCU 0.9 / ESP-12) at COM11; validating separately due to fbuild daemon lock contention with other concurrent #3715 sub-issue builds, will follow up with results as a PR comment.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved microsecond timing reliability for ESP8266 LED output and frame retry handling.
  * Preserved existing delay behavior while routing timing through the platform’s FastLED delay utility.

### DIFF
--- a/src/platforms/esp/8266/clockless_block_esp8266.h
+++ b/src/platforms/esp/8266/clockless_block_esp8266.h
@@ -13,6 +13,7 @@
 #include "fastled_delay.h"
 #include "fl/stl/compiler_control.h"
 #include "fl/stl/noexcept.h"
+#include "fl/system/delay.h"
 
 FL_DISABLE_WARNING_PUSH
 FL_DISABLE_WARNING_DEPRECATED_REGISTER
@@ -56,7 +57,7 @@ public:
 			#ifdef FASTLED_DEBUG_COUNT_FRAME_RETRIES
 			++_retry_cnt;
 			#endif
-			delayMicroseconds(WAIT_TIME * 10);
+			fl::delayMicroseconds(WAIT_TIME * 10);
 			os_intr_lock();
 		}
 		// #if FASTLED_ALLOW_INTTERUPTS == 0
--- a/src/platforms/esp/8266/clockless_esp8266.h
+++ b/src/platforms/esp/8266/clockless_esp8266.h
@@ -9,6 +9,7 @@
 #include "fastled_delay.h"
 #include "fl/stl/compiler_control.h"
 #include "fl/stl/noexcept.h"
+#include "fl/system/delay.h"
 
 FL_DISABLE_WARNING_PUSH
 FL_DISABLE_WARNING_DEPRECATED_REGISTER
@@ -64,7 +65,7 @@ protected:
       #ifdef FASTLED_DEBUG_COUNT_FRAME_RETRIES
       ++_retry_cnt;
       #endif
-      delayMicroseconds(WAIT_TIME);
+      fl::delayMicroseconds(WAIT_TIME);
     }
     mWait.mark();
   }


zackees:
Local hardware compile note: attempted \`bash compile esp8266 --examples Blink\` against the real attached ESP8266 board (NodeMCU 0.9 / ESP-12, COM11). It fails with \`fatal error: SPI.h: No such file or directory\` in \`fl/system/sd/fs_sdcard_arduino.cpp.hpp\` — a completely unrelated SD-card subsystem, not anything touched by this PR.

Confirmed this is pre-existing, not caused by this PR: reproduces identically on a clean \`master\` checkout. Root cause is a local environment gap — this machine has never installed the `framework-arduinoespressif8266` PlatformIO package (`~/.platformio/packages/` has no ESP8266 Arduino framework or platform at all, only the bare xtensa-lx106 toolchain), so no `<SPI.h>` is discoverable anywhere for this board.

The compile did get through all 83 individual source files (including the two files this PR touches, `clockless_esp8266.h`/`clockless_block_esp8266.h`) before hitting the unrelated SD-card file — good evidence this PR's actual changes compile fine; I just can't produce a full linked firmware.bin for this board on this machine without a larger, out-of-scope framework install. Leaving that as a known gap rather than spending session time on an unrelated package install.

PR #3746, opened 2026-07-28
chore(api): take DiodeProfile by reference in set_input_gamut (#3234)

### LINKED ISSUE #3234: chore(api): audit mutable pointer shape for set_input_gamut
﻿## Context

While implementing #3233, the existing RGBW input gamut API surfaced this signature shape:

```cpp
void set_input_gamut(DiodeProfile* profile, InputGamut g, ...);
```

FastLED generally shies away from mutable pointer parameters when the parameter is required. A mutable pointer implies nullable / optional semantics and forces every call path to reason about `nullptr`. Where mutation is required and null is not meaningful, a mutable reference (`DiodeProfile&`) is usually clearer. Where mutation is not required, `const DiodeProfile&` is preferred.

This follow-up should decide whether `set_input_gamut` really needs nullable pointer semantics or whether the public/internal API should be tightened.

## Proposal

Audit `set_input_gamut(DiodeProfile* profile, ...)` and related CFastLED wrappers/callers.

Questions to answer:

- Is `nullptr` an intentional no-op API contract, or just defensive implementation detail?
- If null is not a meaningful user-facing operation, should the API become `DiodeProfile& profile`?
- If the function only needs read access in any overload, should that path use `const DiodeProfile&` instead?
- Should the CFastLED wrapper keep nullable behavior while an internal helper uses references?
- What migration path avoids breaking sketches unnecessarily?

## Acceptance criteria

- All `set_input_gamut` declarations, definitions, wrappers, and call sites are audited.
- The issue is resolved by either:
  - converting required mutable profile parameters to `DiodeProfile&`, or
  - documenting why nullable `DiodeProfile*` is intentional and keeping the current shape.
- Tests or compile coverage are updated if the API shape changes.
- Agent review docs are updated if this becomes a broader FastLED rule for mutable pointer parameters.

## Related

- Follow-up from #3233.

zackees:
## Summary

Closes #3234.

**Audit result:** `profile` is required and mutated, so the nullable pointer advertised an optional argument that doesn't exist. Every call site already passed a real address (`&copy` internally, `&p` throughout the tests) — the `if (profile == nullptr) return;` was defensive, not a contract. And silently doing nothing is a poor answer to "you gave me no profile to configure": the caller gets no gamut applied and no signal.

So `DiodeProfile*` → `DiodeProfile&`.

## What deliberately did *not* change

`white_xy` **stays a pointer.** It is genuinely optional — `nullptr` means "use the gamut's standard reference white" — so the nullable pointer is telling the truth there. The goal is matching the shape to the semantics, not removing pointers on principle.

Answering the issue's questions directly:

| Question | Answer |
|---|---|
| Is `nullptr` an intentional contract? | No — defensive implementation detail, no caller relies on it |
| Should it become `DiodeProfile&`? | Yes |
| Does any overload need `const DiodeProfile&`? | No — every path mutates |
| Should a wrapper keep nullable behaviour? | Yes, as a deprecated forwarder (below) |

## Migration — no sketch breaks

The pointer overloads are kept and marked `FL_DEPRECATED`, forwarding to the reference forms. Existing `set_input_gamut(&profile, ...)` calls keep compiling and get a pointed deprecation message.

The forwarders **preserve the silent no-op on `nullptr`** rather than dereferencing. A sketch that relied on that behaviour must not start crashing at the moment we deprecate it — the deprecation is the signal, not a segfault.

## Testing

- `bash test --cpp` — **357/357**, including all **730 examples linking**, so the new deprecation doesn't break example builds
- `bash lint` — green
- Two new tests: one asserting the pointer and reference paths produce identical output, one pinning the null no-op

Both use `FL_DISABLE_WARNING(deprecated-declarations)` locally so exercising the deprecated path doesn't add warning noise.

## Summary by CodeRabbit

- **New Features**
  - Added reference-based APIs to configure input color gamut for RGBW colorimetric modes, with an optional white-point override.
- **Improvements**
  - Updated gamut configuration so colorimetric results refresh immediately after changes.
  - Kept pointer-based configuration for compatibility, including null-pointer no-op behavior.
- **Bug Fixes**
  - Ensured deprecated pointer overloads forward correctly and match reference overload results for valid profiles.
- **Tests**
  - Expanded coverage to verify correct forwarding, deprecation behavior, and crash-free null handling.

### DIFF
--- a/src/fl/gfx/rgbw.cpp.hpp
+++ b/src/fl/gfx/rgbw.cpp.hpp
@@ -244,7 +244,7 @@ constexpr NamedGamut kDciP3D60  = {{0.6800f, 0.3200f}, {0.2650f, 0.6900f},
 namespace { void invalidate_colorimetric_caches_for(const DiodeProfile* profile) FL_NO_EXCEPT; }
 
 void set_input_gamut(DiodeProfile& profile_ref, InputGamut g,
-                     const float white_xy[2]) FL_NO_EXCEPT {
+                     const float* white_xy) FL_NO_EXCEPT {
     DiodeProfile* const profile = &profile_ref;
     auto apply = [profile](const float r[2], const float gp[2],
                            const float b[2], const float w[2]) {
--- a/src/fl/gfx/rgbw.cpp.hpp
+++ b/src/fl/gfx/rgbw.cpp.hpp
@@ -243,9 +243,9 @@ constexpr NamedGamut kDciP3D60  = {{0.6800f, 0.3200f}, {0.2650f, 0.6900f},
 // after a gamut switch on the currently active profile.
 namespace { void invalidate_colorimetric_caches_for(const DiodeProfile* profile) FL_NO_EXCEPT; }
 
-void set_input_gamut(DiodeProfile* profile, InputGamut g,
+void set_input_gamut(DiodeProfile& profile_ref, InputGamut g,
                      const float white_xy[2]) FL_NO_EXCEPT {
-    if (profile == nullptr) return;
+    DiodeProfile* const profile = &profile_ref;
     auto apply = [profile](const float r[2], const float gp[2],
                            const float b[2], const float w[2]) {
         profile->input_xy_r[0] = r[0];  profile->input_xy_r[1] = r[1];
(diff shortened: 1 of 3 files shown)


zackees:
All three addressed in 5f6f7dd.

**1. Decayed `white_xy[2]` (Major) — fixed, though not the way suggested.**

You were right that the extent was a false promise, and this repo actually has an explicit ratchet for it (`ci/tools/check_array_params.py` + `array_param_baseline.txt`), so the finding is backed by project policy rather than just style.

I didn't use a span/extent-preserving type, because that can't express the contract: `nullptr` genuinely means "use the gamut's standard reference white", and it's exercised by callers. No signature can encode both "exactly 2 floats" *and* "or null".

The ratchet bans the *misleading array spelling*, not pointers — so `const float* white_xy` satisfies it while keeping nullability honest. The length requirement stays in the comment, which is the only place it can live.

I left the lambda parameters alone: `w` is computed as `white_xy != nullptr ? white_xy : kRec709.xy_w`, so it's a pointer by construction and can't bind to a reference-to-array.

**2. Em dash — fixed.** Replaced with `--`.

**3. Partial assertions — fixed.** All eight coordinates now compared, plus a new case for the 3-argument forwarder, which had an extra argument to lose and no equivalence coverage at all.

---

**Unrelated bug found while verifying this**, being fixed separately: `check_array_params.py` and `check_noexcept.py` pass `text=True` to `subprocess.run` without an encoding, so clang-query's UTF-8 output is decoded with the locale codec. On Windows that raises in the reader thread, leaves `stdout` as `None`, and the check dies on `None + str` — meaning the array-param ratchet silently does not run for Windows contributors. That's why I couldn't get a straight answer out of it at first.

Validated: 357/357 with all examples linking, `bash lint` green.

PR #3750, opened 2026-07-28
fix(esp32): bound the I2S clock divider to its register field (#3745)

### LINKED ISSUE #3745: esp32 I2S: solved clock divider N can silently truncate in clkm_div_num bitfield
Follow-up from #3562 / #3744, where the fractional divider solver was hoisted to `platforms/esp/32/core/clock_divider.h`.

## The gap

`solveFractionalDivider()` returns an integer divider `n` with **no upper bound tied to the destination register**. The I2S peripheral then writes it straight into a hardware bitfield:

```cpp
// src/platforms/esp/32/drivers/i2s/i2s_peripheral_esp32dev_esp.cpp.hpp
i2s->clkm_conf.clkm_div_a   = div_A;
i2s->clkm_conf.clkm_div_b   = div_B;
i2s->clkm_conf.clkm_div_num = div_N;   // <-- narrow field, no range check
```

Assigning an out-of-range value to a C bitfield **silently truncates**. There is no warning, no assert, and no runtime error — the peripheral just runs at the wrong clock, which for a clockless LED driver means malformed WS2812 timing rather than an obvious failure.

## When it bites

`n = base_hz / target_hz`, so low target frequencies produce large `n`. At the 80 MHz APB reference:

| target | solved `n` | fits 8 bits (≤255)? |
|---|---|---|
| 8 MHz (current default) | 10 | yes |
| 1 MHz | 80 | yes |
| 313 kHz | 255 | boundary |
| 100 kHz | 800 | **no — truncates to 32** |

So anything below roughly **313 kHz** wraps. `800 & 0xFF == 32`, i.e. the hardware would run at 80 MHz / 32 = 2.5 MHz instead of 100 kHz — a 25× error, silently.

This is not reachable from today's only caller (the I2S clockless engine passes ≥ 8 MHz, defaulting to 8 MHz), so it is **latent, not an active bug**. It matters now because #3744 turned this into a shared helper that SPI / RMT4 / I2S0 are meant to adopt, each with their own field widths.

## Why the divider limits are asymmetric today

The solver already respects the `A` limit — `max_A = 63` is exactly a 6-bit field — but has no equivalent for `n`. That asymmetry is the actual defect: one hardware constraint is encoded, the sibling one is not.

## Suggested fix

Give the solver a `max_N` bound alongside `max_A`, and have each peripheral pass its own field width. Then either saturate or report failure when the target is unreachable, so a caller asking for an impossible frequency finds out instead of getting a silently wrong clock.

Note this **changes results** for targets below the cap, so it should not ride along with a refactor that claims byte-identical output (which is why #3744 deliberately left it out — its `kMaxDivider` is an `int`-representability guard, not a hardware one).

## Please verify before implementing

The 8-bit `clkm_div_num` / 6-bit `clkm_div_a`/`_b` widths above are from the ESP32 TRM and are corroborated by the existing `A ≤ 63` limit in the code, but I could not confirm them against `soc/i2s_struct.h` from this checkout. Confirm the exact widths per peripheral (I2S1 vs I2S0 vs SPI vs RMT4) before encoding them — they are not necessarily the same across peripherals, which is the whole reason for making it a parameter.

## Acceptance

- `solveFractionalDivider()` takes a `max_N` (or equivalent) and never returns a divider the caller's field cannot hold.
- Unreachable targets produce a documented, detectable outcome rather than a truncated register write.
- Host tests cover the boundary and the just-over case for at least the I2S widths.

zackees:
Closes #3745.

## The defect

`solveFractionalDivider()` returned an integer divider `N` with no upper bound tied to the destination register. The I2S driver assigned it straight into a hardware bitfield:

```cpp
i2s->clkm_conf.clkm_div_num = div_N;   // 8-bit field, no range check
```

C bitfield assignment **silently truncates**. At the 80 MHz APB base a 100 kHz target solves to `N = 800`; into 8 bits that is `800 & 0xFF == 32`, so the peripheral runs at 80 MHz / 32 = **2.5 MHz instead of 100 kHz** — a 25× error with no warning, no assert, and for a clockless LED driver, no symptom other than malformed WS2812 timing. Anything below ~313 kHz (80 MHz / 255) wrapped.

The real defect was the asymmetry: `max_A = 63` already encoded the 6-bit `clkm_div_a` field, while its 8-bit sibling had no equivalent.

## Field widths — verified, as the issue asked

The issue explicitly asked not to take the TRM's word for this. Checked against the vendor header in this checkout, `soc/esp32/include/soc/i2s_struct.h:385-387`:

```c
uint32_t clkm_div_num: 8;
uint32_t clkm_div_b:   6;
uint32_t clkm_div_a:   6;
```

That confirms 255 / 63 / 63, and independently explains where the existing `max_A = 63` came from. Widths are not guaranteed identical across peripherals, which is why this is a parameter rather than a baked-in constant.

## Changes

- `solveFractionalDivider()` takes `max_N`, defaulting to `kMaxDivider` — so **every existing call site keeps its exact previous behaviour**. There is a regression test asserting this for the five real pixel-clock rates.
- `ClkDividerNAB` gains a `saturated` flag, so an unreachable target is *detectable* rather than silently wrong.
- Clamping happens **after** the `A == B` adjustment, because that `++N` can push an otherwise in-range solve one past the field width.
- Hitting the int-representability guard does **not** set `saturated` — that means a sub-hertz request, which is a different failure from a hardware limit.
- The I2S wrapper passes `kI2sMaxDivN = 255` and warns on saturation.

## Tests

Boundary (80 MHz / 255) and just-over cases, per the acceptance criteria. The truncation itself is pinned rather than merely described:

```cpp
const ClkDividerNAB unbounded = solveFractionalDivider(kApbClockHz, 100000u, 63, 2);
FL_CHECK(unbounded.n == 800);
FL_CHECK((unbounded.n & 0xFF) == 32);
```

If someone later changes the solver so this no longer truncates, the test says so.

## Validation

- **357/357** host tests on a clean build; `bash lint` green.
- `bash compile esp32dev --examples Blink` clean.
- The modified TU was **confirmed to actually compile** via an `#error` negative control (8 unity TUs hit it) — not inferred from a passing build, since a file that isn't in the build also produces a passing build.

## Summary by CodeRabbit

- **New Features**
  - Added saturation reporting for ESP32 fractional clock divider solving, indicating when `N` is clamped to fit hardware limits.
  - Extended the divider solver with a configurable `max_N` bound.

- **Bug Fixes**
  - Prevented out-of-range `N` values by enforcing the `max_N` cap and ensuring the chosen divider reflects the closest achievable frequency.
  - Clarified failure/limit behavior so clamping impacts the resulting hardware divider as expected.

- **Tests**
  - Added host-side tests validating `max_N` bounds, default compatibility, and handling of invalid limit combinations.

### DIFF
--- a/src/platforms/esp/32/core/clock_divider.h
+++ b/src/platforms/esp/32/core/clock_divider.h
@@ -36,6 +36,13 @@ struct ClkDividerNAB {
     int n;
     int a;
     int b;
+    /// True when `target_hz` was too low to reach with the caller's `max_N`
+    /// and `n` was clamped to it. The returned triple is then the closest
+    /// achievable clock, NOT the requested one. Callers that care about
+    /// accuracy must check this: writing a clamped `n` is still correct
+    /// register-wise, but the peripheral will run faster than asked
+    /// (FastLED#3745).
+    bool saturated;
 };
 
 /// Largest integer divider the solver will report. The real hardware fields
@@ -53,11 +60,24 @@ constexpr int kMaxDivider = 2147483000;
 /// @param base_hz  Reference clock feeding the divider (e.g. kApbClockHz).
 /// @param target_hz Desired output frequency. Must be non-zero — see below.
 /// @param max_A    Largest denominator the hardware accepts (inclusive).
-///                 I2S allows 63.
+///                 I2S allows 63 (6-bit `clkm_div_a`).
 /// @param min_N    Smallest integer divider the hardware accepts. I2S needs 2.
+/// @param max_N    Largest integer divider the caller's destination field can
+///                 hold. I2S `clkm_div_num` is 8 bits, so 255. Defaults to
+///                 `kMaxDivider`, i.e. "no hardware limit, just keep the
+///                 result representable" — so existing callers that do not
+///                 pass a width keep their previous behaviour exactly.
 ///
 /// @return The `(N, A, B)` minimising `|target_hz - base_hz/(N + B/A)|` within
-///         those limits.
+///         those limits, plus a `saturated` flag.
+///
+/// Bounding `N` matters because the result is usually assigned straight into
+/// a narrow hardware bitfield, and C bitfield assignment **silently
+/// truncates**. At an 80 MHz base a 100 kHz target solves to N = 800; writing
+/// that into 8 bits leaves 800 & 0xFF == 32, so the peripheral would run at
+/// 2.5 MHz — a 25x error with no warning, no assert, and for a clockless LED
+/// driver, no symptom other than malformed output. Saturating at `max_N` and
+/// reporting it keeps the failure visible (FastLED#3745).
 ///
 /// A `target_hz` of 0 has no meaningful answer (the divider would be
 /// infinite), so it is clamped to `min_N` with no fractional part rather than
@@ -65,9 +85,16 @@ constexpr int kMaxDivider = 2147483000;
 /// substitute it themselves before calling — that keeps the policy choice at
 /// the call site instead of buried in shared math.
 inline ClkDividerNAB solveFractionalDivider(u32 base_hz, u32 target_hz,
-                                            int max_A = 63,
-                                            int min_N = 2) FL_NO_EXCEPT {
+                                            int max_A = 63, int min_N = 2,
+                                            int max_N = kMaxDivider) FL_NO_EXCEPT {
+    // A max_N below min_N cannot describe real hardware; treat the floor as
+    // authoritative so the result stays something the peripheral accepts.
+    if (max_N < min_N) {
+        max_N = min_N;
+    }
+
     ClkDividerNAB out;
+    out.saturated = false;
     if (target_hz == 0 || base_hz == 0) {
         out.n = min_N;
         out.a = 1;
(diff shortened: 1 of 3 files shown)


zackees:
Good catch — accepted in full, and it was a regression my own commit introduced rather than a pre-existing gap.

**The core mistake:** `max_N` bounds `N`, but the divisor is `N + B/A`. The largest representable divisor is `max_N + (max_A-1)/max_A` = **255 + 62/63 = 255.984**, not 255. Returning `max_N/1` the moment `ratio` crossed `max_N` cost two things:

1. Targets in the **312520–313725 Hz** band (`80MHz/255.984` .. `80MHz/255`) are genuinely reachable as `N=255` plus a fraction. My early return threw the fraction away *and* mislabelled them saturated — making the solver strictly worse in that band than before I added the bound.
2. Out-of-range targets got a needlessly coarse triple: 100 kHz returned `255/1` instead of the closest bounded `255 + 62/63`.

**Fix:** clamp `N` before the fractional sweep instead of returning early. An out-of-range target then leaves a residual > 1, and the existing sweep naturally settles on the largest legal fraction — so the closest bounded triple falls out of the same code path with no special case. The post-`++N` fallback now also uses the maximal fraction rather than `1/0`.

**Warning message:** now derives the clock actually programmed from the solved triple rather than quoting `base / N`, which understated the divider by nearly a full count. `u64` there because `base * a` overflows `u32` (80e6 × 63 = 5.04e9).

**Tests:** 100 kHz must now yield exactly `(255, 63, 62)`, plus new coverage for the 313 kHz fractional band the early return mishandled.

One test changed meaning as a result, worth flagging rather than burying: `max_N below min_N` previously asserted `a == 1, b == 0`. That expectation was an artifact of the coarse clamp, not the property being tested — with the fraction preserved, the closest triple there is `2 + 62/63`. It now asserts what the case was actually about: `n` never drops below `min_N` to chase an unreachable target, because an N the hardware cannot encode is worse than a slow clock.

Validated: 357/357 on a clean build, `esp32dev` compile clean, lint green.

Issue #3492, opened 2026-07-01
lint: promote SingletonElisionChecker to hard-fail (Tier 3, ratchet)


zackees:
Part of #3481 (Tier 3, ratchet closure).

## Task

Once Tier 1 (#3484 #3485 #3486 #3487 #3488) and Tier 2 (#3489 #3490 #3491) are all merged and the `SingletonElisionChecker` warn output is empty (modulo the allowlist for cached-pointer intentional statics), promote the checker from **warn-only** to **hard-fail** in CI.

## Concrete steps

1. Confirm `bash lint` reports zero `SingletonElisionChecker` warnings on `master`.
2. Confirm the allowlist in `ci/lint_cpp_rs/src/checkers/singleton_elision.rs` covers only the intentional cached-pointer variables (typically `s_XXX_state` for each Tier 2 driver) — each with an inline comment justifying why it can't be inside a Singleton.
3. Change the checker's severity in `processor_registry_cli.rs` from warn to error.
4. Update `agents/docs/cpp-standards.md` to document the rule: "namespace-scope variables MUST be wrapped in `fl::Singleton<T>` unless annotated with `[[gnu::used]]` / `FL_KEEP` or on the allowlist".
5. Update `.coderabbit.yaml` if there's a mirror rule that should also promote.

## Acceptance criteria

- `bash lint` fails a PR that introduces a new namespace-scope non-allowlisted variable.
- Existing violations are all fixed (no allowlist bypass for laziness).
- Documentation updated so authors know the rule.

## When to close

Do not close this issue until all Tier 1 + Tier 2 sub-issues are ✅.

zackees:
Recommend **re-scoping this rather than promoting as written.** Following the finding in #3754 / #3486, I audited all 39 violations the checker currently reports.

**39 of 39 are constant-initialized PODs. Zero are dynamic-init objects.**

Every flagged declaration is `T* x = nullptr`, `bool x = false`, `int x = <literal>`, `const char* x = "lit"`, or a zero-init aggregate with constant NSDMIs — all `static` or anonymous-namespace, i.e. already dropped by the compiler when unreferenced, before `--gc-sections` even runs. Promoting to hard-fail today would force 39 migrations that each *cost* flash and RAM and lose race-free constant-initialization, and zero that gain anything.

### The signal is currently inverted

Verified in `src/fl/stl/asio/http/server.cpp.hpp`:

```cpp
[...]

Issue #3486, opened 2026-07-01
singleton: migrate fl/audio/detector state (Tier 1c)


zackees:
Part of #3481 (Tier 1c).

## Task

Migrate the mutable state (not the constants — those belong to #3483) in `src/fl/audio/detector/*` to `fl::Singleton<T>`.

## Files touched

- `src/fl/audio/detector/equalizer.cpp.hpp` (6 hits — but `kBassStart` etc. are constants, they migrate under #3483 to `constexpr`)
- `src/fl/audio/detector/frequency_bands.cpp.hpp` (4 hits — same split: constants → #3483, mutable state → this issue)
- any other detector `.cpp.hpp` files flagged by the scanner

## Approach

1. Confirm which hits are true mutable state vs. compile-time constants (some equalizer/frequency_bands hits will vanish once #3483 lands; only the truly-mutable ones remain here).
2. Wrap the remaining mutable state in a state struct per detector, access via `fl::Singleton<DetectorState>::instance()`.
3. `bash test --cpp audio` to exercise the audio pipeline.
4. `bash bloat esp32dev --examples Blink` — confirm audio state is now elided from a non-audio sketch.

## Acceptance criteria

- Detector `.cpp.hpp` files clean of `SingletonElisionChecker` warnings.
- Audio tests pass.
- Elision proven — Blink flash size drops by the audio-state footprint after this lands.

zackees:
Worked this in #3754, but the outcome is different from what the issue asks for: **all 11 flagged sites in `fl/audio/detector` turned out to be false positives.** I'd suggest not closing this as "migrated."

Breakdown of the 11:

- **9 were compile-time constants** (`kBassStart`..`kTrebleEnd`, `kNumBands`, `kFFTMinFreq`, `kFFTMaxFreq`). Anonymous namespace, never address-taken, constant-folded to zero storage at `-O1`+. Nothing to migrate.
- **2 were genuinely mutable** (`sVibeFFTCount`, `sFrequencyBandsFFTCount`). I did implement the `fl::Singleton` wrap for these, then measured it and reverted it.

Why the wrap was reverted:

1. **It saves nothing.** `ci/boards.py:795` already passes `-ffunction-sections -fdata-sections`, so a file-scope `static int` gets its own `.bss` section and `--gc-sections` already drops it when the detector isn't linked.
[...]

Issue #3483, opened 2026-07-01
lint: split integer-constant globals into PreferConstexprChecker


zackees:
Part of #3481 (Tier 0).

## Task

The scan (`ci/scan_singleton_elision.py`) reports ~30 hits that are NOT true Singleton candidates — they are integer / const-int compile-time constants that should be `constexpr` (compiler will inline, no storage emitted). Bucketing them under the Singleton warning conflates two very different fixes.

## Examples

```cpp
// src/platforms/esp/32/drivers/channel_manager_esp32.cpp.hpp
int PRIORITY_FORCE   = 4;   // should be `constexpr int`
int PRIORITY_PARLIO  = 3;
int PRIORITY_LCD_RGB = 2;
// ... 14 total in this file

// src/fl/audio/detector/equalizer.cpp.hpp
const int kBassStart   = 20;   // should be `constexpr int`
const int kBassEnd     = 250;
// ... 6 total in this file

// src/platforms/arm/rp/rpcommon/spi_hw_manager_rp.cpp.hpp
int PRIORITY_SPI_HW_8 = 3;
```

## Concrete steps

1. Create `ci/lint_cpp_rs/src/checkers/prefer_constexpr.rs` as a sibling to `singleton_elision.rs`.
2. Match namespace-scope `int` / `const int` / `unsigned` etc. definitions with a **trivial integer literal RHS** (int literal, hex, binary, single-token bit-shift like `(1u << 5)`).
3. Emit warning: "namespace-scope `NAME` is a compile-time integer constant; use `constexpr` so the compiler inlines it and no storage is emitted".
4. Move the trivial-literal-with-non-constexpr detection OUT of `SingletonElisionChecker` so the two rules don't overlap.
5. Register in `processor_registry_cli.rs` (warn-only).

## Acceptance criteria

- `bash lint` reports ~30 `PreferConstexprChecker` warnings distinct from ~85 `SingletonElisionChecker` warnings.
- No overlap between the two rules.
- Fixing a hit by adding `constexpr` silences both.

zackees:
This issue is now almost entirely resolved — the ~30 integer constants it was filed against are gone from the checker output.

**Neither cited example still appears.** Current `fastled-lint --checker singleton_elision` on master (34 findings) contains no hits in either file:

[...]

PR #3758, opened 2026-07-29
fix(esp32): make MCPWM_TIMER_TAG a compile-time constant (#3483)

zackees:
Closes out the last remaining item of #3483.

## The change

`static const char* MCPWM_TIMER_TAG` is a **mutable pointer** — it occupies storage of its own in addition to the string literal it points at. As `static constexpr const char*` it emits nothing unless its address is taken.

Verified safe: the tag is only ever *read*, as the first argument to `ESP_LOGE`/`ESP_LOGI`/`ESP_LOGW`. There is no assignment to it anywhere in the file, so nothing depended on the pointer being mutable.

## Most of #3483 was already done

The issue was filed against ~30 integer constants. **None of its cited examples still appear** in `fastled-lint --checker singleton_elision` on master:

| Cited file | Cited hits | Actual now |
|---|---|---|
| `channel_manager_esp32.cpp.hpp` (`PRIORITY_FORCE` etc.) | 14 | **0** |
| `equalizer.cpp.hpp` (`kBassStart`, `kBassEnd`) | 6 | **0** |

The original count came from `ci/scan_singleton_elision.py`, which still reports 101 where the Rust checker reports 34. That gap is mostly #3754's real literal tokenizer removing false positives — so 34 is the truer number, and the constant-splitting work has largely already happened. Full accounting is in [the issue comment](https://github.com/FastLED/FastLED/issues/3483).

## One thing I deliberately did NOT do

`objectfled_spi_mode.cpp.hpp:100` is still flagged:

```cpp
static constexpr u32 kSpiMaxOutputWords = kSpiMaxInputBytes * kSpiWordsPerByte;
```

It is **already `static constexpr`** — already what a "split constants out" fix would produce. It's flagged only because the triviality test accepts a *literal* RHS but not one computed from other constants.

This blocks **#3492** (promote to hard-fail at zero): the finding can't be driven to zero, because "wrap in `fl::Singleton<T>`" is not meaningful advice for a compile-time scalar.

I prototyped skipping `constexpr` **scalars** while keeping `constexpr` **tables** flagged (an array is far more likely to be ODR-used through a pointer, and then really does occupy flash) — then **reverted it**. `singleton_elision_still_flags_non_literal_constexpr` in `lint_core/tests.rs` deliberately pins the current behaviour for exactly those RHS forms, added under #3486. That's a documented decision, so I raised it on the issue rather than flipping it unilaterally.

## Validation

- 357/357 host tests, `bash lint` green
- Checker no longer reports `MCPWM_TIMER_TAG`
- **Not yet locally confirmed:** this is ESP-only code, so the 357 host tests do not
  exercise it. A local `esp32dev` compile was still running when this was opened, so
  I am relying on CI's ESP32 job for that leg rather than claiming a green build I
  have not seen. The change is a `const char*` -> `constexpr const char*` on a symbol
  that is only ever read, so the risk surface is a reassignment somewhere -- and I
  grepped: there is none.

### DIFF
--- a/src/platforms/esp/32/drivers/gpio_isr_rx/mcpwm_timer.cpp.hpp
+++ b/src/platforms/esp/32/drivers/gpio_isr_rx/mcpwm_timer.cpp.hpp
@@ -73,7 +73,10 @@ using fl::u32;
 // Constants
 // ============================================================================
 
-static const char* MCPWM_TIMER_TAG = "mcpwm_timer";
+// constexpr, not `const char*`: the latter is a mutable pointer that occupies
+// storage of its own on top of the string literal. As a compile-time constant
+// it emits nothing unless its address is taken (FastLED#3483).
+static constexpr const char* MCPWM_TIMER_TAG = "mcpwm_timer";
 
 // MCPWM configuration
 #define MCPWM_GROUP_ID 0


zackees:
Local `esp32dev` compile finished: **`All compilations completed successfully in 16m:57s`**, exit 0.

Being precise about what that does and doesn't prove, since a green build of a file that isn't in the build proves nothing:

- I did **not** run an `#error` negative control on this TU (it would cost another ~17 min for a one-line change that CI already compiles).
- What I did verify is that both guards wrapping the file resolve true for `esp32dev`:
  - `FASTLED_RMT5` → `1` (auto-detected: `ESP_IDF_VERSION >= 5.0.0 && FASTLED_ESP32_HAS_RMT`)
  - `SOC_MCPWM_SUPPORTED` → `1` in the vendor `soc_caps.h` for classic ESP32
- The compile log has 0 literal mentions of `gpio_isr_rx`, which is expected — the driver goes through the `_build.cpp.hpp` unity aggregator, so individual filenames don't appear as separate compile lines.

So: the guards say it compiles, and the build is green. CI's ESP32 job is the authoritative check for that leg, and I'm treating it as such rather than claiming more than I measured.

The functional risk remains what the PR body says — a reassignment of the tag somewhere — and there is none; all 10 uses are read-only `ESP_LOGE`/`ESP_LOGI`/`ESP_LOGW` first arguments.

PR #3809, opened 2026-08-01
test(gfx): assert hue ramps are continuous instead of printing them (#668)

### LINKED ISSUE #668: fill_rainbow() - Unexpected red pixel
I'm using the builtin fill_rainbow function for generating a rainbow spectrum on an LED strip. The issue is, I get a random red pixel midway through the yellow region. I've tried it on 3 different Arduino nano's, and two different LED-strips. Here's the code:

```
#include <FastLED.h>

#define DATA_PIN 2
#define NUM_LEDS 114
CRGB leds[NUM_LEDS];

uint8_t hue = 0;

void setup() {
  Serial.begin(115200);
  FastLED.addLeds<WS2812B, DATA_PIN, GRB>(leds, NUM_LEDS);

  fill_rainbow( leds, NUM_LEDS, hue, 1);
  for(int i=0; i<NUM_LEDS; i++) {
    Serial.print(i);
    Serial.print(' ');
    Serial.print(leds[i].r);
    Serial.print(' ');
    Serial.print(leds[i].g);
    Serial.print(' ');
    Serial.println(leds[i].b);
  }

  CHSV hsv;
  CRGB rgb;
  //rgb = hsv;
}


void loop() {
  fill_rainbow( leds, NUM_LEDS, hue, 1);
  FastLED.show();
}
```

Here pixel number 60 gets the rgb code 160, 0, 0. When it should have something like 160, 150, 0.

The strangest part of all is that I can get rid of the problem in two ways, either by uncommenting the line rgb = hsv;, or by commenting out the fill_rainbow function in the loop function. Any ideas whats going wrong?

zackees:
Closes #668.

## The finding

`tests/fl/hsv2rgb_accuracy.cpp` had a test case named **"HSV to RGB Conversion - Hue Sweep Test"**. It contained **no assertion** — 22 lines of `FL_WARN` printing 64 sampled rows, then nothing. It could not have failed for any input, on any platform, ever.

It also stepped the hue by 4:

```cpp
// Increased step from 1 to 4 for performance (256 -> 64 iterations = 75% reduction)
// Still provides excellent hue sweep coverage with 64 samples
for (int hue = 0; hue < 256; hue += 4) {
```

A single bad hue is the *entire* failure mode here, and sampling steps right over it. So the test was doubly unable to catch what it nominally covered.

## Why that matters for #668

#668 (2018) reported one unexpected red pixel partway through the yellow region of a 114-LED strip filled with `fill_rainbow(leds, 114, hue, 1)`, reproduced across three Nanos and two strips.

`hsv2rgb_rainbow` is a hand-written piecewise function: eight 32-hue sections selected by bit tests on the hue, joined by magic constants (`K85`/`K170`/`K171`) and compile-time switches (`Y1`/`Y2`/`G2`/`Gscale`). Nothing structurally forces one section to end on the value the next one starts from — a wrong constant surfaces as exactly one off-colour pixel mid-ramp.

## The change

Walk every adjacent hue pair, including the 255→0 wrap, and bound the per-channel step, for all three converters. Plus a case reproducing #668's exact call.

Measured maxima at full saturation/value:

| converter | max per-channel step |
|---|---|
| rainbow | 6 |
| spectrum | 4 |
| fullspectrum | 6 |

Bound set at 8 — room to nudge a constant without hiding a real break, which is tens.

## Verified the test can fail

The point of this PR is a test that fails on the bug it targets, so I injected one. Changing `K171` → `151` in the O→Y section:

```
WARN: rainbow: discontinuity of 21 between hue 31 and 32 (allowed 8)
FAIL: HSV to RGB Conversion - hue ramps have no discontinuity
FAIL: fill_rainbow has no isolated off-colour pixel (issue #668)
```

It pinpoints the exact section boundary and the magnitude, and both cases fire. Reverted before commit.

## Status of #668 itself

**It does not reproduce.** The current ramps are continuous everywhere — sections join within 6/255 at every boundary. The colour math is sound today; this pins it so it stays that way.

## Validation

| check | result |
|---|---|
| `bash test --cpp` | 275/275 unit, 83/83 examples pass |
| `bash lint` | clean |
| mutation test | fails as shown above |

Tests-only change; no `src/` diff.

## Summary by CodeRabbit

* **Tests**
  * Expanded color accuracy testing across all HSV-to-RGB conversion methods.
  * Added exhaustive checks for smooth color transitions between adjacent hues.
  * Added rainbow animation continuity coverage across a 114-LED sequence.
  * Tests now identify excessive per-channel color changes and fail when configured limits are exceeded, helping prevent visible color flicker or discontinuities.

### DIFF
--- a/tests/fl/hsv2rgb_accuracy.cpp
+++ b/tests/fl/hsv2rgb_accuracy.cpp
@@ -6,6 +6,7 @@
 #include "fl/stl/pair.h"
 #include "fl/stl/utility.h"
 #include "fl/stl/vector.h"
+#include "fl/gfx/colorutils.h"
 #include "chsv.h"
 #include "crgb.h"
 #include "test.h"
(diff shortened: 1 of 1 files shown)


zackees:
Correcting a claim in the description above, since leaving it would be worse than the gap it papers over.

I wrote that #668 "does not reproduce" and that "the colour math is sound." That is wrong about the cause. #668 **did** reproduce, and #3808 — merged shortly before this — is its actual fix: `hsv2rgb_rainbow` has three paths that call `scale8_video_LEAVING_R1_DIRTY` and then return without `cleanup_R1()`, leaving AVR's `__zero_reg__` non-zero in violation of the ABI. That explains every reported symptom, including why the bad pixel moved under recompilation and why it showed on three different Nanos.

What I actually measured was that the ramps are continuous **in the native build**. That is a different property, and it could never have caught #668: the R1 hazard is AVR-only inline-asm ABI damage, invisible to a host-compiled test.

So, to be precise about what this PR does and does not do:

- **Does**: replace an assertion-free `FL_WARN` loop with a real check, and pin section-boundary continuity of the three hue ramps.
- **Does not**: cover the AVR `LEAVING_R1_DIRTY` / `cleanup_R1()` contract. Nothing here would fail if a fourth such path were introduced tomorrow.

That second gap is the more valuable target and wants a different mechanism — a lint or AST check pairing every `*_LEAVING_R1_DIRTY` call against a `cleanup_R1()` on all reachable exits, rather than a runtime test. Filing that separately.

The continuity guard still stands on its own merits (verified failing via the injected-constant mutation above); it just should not be read as coverage of #668's root cause.

Issue #743, opened 2019-02-22
Gradient palettes get stored in RAM on ESP8266


Aircoookie:
Hello!
I've noticed that if I use the following syntax to define a palette:

```
DEFINE_GRADIENT_PALETTE( heatmap_gp ) {
  0,     0,  0,  0,   //black
128,   255,  0,  0,   //red
224,   255,255,  0,   //bright yellow
255,   255,255,255 }; //full white
```

the palette gets stored in RAM, not PROGMEM, even though the docs say so.

In `platforms/esp/8266/led_sysdefs_esp8266.h` I found this definition:

```
// Default to NOT using PROGMEM here
#ifndef FASTLED_USE_PROGMEM
# define FASTLED_USE_PROGMEM 0
#endif
```

I have some ~2kB worth of palette definitions in my application and would like to free up some memory.
Can someone explain the issue of enabling PROGMEM on ESP8266?

kriegsman:
Thank you for pointing this out; I think it’s probably just an oversight and something we can change easily. I’ll take a look later today if I can.

Aircoookie:
Thanks for your quick response! I just tried `# define FASTLED_USE_PROGMEM 1` in my sketch.
Unfortunately, it seems like the default of disabling it for esp8266 was in fact intentional.
It compiles just fine and reduces the memory usage as expected, but as soon as a palette stored in PROGMEN is selected, the application crashes (using latest 2.5.0 esp8266 core and FastLED 2.3.6):

```
Exception 9: LoadStoreAlignmentCause: Load or store to an unaligned address
Decoding 10 results
0x402080b6: CRGBPalette16::operator=(unsigned char const*) at C:\Users\cschwinne\Documents\Arduino\libraries\FastLED/colorutils.h line 836
0x40269a7d: lwip_init at ?? line ?
0x4020966a: WS2812FX::handle_palette() at C:\Users\CSCHWI~1\AppData\Local\Temp\arduino_build_5212\sketch/WS2812FX_fcn.cpp line 699
0x40100370: millis at C:\Users\cschwinne\AppData\Local\Arduino15\packages\esp8266\hardware\esp8266\2.5.0\cores\esp8266/core_esp8266_wiring.c line 183
0x40100370: millis at C:\Users\cschwinne\AppData\Local\Arduino15\packages\esp8266\hardware\esp8266\2.5.0\cores\esp8266/core_esp8266_wiring.c line 183
[...]

PR #3818, opened 2026-08-01
fix(esp8266): unbreak all ESP8266 builds by guarding on SPI.h

zackees:
## Bug

`bash compile esp8266` fails on current master, for **any** sketch:

```
lib/FastLED/fl/system/sd/fs_sdcard_arduino.cpp.hpp:12:10:
fatal error: SPI.h: No such file or directory
```

The Arduino SD backend is gated on:

```c
#if !defined(FL_IS_TEENSY) && FL_HAS_INCLUDE(<SD.h>) && FL_HAS_INCLUDE(<fs.h>)
```

Both resolve on ESP8266 — `fs.h` is the core's *own* filesystem header, and the core ships an SD library — so the guard opens. But the body then includes `<SPI.h>` unconditionally, and SPI isn't on the include path unless a sketch pulls it in.

Net effect: the SD translation unit failed to compile on **every ESP8266 build**, whether or not the sketch has anything to do with SD cards.

## Fix

Add `FL_HAS_INCLUDE(<SPI.h>)` to the guard, so it tests what the body actually requires.

This can only make the block compile in *fewer* configurations, never more — any platform newly excluded is one where `<SPI.h>` is absent, and there the body could not have compiled anyway.

## Verification

| check | result |
|---|---|
| `bash compile esp8266 --examples ColorPalette` | **hard failure → 269475 B flash / 28012 B RAM** |
| `bash compile teensy41 --examples Blink` | ✅ 72704 B, unchanged |
| `bash test --cpp --clean` | ✅ 359/359 |
| `bash lint` | ✅ clean |

**Not re-verified locally: esp32dev and uno.** The fbuild daemon wedged partway through this session (`daemon did not become healthy after 3 spawn attempts`). I confirmed those same commands fail identically on **unmodified master**, so the failures are environmental, not caused by this change — flagging it rather than implying coverage I don't have. CI will cover both.

Found while investigating #743, which needs a working ESP8266 build to assess.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved SD-card compatibility checks to prevent builds when the required SPI support is unavailable.

### DIFF
--- a/src/fl/system/sd/fs_sdcard_arduino.cpp.hpp
+++ b/src/fl/system/sd/fs_sdcard_arduino.cpp.hpp
@@ -6,7 +6,11 @@
 #include "fl/system/sd/fs_sdcard_arduino.h"
 #include "platforms/is_platform.h"
 
-#if !defined(FL_IS_TEENSY) && FL_HAS_INCLUDE(<SD.h>) && FL_HAS_INCLUDE(<fs.h>)
+// SPI.h is checked alongside SD.h/fs.h because the body below includes it
+// unconditionally. On ESP8266 both <SD.h> and <fs.h> resolve (fs.h is the
+// core's own filesystem header), so the guard passed while <SPI.h> was not on
+// the include path -- breaking every ESP8266 build, SD-card sketch or not.
+#if !defined(FL_IS_TEENSY) && FL_HAS_INCLUDE(<SD.h>) && FL_HAS_INCLUDE(<fs.h>) && FL_HAS_INCLUDE(<SPI.h>)
 
 // IWYU pragma: begin_keep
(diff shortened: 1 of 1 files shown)


zackees:
Closing as a duplicate — #3815 landed the identical fix while this was in flight, adding `FL_HAS_INCLUDE(<SPI.h>)` to the same guard for the same reason.

Same diagnosis independently: ESP8266 satisfied the `<SD.h>`/`<fs.h>` half of the condition (its core ships an unrelated `<fs.h>` for the flash filesystem) while `<SPI.h>` wasn't on the include path, so the SD translation unit failed and took every ESP8266 build with it — SD sketch or not.

No action needed here; master already has it.

Issue #3864, opened 2026-08-06
chore(rp2350w): hand off final #3832 hardware validation


zackees:
## Context

The RP2350W source and deployment work is ready for final board validation, but the attached hardware is not currently usable for HIL.

- Parent bring-up: #3832
- Implementation PR: #3863 (`feat/rp2350w-btstack-transport`, head `33cf3d6524e85bfbd9653d92b9939dec19d7a4bc`)
- Required fbuild deployment hardening was merged in FastLED/fbuild#1263 (`2976d863`).
- Source validation on the implementation branch passed: targeted pytest (184 tests), `bash lint`, and `bash compile rp2350w --examples AutoResearch` using the local fbuild development executable.

The board scan last showed an RP2350-associated COM17 as `health=phantom selectable=no` (USB ID `2E8A:F00F`, serial `2DCB876B587EA334`). Windows also reported an unidentified device on a root port with Code 43; fbuild cannot safely associate that unidentified USB device with the phantom COM port. COM9 was a healthy ESP32-S3, not the required ESP32-C6 network peer.

A firmware watchdog can recover a running firmware stall, but cannot repair a failed/absent host USB enumeration. Do not use raw serial tools or direct flashers; deploy only through fbuild/autoresearch after the device is healthy and selectable.

## Handoff procedure

1. Re-establish the RP2350W as a healthy, selectable serial device (try a known-good data cable/root port and inspect the physical wiring, including the GPIO 0/1 test bridge). Confirm the real ESP32-C6 peer and port as well.
2. Run the following fbuild-backed HIL checks, substituting the confirmed C6 port:

```sh
bash autoresearch rp2350w --watchdog-soak --upload-port COM17 --timeout 120s
bash autoresearch rp2350w --flex-io --rp-pio-index 2 --tx-pin 0 --rx-pin 1 --upload-port COM17 --timeout 120s
bash autoresearch rp2350w --net-peer --upload-port COM17 --peer-environment esp32c6 --peer-upload-port <C6_PORT> --timeout 120s
bash autoresearch rp2350w --ble --upload-port COM17 --timeout 120s
bash autoresearch rp2350w --net-peer --ota --upload-port COM17 --peer-environment esp32c6 --peer-upload-port <C6_PORT> --timeout 360s
```

[...]

PR #3865, opened 2026-08-06
fix(ci): install libslirp0 for QEMU jobs

zackees:
## Problem

`esp32s3_qemu_rmt / qemu_test` fails with exit code 127 before running any test:

```
qemu-system-xtensa: error while loading shared libraries:
libslirp.so.0: cannot open shared object file: No such file or directory
QEMU ESP32S3 test-emu failed: failed: QEMU exited with code 127
```

Observed on https://github.com/FastLED/FastLED/actions/runs/31091411573/job/92583052211.

## Cause

The Espressif `qemu-system-*` binaries that fbuild downloads are dynamically linked
against `libslirp.so.0` (user-mode networking) in addition to `libSDL2-2.0.so.0`.
`.github/workflows/qemu_template.yml` installed only `libsdl2-2.0-0`, and
`ubuntu-latest` does not ship `libslirp0` by default. The failure is intermittent
across jobs only because the runner image occasionally has it pulled in transitively.

## Fix

Add `libslirp0` to the existing runtime-library install step in the shared QEMU
workflow template. This single change covers all three consumers:
`qemu_esp32c3_test.yml`, `qemu_esp32dev_test.yml`, `qemu_esp32s3_test.yml`.

## Validation

CI is the test for this change — the QEMU jobs on this PR exercise the modified
step directly. No source files are touched.

Refs #3864

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Espressif QEMU workflow failures by installing required runtime libraries.
  * Added documentation explaining QEMU’s dynamic dependencies and missing-library errors.

### DIFF
--- a/.github/workflows/qemu_template.yml
+++ b/.github/workflows/qemu_template.yml
@@ -42,12 +42,14 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Install QEMU runtime libraries (libSDL2)
+      - name: Install QEMU runtime libraries (libSDL2, libslirp)
         # The Espressif QEMU binary that fbuild downloads is dynamically
-        # linked against libSDL2-2.0.so.0 on Linux.
+        # linked against libSDL2-2.0.so.0 and libslirp.so.0 on Linux.
+        # Without libslirp0 the binary aborts at startup with
+        # "libslirp.so.0: cannot open shared object file" (exit 127).
         run: |
           sudo apt-get update -y
-          sudo apt-get install -y --no-install-recommends libsdl2-2.0-0
+          sudo apt-get install -y --no-install-recommends libsdl2-2.0-0 libslirp0
 
       - name: Pin python version
         run: echo "3.11" > .python-version


zackees:
### Note on scope of this fix — one open question

`libslirp0` is the right thing to install regardless, but I want to flag an asymmetry so this isn't over-claimed:

In the failing run ([31091411573](https://github.com/FastLED/FastLED/actions/runs/31091411573)), `esp32s3_qemu_rmt` died with `libslirp.so.0: cannot open shared object file` (exit 127) while **`esp32s3_qemu_lcd` passed** — same workflow file, same template, same `ubuntu-latest` image, same `qemu-system-xtensa`. A plain missing-system-package would have failed both identically.

That points at the Espressif QEMU tarball, which bundles its own `lib/libslirp.so.0` and resolves it via rpath: a partial/raced extract in fbuild's toolchain cache would leave one job's copy missing the bundled lib while another job's cache entry is intact. Installing the system `libslirp0` still fixes that case, because the loader falls back to the system path — so this change is a valid mitigation either way, but the underlying cache race (if that's what it is) would belong in fbuild.

The QEMU jobs on this PR exercise the modified step directly; I'll confirm from their results before merging.

Refs #3864

zackees:
Closing unmerged — this fix belongs in fbuild, not here.

fbuild downloads, caches, extracts and invokes the QEMU binary; FastLED never sees the toolchain. Patching `qemu_template.yml` to install a shared library on fbuild's behalf would go stale the moment fbuild bumps the QEMU revision, and it would mask the underlying bug rather than fix it — the same reason deployment is fbuild's job.

The asymmetry I flagged above reinforces it: `esp32s3_qemu_lcd` passed while `esp32s3_qemu_rmt` failed in the same run on the same image, which a missing distro package cannot explain. That points at a raced/partial extract of the Espressif tarball's bundled `lib/libslirp.so.0` in fbuild's toolchain cache — squarely fbuild's territory.

Filed as **FastLED/fbuild#1266**, with atomic toolchain extraction as the suggested fix so the whole class is addressed. That issue also proposes fbuild provision the QEMU runtime deps outright, which would let us delete the pre-existing `libsdl2-2.0-0` workaround from `qemu_template.yml` too.

Refs #3864

PR #3880, opened 2026-08-17
fix(esp32-c6): validate PARLIO with GPIO ISR RX capture backend (#3586)

### LINKED ISSUE #3586: ESP32-C6: RMT and PARLIO clockless TX not captured on loopback bench — UART/SPI pass on the same jumper (#3576 Phase 7)
Found during the #3576 Phase 7 C6 robustness bring-up (WROOM+C6 dual-device bench, C6 = COM9, loopback jumper GPIO1→GPIO0, `gpioBaseline` passes at boot).

## Symptom matrix (runSingleTest, 10 LEDs, ALL_ONES, RMT-RX capture)

| Driver | Result | Capture detail |
|--------|--------|----------------|
| SPI | ✅ 4/4 byte-exact | |
| UART | ✅ 4/4 byte-exact | |
| RMT | ❌ 0/4 | `captureWaitResult:1` (SUCCESS) but `rawEdgesAfterWait:0`; `show_duration_ms:684` for four 300 µs frames (~170 ms/frame — internal TX-done timeout?) |
| PARLIO | ❌ 0/4 | `captureWaitResult:0` (TIMEOUT), 0 edges; `show_duration_ms:88` (~22 ms/frame vs ~0.5 ms expected) |

Also: `rxBackend:ISR` fails to arm on C6 at all (`captureWaitResult:-1`) — separate bug worth its own look.

## PARLIO evidence gathered (bench probes via peekMem + temp firmware instrumentation)

All taken DURING transmit (post-test reads are useless — `parlio_del_tx_unit` restores pads):
- Engine debug metrics healthy: `isrCount:1, txDoneCount:1, chunksQueued/Completed:1, bytesTotal/Transmitted:30/30, errorCode:0, underruns:0`.
- GPIO1 FUNC_OUT_SEL = 47 (`PARL_TX_DATA0`) with `GPIO_ENABLE` bit set — routing + OEN correct.
- DMA payload is a real waveform (first words `0xF0F0F0F0`, head popcount 417), `bit_count:4224` for 240 LED bits, idle_value 0.
- `PCR_PARL_CLK_TX_CONF = 0x0005001D` → div_num 29 (240 MHz / 30 = 8 MHz) — requested clock achieved per the divider register.
- An in-firmware `GPIO_IN` poll on the jumpered pad during TX sees **269 edges** — the wire is toggling; the RMT capture still records nothing.
- `timing=WS2811-400` (2× slower pulses) also fails → not a marginal pulse-width filter issue.

## Open questions
- Why does `show()` take ~22 ms (PARLIO) / ~170 ms (RMT) per 300 µs frame? Both smell like waits timing out — yet PARLIO's txDone metric increments.
- Why does the RMT-RX capture see UART/SPI waveforms but not RMT/PARLIO ones on the same pins? (Possible interaction: RMT group interrupt sharing when RMT TX and RX run together — but that doesn't explain PARLIO.)
- `parlioStreamValidate` RPC ignores its params (always 16 lanes) and its pin pool yields only 12 valid pins → `channelsOk:false`; needs fixing before C6 lane-sweeps can use it.

## Bench context
- C6 serial is USB-Serial-JTAG; response truncation at the 64-byte FIFO was fixed by `Serial.setTxTimeoutMs(100)` once a host attaches (shipping separately).
- `parlioMetrics` RPC (engine debug counters) shipping separately — used for the evidence above.

Blocks: C6 PARLIO max-lane sweep ± WiFi (#3576 Phase 7). UART/SPI robustness sweeps can proceed meanwhile.

Refs #3576, #2682.

zackees:
Recovered from a stale worktree during a branch cleanup — the commit dates to 2026-07-29 but no PR was ever opened, so the work never reached master. Rebased onto current master (111 commits of drift) and re-verified.

Closes #3586.

## Problem

ESP32-C6 RMT RX does not preserve PARLIO's sub-microsecond low phases, even though the same pad's GPIO input does see them. That makes the RMT capture backend unusable for validating C6 PARLIO clockless TX on the loopback bench — which is the symptom reported in #3586 (UART/SPI pass on the same jumper; RMT and PARLIO do not).

## Changes

**Backend selection** (`src/fl/channels/validation.h`)
- `resolveCaptureBackend()` — routes C6 + PARLIO to the GPIO ISR timestamp backend by default. An explicit caller override stays authoritative, so diagnostics can still force RMT.
- `captureEdgeCapacity()` — sizes the RX edge buffer from the frame under test (2 edges/bit + framing headroom, rounded to a power of two, which GPIO ISR RX requires) instead of from AutoResearch's oversized shared byte buffer. Overflow-guarded at each step, returning 0 rather than wrapping.

**GPIO ISR RX** (`dual_isr_context.h`, `fast_isr.S`, `gpio_isr_rx_mcpwm.cpp.hpp`)
- Adds `gpio_status_w1tc_reg_addr` to `DualIsrContext` so the fast ISR clears the GPIO interrupt status (write-one-to-clear) after sampling each edge.
- Replaces the `lr.w`/`sc.w` atomic increment of `write_index` with a single-producer release-fence + aligned word store, paired with an acquire load on the consumer. There is exactly one producer, so the RMW was never needed.
- Adds `FL_STATIC_ASSERT(__builtin_offsetof(...))` guards for every context offset `fast_isr.S` hardcodes, under `#if defined(__riscv)`.

**Tests** (`tests/fl/channels/validation.cpp`) — 2 new cases covering backend selection (4 combinations incl. the override) and edge-capacity sizing (ISR power-of-two vs RMT byte-derived).

## Verification

- `bash test --cpp` — 279/279 unit, 83/83 examples pass
- `bash lint` — clean
- Rebase onto master applied without conflicts; all added symbols confirmed present afterward

## Reviewer notes

- The offset asserts cover `buffer`(0), `armed`(20), `mcpwm_capture_reg_addr`(24), `gpio_in_reg_addr`(28), `gpio_bit_mask`(32), `gpio_status_w1tc_reg_addr`(36). The four between `buffer` and `armed` — `buffer_size`, `buffer_size_mask`, `write_index`, `read_index` — are not individually asserted, but are bracketed: any size drift among them moves `armed` off 20 and trips that assert. Reordering within the group would not be caught. Worth tightening in a follow-up.
- The asserts are `__riscv`-only, so host CI does not exercise them; they fire on an actual C6 build.
- **Not validated on hardware in this session.** The original commit's premise (C6 RMT RX misses PARLIO's short low phases) came from bench work on 2026-07-29 and has not been re-confirmed against a live board post-rebase. This needs a loopback run on the C6 bench before merge.

## Summary by CodeRabbit

* **Improvements**
  * Improved ESP32-C6 PARLIO capture reliability by automatically selecting the appropriate GPIO ISR backend when needed.
  * Added backend-aware capture buffer sizing with overflow detection to prevent invalid captures.
  * Improved GPIO interrupt handling and synchronization for faster, more reliable signal capture.

* **Tests**
  * Added coverage for backend selection and capture-buffer capacity across supported capture modes.

### DIFF
--- a/examples/AutoResearch/AutoResearchRemoteRunSingleTest.cpp
+++ b/examples/AutoResearch/AutoResearchRemoteRunSingleTest.cpp
@@ -891,6 +891,22 @@ fl::json AutoResearchRemoteControl::runSingleTestImpl(const fl::json& args) {
         }
     }
 
+    // ESP32-C6 RMT RX merges PARLIO's waveform into its frame envelope even
+    // while the GPIO input sees every edge (FastLED#3586). Keep explicit
+    // backend overrides available for diagnostics, but use the independent
+    // GPIO ISR timestamp path for normal C6 PARLIO wire validation.
+#if defined(FL_IS_ESP_32C6)
+    constexpr bool is_esp32_c6 = true;
+#else
+    constexpr bool is_esp32_c6 = false;
+#endif
+    rx_backend_override = fl::validation::resolveCaptureBackend(
+        rx_backend_override, have_backend_override,
+        driver_name == "PARLIO", is_esp32_c6);
+    if (rx_backend_override != fl::RxBackend::PLATFORM_DEFAULT) {
+        have_backend_override = true;
+    }
+
     if (have_backend_override) {
         fl::RxChannelConfig rx_cfg(pin_rx, rx_backend_override);
         rx_channel_to_use = FastLED.addRx(rx_cfg);
--- a/examples/AutoResearch/AutoResearchTest.cpp
+++ b/examples/AutoResearch/AutoResearchTest.cpp
@@ -455,7 +455,12 @@ size_t capture(fl::shared_ptr<fl::RxChannel> rx_channel,
     bool is_uart_driver = (fl::strcmp(driver_name, "UART") == 0)
                        || (fl::strcmp(driver_name, "LPUART") == 0);
     bool is_object_fled_driver = (fl::strcmp(driver_name, "OBJECT_FLED") == 0);
-    rx_config.edge_capacity = rx_buffer.size() * 8;
+    rx_config.edge_capacity = fl::validation::captureEdgeCapacity(
+        rx_buffer.size(), expected_data_bytes, rx_channel->backend());
+    if (rx_config.edge_capacity == 0) {
+        FL_ERROR("[CAPTURE] RX edge-capacity overflow");
+        return 0;
+    }
 #if defined(FL_IS_RP2040) || defined(FL_IS_RP2350)
     // RP's platform-default RX resolves to PIO, while RxChannel retains the
     // requested PLATFORM_DEFAULT enum. Recognize both that public default and
--- a/src/fl/channels/validation.h
+++ b/src/fl/channels/validation.h
@@ -10,6 +10,7 @@
 #include "fl/stl/optional.h"
 #include "fl/stl/int.h"
 #include "fl/stl/noexcept.h"
+#include "fl/channels/rx/types.h"
 
 // Forward declarations for detail modules
 namespace fl {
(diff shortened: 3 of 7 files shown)


zackees:
## ⛔ Do not merge — pending hardware validation

Holding this PR until the premise is confirmed on real silicon.

**Why:** the core claim — that ESP32-C6 RMT RX drops PARLIO's sub-microsecond low phases while the same pad's GPIO input sees every edge — comes from bench work on 2026-07-29 and has not been re-confirmed since. This PR was recovered from a stale worktree and rebased across 111 commits of drift, so the surrounding RX code is not what it was when that observation was made.

**Host tests cannot cover it.** `bash test --cpp` (279/279 unit, 83/83 examples) exercises only the two pure functions, `resolveCaptureBackend()` and `captureEdgeCapacity()`. It never touches `fast_isr.S` or `dual_isr_context.h` — the offset static-asserts are `#if defined(__riscv)`, so they don't even compile on the host.

**What's actually unverified:**
- The `lr.w`/`sc.w` → release-fence + aligned-store change in `fast_isr.S` (correct by single-producer reasoning, but unexercised on hardware)
- The new `gpio_status_w1tc_reg_addr` write-one-to-clear in the fast ISR
- That the GPIO ISR backend captures the edges RMT RX drops — i.e. that this actually fixes #3586

### Gate before merge

```
# with an ESP32-C6 attached to the loopback bench
bash autoresearch --parlio
```

Confirm the GPIO ISR timestamp backend captures PARLIO edges that the RMT backend misses on the same jumper.

### Scope note

Blast radius is the capture/bench path, not LED output: `gpio_isr_rx` reaches production only via `src/fl/channels/rx.cpp.hpp`. `captureEdgeCapacity()` returns `bytes * 8` for non-ISR backends, identical to the previous `rx_buffer.size() * 8`, so only the ISR path changes behavior. A wrong premise degrades C6 bench validation — which #3586 reports as already broken — rather than regressing anyone's strips.

Marked draft to prevent an accidental merge. Ready for review otherwise; CI is running.

Issue #2639, opened 2026-05-31
ESP32 QEMU workflows fail: libSDL2-2.0.so.0 not installed on runner


zackees:
﻿## Symptom

All three ESP32 QEMU workflows are red on master, every run:

- **ESP32-C3 QEMU Test** (workflow `qemu_esp32c3_test.yml`)
- **ESP32-DEV QEMU Test** (workflow `qemu_esp32dev_test.yml`)
- **ESP32-S3 QEMU Test** (workflow `qemu_esp32s3_test.yml`)

Most recent failure: run [26700353290 (esp32c3 / Test)](https://github.com/FastLED/FastLED/actions/runs/26700353290) on master @ `1d9a638`. Same signature on the dev / s3 runs.

The fbuild stage that actually invokes `qemu-system-riscv32` (or `qemu-system-xtensa`) crashes immediately on the dynamic-linker check:

```
/home/runner/.fbuild/prod/cache/toolchains/qemu-riscv32/7d8e371b8c95b3f4/esp-develop-9.2.2-20250817/qemu/bin/qemu-system-riscv32:
  error while loading shared libraries: libSDL2-2.0.so.0: cannot open shared object file: No such file or directory
QEMU ESP32C3 test-emu failed: failed: QEMU exited with code 127
```

Exit code 127 = ELF interpreter / dynamic loader failure — happens before QEMU's own argv parsing, so no QEMU flag (`-display none` / `-nographic` / etc.) helps.

## Root cause

The Espressif QEMU binary that fbuild downloads on demand (under `~/.fbuild/prod/cache/toolchains/qemu-*/.../qemu/bin/qemu-system-*`) is dynamically linked against `libSDL2-2.0.so.0`. The `ubuntu-latest` GitHub runner doesn't ship that library — Ubuntu 24.04's default tooling images don't bundle SDL2 user-mode libs. Result: dynamic linker can't resolve `libSDL2-2.0.so.0` and bails before QEMU even starts.

Confirmed identical failure across all three workflow runs — same root cause, same line, same exit code.

## Fix

Install `libsdl2-2.0-0` (the runtime shared library, not the `-dev` package — we don't compile against it) once at job start in the shared template `qemu_docker_template.yml`. Adds ~1.5 MB and < 2 s on a fresh runner.

```yaml
- name: Install QEMU runtime libraries (libSDL2)
  run: |
    sudo apt-get update -y
    sudo apt-get install -y --no-install-recommends libsdl2-2.0-0
```

[...]

PR #3964, opened 2026-08-22
ci(qemu): install libslirp0 so Espressif QEMU can start

zackees:
## Problem

All three README QEMU badges (ESP32-DEV, ESP32-C3, ESP32-S3) are red. Every leg fails identically:

```
emulator error: package error: QEMU at .../qemu/bin/qemu-system-xtensa cannot start: a required shared library is missing.
.../qemu-system-xtensa: error while loading shared libraries: libslirp.so.0: cannot open shared object file: No such file or directory
```

The Espressif QEMU 9.2.2 binaries fbuild downloads are dynamically linked against `libslirp.so.0`, which `ubuntu-latest` (24.04) does not preinstall. Affects both `qemu-system-xtensa` (esp32dev, esp32s3) and `qemu-system-riscv32` (esp32c3).

This is **not** an fbuild version problem — the pin is current. fbuild's "The cached QEMU toolchain appears incomplete or corrupt / rm -rf ..." message is a misleading diagnostic: the download is intact, the host is missing a system library. Same class of failure as #2639 / #2640 (libSDL2).

## Fix

Add `libslirp0` next to the existing `libsdl2-2.0-0` in the shared `qemu_template.yml` apt step.

## Verification

`.github/workflows/qemu_template.yml` is in the `pull_request` path filter of all three QEMU workflows, so this PR triggers all of them — green here is the proof.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QEMU runtime setup by installing required SDL2 and SLIRP libraries.
  * Added documentation for shared libraries needed by downloaded Espressif QEMU binaries.

### DIFF
--- a/.github/workflows/qemu_template.yml
+++ b/.github/workflows/qemu_template.yml
@@ -42,12 +42,15 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Install QEMU runtime libraries (libSDL2)
+      - name: Install QEMU runtime libraries (libSDL2, libslirp)
         # The Espressif QEMU binary that fbuild downloads is dynamically
-        # linked against libSDL2-2.0.so.0 on Linux.
+        # linked against libSDL2-2.0.so.0 and libslirp.so.0 on Linux.
+        # Neither is preinstalled on ubuntu-latest (24.04); without them
+        # qemu-system-xtensa / qemu-system-riscv32 fail at exec with
+        # "error while loading shared libraries".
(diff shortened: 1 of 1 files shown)


zackees:
Nope this is the wrong fix.

File an issue to update code rabbit configuration that missing so/dll files related to anything fbuild related needs to be handled by fbuild and not patched by a github action which just hides the error without enforcing the solution upstream

Issue #3836, opened 2026-08-02
rp2040 and rp2450(w) board names should be in fbuild - not local resolution!!!


zackees:
these usb id identifiers should be built in to fbuild NOT fastled.

fbuild should get these identifiers from fastled/boards through the normal build process.

Please do not push this here. Both of the rp2040 and rp2350 and w variants should be ingested this way.

Also add this to the agent docs that this is the proper way to get these board name resolutions.

You may merge this if you do a follow up

zackees:
Groundwork for this issue: **#3965**.

I audited every USB VID:PID literal in the `ci/` tree against the live FastLED/boards artifact (`usb-vids.proto.zstd`, 36 vendors / 1056 products). **14 of 15 already resolve upstream — including all four RP2xxx identities this issue names:**

| VID:PID | Identity | Registry |
| --- | --- | --- |
| `2E8A:000A` | rp2040 / rpipico application CDC | resolved |
| `2E8A:F00A` | rpipicow application CDC | resolved |
| `2E8A:000F` | rp2350 / rpipico2 application CDC | resolved |
| `2E8A:F00F` | rp2350w / rpipico2w application CDC | resolved |
| `2E8A:0003` | RP2 ROM BOOTSEL | resolved |

So the data side of "should be in fbuild - not local resolution" is **already done upstream**. `ENVIRONMENT_TO_VCOM_VID_PIDS` is not filling a registry gap for RP — it is shadowing published data.

#3965 documents the rule (`agents/docs/usb-vid-pid-registry.md`), freezes both legacy tables, adds `ci/util/audit_usb_registry.py` so this stays verifiable, and cascades the pin to fbuild 2.5.19 — which selects the exact runtime RP board profile from the verified registry rather than a local guess (FastLED/fbuild#1335).

**What remains for this issue** is the code change #3965 deliberately does not make: `_resolve_port_and_environment` (`ci/autoresearch/phases.py:1320`) still only skips local port detection for RP when `--rpc-smoke` or `--watchdog-soak` is set, so driver mode (e.g. `--parlio`) still calls `auto_detect_upload_port`. The RED test asserting the correct behavior already exists on branch `fix/3836-fbuild-rp-port-resolution` (`test_rp2xxx_driver_mode_delegates_port_selection_to_fbuild`) and fails for all four RP environments.

One unrelated gap the audit turned up: FTDI FT232H `0403:6014` is absent registry-wide — FastLED/boards#60.

[...]

Issue #3838, opened 2026-08-02
ci(teensy41): all-examples link omits Audio and Adafruit_NeoPixel implementations


zackees:
The 	eensy41 all-examples workflow is still red on master, but the current failure set extends beyond the SPI-only scope tracked in FastLED/FastLED#3775.\n\n## Current master evidence\n\n- Master run: https://github.com/FastLED/FastLED/actions/runs/30733721958\n- Commit: d5c784cfda03bb621b093c93303bc16e11a3d912\n- Workflow input: 	eensy41 all\n- fbuild: 2.5.6\n- Result: failure\n\nThe same classes of link failures also appear on unrelated PR #3837, confirming that they are baseline failures rather than RP2350 changes.\n\n## Failure classes\n\n### Teensy Audio core symbol is absent\n\nMultiple Audio examples compile and then fail at link:\n\n`\nundefined reference to set_audioClock(int, long, unsigned long, bool)'\ncollect2: error: ld returned 1 exit status\n`\n\n### Adafruit_NeoPixel implementation is absent\n\nThe compatibility-driver example compiles against the header but fails at link:\n\n`\nundefined reference to Adafruit_NeoPixel::Adafruit_NeoPixel(unsigned short, unsigned char, unsigned short)'\nundefined reference to Adafruit_NeoPixel::begin()'\nundefined reference to Adafruit_NeoPixel::show()'\nundefined reference to Adafruit_NeoPixel::~Adafruit_NeoPixel()'\n`\n\n### SPI is still absent in some examples\n\nThe same run also retains the SPI link failures already documented in #3775, including SPIClass::begin(), SPIClass::transfer(...), and the global SPI.\n\n## Why this is separate from #3775\n\n#3775 is specifically framed around FastLED's Teensy LPSPI drivers and the missing Teensy SPI library. This issue tracks the broader current CI symptom: multiple framework-library implementations are omitted even when their headers are present, so fixing only the FastLED SPI proxy does not restore the 	eensy41 all workflow.\n\n## Acceptance criteria\n\n- ash compile teensy41 all links the Audio examples, the Adafruit_NeoPixel compatibility example, and SPI-dependent examples from a cold cache.\n- The 	eensy41 GitHub Actions workflow passes on master.\n- The fix does not add per-example include/link workarounds solely to appease the build scanner.\n\nRelated: #3775 and FastLED/fbuild#1214.

zackees:
Reopening — I closed this prematurely.

[...]

PR #3965, opened 2026-08-22
docs(usb): make FastLED/boards the documented VID/PID source of truth

zackees:
Refs #3836.

## Why

USB board identity belongs in [FastLED/boards](https://github.com/FastLED/boards),
which publishes a zstd-compressed protobuf (`usb-vids.proto.zstd`) that fbuild
ingests both at build time and as its offline fallback. FastLED consumes it
through fbuild. Nothing in this repo should ever be the place a VID:PID *first
exists* — but until now that rule was not written down here, and two local
tables had grown as a result.

That is the shape #3836 is complaining about, stated as a rule rather than as a
one-off cleanup.

## What this does

**Documents the rule.** New `agents/docs/usb-vid-pid-registry.md` covers the
publication pipeline, the protobuf schema (and why one nested artifact yields
both the VID→vendor and VID:PID→{vendor,product} projections, so a consumer
never needs its own map), what to do when an identity is missing, and the fbuild
version-cascade procedure. Added to `CLAUDE.md` as a Core Rule plus a task-table
row, and cross-linked from `hardware-autoresearch.md` and
`driver-bringup-postmortems.md`.

**Freezes the legacy tables at the point of temptation.**
`ENVIRONMENT_TO_VCOM_VID_PIDS` (`ci/util/port_utils.py`) previously said *"Add
new entries here when porting to a new board"* — the exact instruction that
produced the drift. Both it and `BOARD_FINGERPRINTS`
(`ci/util/serial_probe.py`) now carry a DO-NOT-EXTEND note pointing at the
registry and the cascade.

**Makes the migration verifiable rather than asserted.** New
`ci/util/audit_usb_registry.py` fetches the live artifact, decodes it, and exits
non-zero while any audited literal is unresolvable:

```
$ uv run python ci/util/audit_usb_registry.py
FastLED/boards registry: 36 vendors, 1056 products
Auditing 15 literals from ci/
  OK   2E8A:000F  rp2350 / rpipico2 application CDC  -> Arduino / Shared USB identity (4 products)
  ...
  GAP  0403:6014  FTDI FT232H  [serial_probe.py]
14/15 resolve from FastLED/boards
```

`AUDITED_LITERALS` is the checklist; it should shrink to empty as literals
retire.

## Audit result

**14 of 15 literals already resolve upstream** — including every RP2xxx identity
#3836 names. The lone gap is FTDI FT232H `0403:6014`, absent registry-wide (the
FTDI vendor entry carries only `6001` and `6010`). Filed as FastLED/boards#60
and tracked there rather than papered over here, which is exactly the behavior
the new rule prescribes.

## Version cascade

Pin moves to **fbuild 2.5.19**, which selects the exact runtime RP board profile
from the verified registry instead of a local VID/PID guess
(FastLED/fbuild#1335) and adds ClearCore SAME53 builds (FastLED/fbuild#1333).
Confirmed published on PyPI before moving the pin; `uv run fbuild --version`
reports 2.5.19.

Note `uv.lock` is gitignored in this repo, so the `pyproject.toml` pin is the
only committed half of the cascade — the docs say so explicitly, since relocking
locally is still required or you keep running the previous wheel.

## Testing

- `bash lint` — clean.
- `uv run pytest ci/tests/ -k "port or probe or serial or autoresearch"` — 301
  passed, 4 skipped.
- Audit script exercised end-to-end against the live artifact.

## Scope

Docs, comments, one new audit script, and the pin bump. **No behavior change** —
this PR does not implement #3836's code change (routing RP port selection to
fbuild in `_resolve_port_and_environment`); it establishes the rule and proves
the data is ready for it.

## Summary by CodeRabbit

* **New Features**
  * Added a USB VID/PID registry auditor to identify missing, unresolved, or outdated board records.
  * Improved handling of registry data, including tracked upstream gaps and compatibility with future formats.

* **Documentation**
  * Documented centralized USB identity management, board bring-up checks, update workflows, and audit results.
  * Clarified that existing USB tables are frozen and reserved for compatibility and diagnostics.

* **Chores**
  * Added ignore rules for temporary build and compilation files.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -116,6 +116,9 @@ examples/**/*.ino.cpp
 ./*.log
 .loop
 .fbuild/*
+.fbuild/
+.compile-tmp/
+.tmp/
 
 # WASM frontend build (Vite)
 src/platforms/wasm/compiler/dist/
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -8,6 +8,7 @@
 |------|------|
 | Writing/editing C++ code | `agents/docs/cpp-standards.md` |
 | Verifying a peripheral exists on a chip before writing driver code | `agents/docs/peripheral-existence.md` (halt on phantom — do not fabricate missing `<Peripheral>_Type` in vendor headers) |
+| Adding or changing a board USB VID/PID | `agents/docs/usb-vid-pid-registry.md` (it lands in FastLED/boards first — never as a new literal in `ci/`) |
 | Defining a register map / accessing MCU peripherals | `agents/docs/register-maps.md` (use vendor CMSIS PAL — do not hand-roll shims) |
 | Looking up MCU datasheets / user manuals | Prefer https://github.com/FastLED/datasheets when available, then vendor primary documentation |
 | Creating an API wrapper type | `agents/docs/cpp-standards.md` → "API Object Pattern" |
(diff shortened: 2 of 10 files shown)


zackees:
**CI note:** the `teensy41 / build / build` failure here is **pre-existing on master**, not caused by this PR.

It is `Fx/FxSdCard` failing to link with `undefined reference to 'SPIClass::begin()'`. The identical failure — same example, same symbol — is present on master `9d29d77966` (2026-08-21, job 96636629459), which ran against the *previous* fbuild 2.5.18 pin, so the 2.5.19 bump in this PR is not implicated. Tracked in #3838, where I've added the evidence.

Every other check is green.

**Review fixes applied** in `69dc752f` (three findings from a code review of the first commit):
- `zstandard` was reachable only transitively (via `clang-tool-chain-bins`) — now declared in `pyproject.toml`, so the documented audit command can't break out from under us.
- The protobuf decoder rejected wire types 1/5, which would have turned a forward-compatible upstream schema addition (a checksum or timestamp field) into a total "artifact undecodable" failure while every audited VID:PID was still resolvable. Now skipped; 3/4/6/7 still raise.
- Added `KNOWN_GAPS` so the audit exits 0 while a gap already filed upstream is outstanding — previously the documented "check at any time" command was guaranteed red, making it unusable in CI. An unfiled gap still exits 1, and so does a `KNOWN_GAPS` entry that has since been published, so the allowlist can't quietly mask a resolved gap.

Verified each path directly: fixed32/fixed64 injection still decodes to the same 36 vendors; unfiled-gap and stale-entry paths both exit 1; the real run exits 0.

PR #3967, opened 2026-08-22
fix(hooks): root hooks via CLAUDE_PROJECT_DIR, not git rev-parse

zackees:
Replaces the `cd "$(git rev-parse ...)"` cwd-rooting trick in hook commands with `${CLAUDE_PROJECT_DIR:-$(git rev-parse --show-toplevel)}`.

The old form breaks when the session cwd drifts: outside a submodule, `--show-superproject-working-tree` succeeds with empty stdout, the `||` fallback never fires, and `cd ""` silently no-ops. `CLAUDE_PROJECT_DIR` is injected into hook environments and stays at the session project root; rev-parse remains as a fallback for non-Claude frontends.

Ref: zackees/clud#965

## Summary by CodeRabbit

- **Chores**
  - Improved project directory detection for development hooks.
  - Hooks now support an explicitly configured project directory and otherwise use the repository’s top-level directory.
  - Updated all configured session, tool, notification, and validation hooks consistently.

### DIFF
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -6,7 +6,7 @@
         "hooks": [
           {
             "type": "command",
-            "command": "cd \"$(git rev-parse --show-superproject-working-tree 2>/dev/null || git rev-parse --show-toplevel 2>/dev/null || echo .)\" && uv run python ci/hooks/check_forbidden_commands.py"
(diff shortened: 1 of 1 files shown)


zackees:
Superseded: master already replaced the rev-parse rooting with the pyproject.toml walk-up scheme.

PR #3976, opened 2026-08-22
fix(usb): land the registry simplification stranded by #3965

zackees:
Follow-up to #3965, which merged one commit too early.

## Why this exists

#3965 was merged before its final commit landed on the branch. Master therefore
still carries the version of the audit script that the follow-up work replaced:
a `zstandard` dependency in `pyproject.toml` and a hand-rolled protobuf
wire-format decoder in `ci/util/audit_usb_registry.py`. This lands that stranded
commit, plus two review findings from #3965.

## 1. Drop the protobuf decoder and the zstandard dependency

FastLED/boards publishes the same catalogue at two endpoints — `usb-ids.json`
(plain JSON) and `usb-vids.proto.zstd` (zstd-compressed protobuf). fbuild reads
the protobuf because it wants the compact binary form. That choice got copied
into a Python script where it bought nothing: reading the compressed protobuf
required a new dependency *and* a hand-written varint / wire-type decoder to
parse data that is also published as plain JSON.

The script now reads the JSON, is standard-library only, and is ~90 lines
shorter. `zstandard` is removed from `pyproject.toml`.

## 2. Publish the FT232H gap instead of tracking it

We own FastLED/boards, so a missing identity is a push, not a ticket. FTDI
FT232H `0403:6014` is now published as a curated record on the `other` data
branch (FastLED/boards#60, closed). The audit reports **15/15** and exits 0.

That also retires the `KNOWN_GAPS` allowlist, which existed only to keep CI
green around a gap that could simply be closed.

Docs in this repo, fbuild, and boards now all say: push to boards immediately,
verify with `extract_other.py` first, run `Build site`. A data-only change needs
no fbuild release — fbuild fetches the catalogue at runtime.

## 3. Review findings from #3965

**Do not infer a registry gap from an unnamed port scan** (major). The bring-up
step in `hardware-autoresearch.md` said an unnamed `fbuild port scan` means the
identity is missing upstream. It doesn't — a stale cache, an old fbuild, and a
real gap are indistinguishable from that output, so the guidance routed the
first two to FastLED/boards as false gaps. It now lists all three causes and
requires ruling out the first two (`fbuild --version` against the pin, then the
audit against the published catalogue) before concluding the third. This also
contradicted the caveat the registry doc already carried.

**MD040** (minor) — the pipeline diagram's code fence is now typed `text`.

A third finding asked to convert `_read_varint` / `_iter_fields` to typechecked
dataclasses. Skipped: those functions no longer exist — it reviewed the
pre-simplification commit.

## Testing

- `uv run python ci/util/audit_usb_registry.py` → 15/15, exit 0
- `bash lint` → clean

## Summary by CodeRabbit

- **Documentation**
  - Clarified how to identify and register unknown USB devices.
  - Added guidance for distinguishing stale local data from missing board records.
  - Updated verification, publication, site-refresh, and cache-refresh instructions.
  - Simplified registry updates by removing unnecessary release and dependency-locking steps.

- **Bug Fixes**
  - USB registry audits now use the published JSON catalogue.
  - Unresolved device identities are reported as failures.
  - Improved handling of malformed catalogue entries.
  - Confirmed all currently tracked device literals resolve successfully.


zackees:
Review triage:

| Finding | Disposition |
|---|---|
| Audit is not an arbitrary identity lookup (Major) | **Valid — filed #3996.** The instruction is genuinely broken: an unknown pair from a port scan is not in `AUDITED_LITERALS`, so the audit never inspects it and exits 0. Needs a `--lookup VID:PID` mode; too large to bolt on here. |
| Keep the procedure in the project root (Minor) | **Valid** — the doc uses `cd .extern-repos/boards`, against the repo rule. Will fix inline with `git -C`. |
| Make the cascade rule conditional everywhere (Minor) | **Valid** — I updated some passages to say a data-only change needs no fbuild release but left lines 9-10, 39 and 123 describing the cascade as unconditional. Self-inconsistent; will fix inline. |

Also filed from this line of work: #3997 (retire the legacy tables now that 15/15 resolve upstream — `AUDITED_LITERALS` shrinking to empty is that issue's definition of done).

PR #3977, opened 2026-08-22
fix(rp): give fbuild ownership of RP port selection in every mode

### LINKED ISSUE #3836: rp2040 and rp2450(w) board names should be in fbuild - not local resolution!!!

zackees:
Closes #3836.

## The bug

RP port selection was delegated to fbuild only for the stock rpc-smoke and
watchdog-soak transports:

```python
stock_rp2xxx_transport = bool(rp2xxx_environment) and (
    args.rpc_smoke or args.watchdog_soak
)
```

Every other RP run mode — `--parlio` and the rest of the driver modes — still
called `auto_detect_upload_port` and matched a local VID:PID table before
deploying.

That is wrong for exactly the reason it was wrong for the stock transports. An
RP board enumerates as BOOTSEL mass-storage *before* deploy and only publishes
its application CDC endpoint afterwards. A pre-deploy scan is therefore either
racing that transition or matching some other attached board — the failure this
issue is titled after. `fbuild deploy` already discovers RPI-RP2 and returns the
application port, so the local scan is redundant at best and actively wrong when
several boards are attached.

## The fix

**Pre-deploy gate** now covers any RP2xxx environment, renamed
`fbuild_owns_rp_port` since it is no longer about "stock" transports. An
explicit `--upload-port` still wins — it short-circuits detection regardless.

**Post-deploy recovery gate** widened to match. This half is load-bearing, not
cosmetic: driver-mode runs now also reach that code with no pre-deploy port, so
leaving the recovery scoped to rpc-smoke/watchdog-soak would have converted a
deploy that returned no port from a recoverable identity-poll into an outright
`❌ No application serial port was returned` failure. Widening one gate without
the other would have traded a wrong-port bug for a hard-failure bug.

## Test

The assertion already existed as a RED test —
`test_rp2xxx_driver_mode_delegates_port_selection_to_fbuild`, which makes
`auto_detect_upload_port` raise if it is called at all, parameterised across
`rp2350`, `rpipico2`, `rp2350w`, `rpipico2w`. It was failing on all four; it now
passes.

## Testing

- 18 RP tests pass (4 were failing)
- 181 in `test_autoresearch_phases.py`
- 922 passed / 34 skipped across `ci/tests` excluding QEMU
- `bash lint` clean

## Related

Groundwork landed separately: #3965 established FastLED/boards as the USB
identity source of truth and pinned fbuild 2.5.19, whose
`fix(rp): select exact runtime board profile` (FastLED/fbuild#1335) is what
makes fbuild's returned port exact rather than best-effort. An audit against the
published registry confirmed all four RP identities resolve upstream, so nothing
here depends on the local table that #3965 froze.

#3976 lands a follow-up simplification stranded by #3965's merge; it is
independent of this PR.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RP2xxx deployment port handling across all run modes.
  * Explicit upload ports continue to be honored.
  * Missing application CDC ports are now recovered consistently after deployment.
  * Deployment setup is more reliable across RP2040 and RP2350 environments.
* **Tests**
  * Expanded coverage to verify consistent port selection and recovery across supported RP2xxx environments.

### DIFF
--- a/ci/autoresearch/phases.py
+++ b/ci/autoresearch/phases.py
@@ -1319,12 +1319,19 @@ async def _resolve_port_and_environment(ctx: RunContext) -> int | None:
         ctx.final_environment is not None and "teensy" in ctx.final_environment.lower()
     )
     rp2xxx_environment = _active_rp2xxx_environment(ctx.final_environment)
-    stock_rp2xxx_transport = bool(rp2xxx_environment) and (
-        args.rpc_smoke or args.watchdog_soak
-    )
+    # RP port selection belongs to fbuild for EVERY run mode, not just the
+    # stock rpc-smoke / watchdog-soak transports (FastLED#3836). An RP board
+    # enumerates as BOOTSEL mass-storage before deploy and only publishes its
+    # application CDC endpoint afterwards, so a pre-deploy scan here is either
+    # racing that transition or matching some other attached board. fbuild
+    # discovers RPI-RP2 and returns the application port from `deploy`.
+    #
+    # An explicit --upload-port still wins: it is read into `upload_port`
+    # above and short-circuits the detection below regardless of this flag.
+    fbuild_owns_rp_port = bool(rp2xxx_environment)
 
     upload_port = args.upload_port
-    if not upload_port and not stock_rp2xxx_transport:
+    if not upload_port and not fbuild_owns_rp_port:
         expected_environment = None if is_teensy else ctx.final_environment
         max_wait_s = 60
         poll_interval_s = 1.0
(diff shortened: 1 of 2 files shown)


zackees:
Review triage:

| Finding | Disposition |
|---|---|
| Preserve RP board identity during CDC recovery (Major) | **Valid — filed #3998.** Two identical RP boards present the same VID:PID, so `auto_detect_upload_port` cannot separate them and takes the first match. Worth being precise about blame: the ambiguity is pre-existing and inherent to identity-only matching — this PR does not introduce it, it widens the gate so the path is reachable in many more runs. The real fix needs a board handle carried *through* deploy rather than re-derived after, which is fbuild-side work. Failing closed on ambiguity is a cheap interim step, noted in the issue. |
| Include all RP2xxx environments in the test (Minor) | **Valid, and worth stating clearly: the *behaviour* is already correct.** `RP2XXX_ENVIRONMENTS = RP2040_ENVIRONMENTS \| RP2350_ENVIRONMENTS`, so `rp2040`/`rpipico`/`rpipicow` already go through the widened gate. The test just under-covers them. Will extend the parameterisation inline. |
| Annotate `self` (Minor) | **Valid** per repo guidelines. Will fix inline. |

PR #3982, opened 2026-08-22
test(wdt): drive watchdog timeout tests off a mock clock

### LINKED ISSUE #3978: test(wdt): watchdog feed test is timing-flaky on loaded CI runners
## Symptom

`tests/fl/wdt/watchdog.cpp` fails intermittently on CI, macOS runners so far:

```
../../tests/fl/wdt/watchdog.cpp:172: !(fired.load())
../../tests/fl/wdt/watchdog.cpp:173: 1 == 0
[FastLED.watchdog] WARN: nested FL_WATCHDOG_AUTO() detected — only one scoped guard should be alive at a time
[doctest] test cases: 26 | 25 passed | 1 failed
```

Seen on job 97100960253 (PR #3976, a Python/docs-only change that cannot touch
this code). Re-running the same job cleared it — the failure did not reproduce.

## Why it is flaky

The test ("watchdog does not fire when fed") arms a **200 ms** watchdog and
feeds it from a loop that sleeps **50 ms** per iteration:

```cpp
dog.begin(200);
for (int i = 0; i < 10; ++i) {
    fl::this_thread::sleep_for(fl::chrono::milliseconds(50));  // ok sleep for
    dog.feed();
}
dog.disable();

FL_CHECK_FALSE(fired.load());
```

The margin between feed interval and timeout is **4×**, but `sleep_for` only
guarantees a *lower* bound. On a loaded runner a single 50 ms sleep overshooting
to >200 ms — entirely ordinary under CPU contention on shared CI — means the
watchdog fires exactly as designed and the assertion fails. The test asserts a
real-time guarantee the host cannot make.

The stub's 10 ms tick granularity narrows the effective margin further.

## Note on the sibling test

The test above it has the same shape in the other direction — `dog.begin(50)`
then a single 300 ms sleep asserting `fired` **is** set. That one is safe: a
sleep can only overshoot, which only makes firing more certain. The asymmetry is
worth preserving in any fix — the failure mode is one-directional.

## Suggested fix

Options, roughly in order of preference:

1. **Widen the margin substantially** (e.g. `begin(2000)` with 50 ms feeds).
   Cheapest, keeps the test's intent, makes overshoot implausible rather than
   merely unlikely.
2. **Drive the clock deterministically** — if the stub's timer source can be
   injected or advanced manually, the test stops depending on wall time at all.
   Correct fix, more work.
3. **Assert on feed bookkeeping** rather than on not-firing, if the watchdog
   exposes a feed count or last-fed timestamp.

Option 1 alone would likely take this from occasionally-red to effectively
never, without restructuring anything.

## Impact

Low severity, real cost: an unrelated PR shows a red `test / build` that has
nothing to do with its diff, which trains reviewers to ignore that signal.

zackees:
Closes #3978.

## The flake

`tests/fl/wdt/watchdog.cpp` armed a **200 ms** watchdog and fed it from a loop
sleeping **50 ms** per iteration, then asserted it never fired.

`sleep_for` only guarantees a *lower* bound. One 50 ms sleep overshooting 200 ms
on a loaded runner — ordinary under CPU contention — means the correctly-fed
watchdog expires **exactly as designed**, and the test fails for something that
is not a bug. Hit on macOS against #3976, a Python/docs-only PR that cannot
touch this code; a rerun of the same job passed.

The test was asserting a real-time guarantee the host does not offer.

## Why not just widen the margin

That was my first instinct and it is wrong. Going from 4× to 40× makes the
failure less likely, never impossible — the test would still be asserting
something the scheduler is free to violate, and it would still fail
occasionally, just rarely enough to be harder to diagnose.

## Why not `inject_time_provider`

FastLED already has one, and it does not reach here. It covers `millis()`;
`micros()` explicitly opts out:

```cpp
fl::u32 micros() {
    // Note: micros() does not support time injection
    return fl::platforms::micros();
}
```

`fl::chrono::steady_clock::now()` is built on `micros()`, and the stub's
deadline is built on `steady_clock`. Making `micros()` globally injectable
would work, but it freezes the clock for **every** subsystem sharing the test
process — a process-wide behavior change to fix one test, on a hot path.

## What this does

Adds a clock seam to the stub watchdog itself. `fl::testing::setWatchdogClock()`
installs a manual millisecond source used by all three places that consult the
clock — `begin()`, `feed()`, and the worker's expiry comparison. Mocking fewer
than all three would compare a mocked deadline against a real clock.

Default behavior is unchanged real time. `nullptr` restores it, and the tests
install it via an RAII guard so a failing assertion cannot leak the mock into
later cases in the same binary.

Both timeout tests now advance time explicitly. The negative test **cannot**
expire: mock time moves 50 ms per feed against a 200 ms timeout, and nothing
else moves it.

## The one place real time remains

The worker observes the clock on a real 10 ms tick, so the *positive* test still
waits for that observation after advancing. That direction is safe — waiting
longer can only make a pass slower, never produce a false failure — and it is
bounded so a genuine regression fails rather than hangs.

Worth noting the original positive test was already sound for the same reason
(a sleep can only overshoot, which only makes firing more certain). The bug was
one-directional, and this preserves that asymmetry rather than "fixing" both.

## Verification

The risk with mocking is a test that passes vacuously, so I mutation-tested it:
making `feed()` a no-op still fails the negative test at `watchdog.cpp:228`. It
is still detecting the thing it exists to detect.

- 6 consecutive green runs of the watchdog binary
- `bash test --cpp` → 283/283 unit, 83/83 examples
- `bash lint` clean
- Test execution drops from ~0.8 s of sleeping to ~0.6 s total

## Summary by CodeRabbit

- **New Features**
  - Added optional clock overrides for stub-platform timing.
  - Added consistent 64-bit microsecond timing across related time APIs.
  - Added controls to install, clear, and check the active clock override.

- **Bug Fixes**
  - Improved watchdog timeout handling when the millisecond counter wraps around.
  - Watchdog feeds now reliably prevent premature expiration across timer boundaries.

- **Tests**
  - Added deterministic watchdog timeout coverage with controlled time advancement.
  - Reduced timing-sensitive waits through bounded polling.

### DIFF
--- a/src/platforms/stub/platform_time.cpp.hpp
+++ b/src/platforms/stub/platform_time.cpp.hpp
@@ -8,10 +8,12 @@
 #if defined(FASTLED_STUB_IMPL) && !defined(FL_IS_WASM)
 
 #include "fl/stl/thread.h"
+#include "fl/stl/atomic.h"
 #include "fl/stl/compiler_control.h"
 #include "fl/stl/function.h"
 #include "fl/stl/chrono.h"  // platform-specific time implementation
 #include "fl/stl/noexcept.h"
+#include "platforms/stub/time_stub.h"
 
 // Forward declare delay override (defined in time_stub.cpp.hpp)
 extern fl::function<void(fl::u32)> g_delay_override;
@@ -41,22 +43,66 @@ void delayMicroseconds(fl::u32 us) FL_NO_EXCEPT {
     fl::this_thread::sleep_for(fl::chrono::microseconds(us));
 }
 
-fl::u32 millis() FL_NO_EXCEPT {
-    auto current = std::chrono::system_clock::now();  // okay std namespace
-    auto elapsed = std::chrono::duration_cast<std::chrono::milliseconds>(  // okay std namespace
-        current - start_time);
-    return static_cast<fl::u32>(elapsed.count());
+// ---------------------------------------------------------------------------
+// Single root clock.
+//
+// millis() and micros() both derive from micros64() rather than each taking
+// their own independent reading of the host clock. Two independent reads could
+// disagree — micros() could report 1'999'999 while a millis() taken a moment
+// later reports 2000 — and there is only ever one thing to stub.
+//
+// The 64-bit root also keeps the derivation honest: the u32 return types are a
+// truncation applied at the very end (preserving the existing wrap semantics
+// callers expect), not a limit baked into the source of truth.
+// ---------------------------------------------------------------------------
+
+// Test seam. Atomic because it is read from every thread that asks the time —
+// including the stub watchdog's timer thread — while the installing test
+// writes it from the main thread. A plain function pointer (rather than
+// fl::function, as g_delay_override uses) keeps that read lock-free, which
+// matters because micros() sits on hot paths.
+fl::atomic<fl::u64 (*)()>& clockOverrideRef() FL_NO_EXCEPT {
+    static fl::atomic<fl::u64 (*)()> fn{nullptr};
+    return fn;
 }
 
-fl::u32 micros() FL_NO_EXCEPT {
+fl::u64 micros64() FL_NO_EXCEPT {
+    fl::u64 (*fn)() = clockOverrideRef().load();
+    if (fn) {
+        return fn();
+    }
     auto current = std::chrono::system_clock::now();  // okay std namespace
     auto elapsed = std::chrono::duration_cast<std::chrono::microseconds>(  // okay std namespace
         current - start_time);
-    return static_cast<fl::u32>(elapsed.count());
+    return static_cast<fl::u64>(elapsed.count());
+}
+
+fl::u32 millis() FL_NO_EXCEPT {
+    return static_cast<fl::u32>(micros64() / 1000u);
+}
+
+fl::u32 micros() FL_NO_EXCEPT {
+    return static_cast<fl::u32>(micros64());
 }
 
 
 }  // namespace platforms
 }  // namespace fl
+
+// Test-facing clock override, mirroring setDelayFunction/clearDelayFunction.
+// Installing a source freezes host time: it advances only when the test says
+// so, which is what makes timing assertions deterministic instead of a bet on
+// the scheduler. Pass nullptr (or call clearClockFunction) to restore real time.
+void setClockFunction(fl::u64 (*micros64Func)()) FL_NO_EXCEPT {
+    fl::platforms::clockOverrideRef().store(micros64Func);
+}
+
+void clearClockFunction() FL_NO_EXCEPT {
+    fl::platforms::clockOverrideRef().store(nullptr);
+}
+
+bool isClockOverrideActive() FL_NO_EXCEPT {
+    return fl::platforms::clockOverrideRef().load() != nullptr;
+}
 
 #endif  // FASTLED_STUB_IMPL && !__EMSCRIPTEN__
--- a/src/platforms/stub/time_stub.h
+++ b/src/platforms/stub/time_stub.h
@@ -28,4 +28,33 @@ void clearDelayFunction() FL_NO_EXCEPT;
 
 // Check if delay override is active (for fast testing)
 bool isDelayOverrideActive(void) FL_NO_EXCEPT;
+
+// ---------------------------------------------------------------------------
+// Clock override (for deterministic timing tests)
+// ---------------------------------------------------------------------------
+//
+// The stub derives millis(), micros(), and therefore fl::chrono::steady_clock
+// from one root, fl::platforms::micros64(). Overriding that root is enough to
+// put a test in full control of host time — there is no second clock to keep
+// in sync, and the three views can never disagree.
+//
+// This is what timing tests should use instead of sleeping. `sleep_for` only
+// guarantees a LOWER bound, so any assertion that depends on time NOT having
+// passed is a bet on the host scheduler; under CI load that bet loses (see
+// FastLED#3978). With a frozen clock, elapsed time is whatever the test says.
+//
+// The installed function is called from every thread that asks for the time,
+// so it must be safe to invoke concurrently. Always clear it when done —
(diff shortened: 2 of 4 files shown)


zackees:
**Reworked** — the PR description above describes the first approach; this is what it does now.

Original version added a clock seam to the stub *watchdog*. That was the wrong layer. Two things pushed it to the right one:

**1. Mock at the stub platform, not in the watchdog.** `fl::platforms::micros()` is already inside `#if defined(FASTLED_STUB_IMPL)`, so overriding it costs real firmware nothing and needs no watchdog-specific API. My earlier objection — "global time mocking has too much blast radius" — was about making `fl::micros()` injectable for *all* platforms. Stub-scoped is a different, safer proposition, and `platform_time.cpp.hpp` already had `g_delay_override` ("for fast testing") as precedent, with `setDelayFunction`/`clearDelayFunction` exposed in `time_stub.h`.

**2. One root instead of two.** The stub computed `millis()` and `micros()` from two independent readings of the host clock. That is a coherence hazard in its own right — `micros()` can report `1'999'999` while a `millis()` taken a moment later reports `2000` — and it means two seams to keep in sync. Both now derive from a single `fl::platforms::micros64()`. One function to stub, and the three views (`millis`, `micros`, `steady_clock`) can never disagree.

The u32 returns become a truncation applied at the end, preserving the wrap semantics callers already expect, rather than a limit baked into the source of truth.

**Net effect on the diff:** watchdog production code is now untouched — no seam in `watchdog.h`, none in `watchdog_stub.impl.hpp`. The change is the stub clock root, its test-facing API next to the existing delay trio, and the test.

**Re-verified after the rework:**
- 6 consecutive green runs
- `bash test --cpp` → 283/283 unit, 83/83 examples — this is the load-bearing check, since `millis()`/`micros()` now change for *every* host test
- mutation test repeated: making `feed()` a no-op still fails the negative test, so it has not gone vacuous
- `bash lint` clean

One follow-up worth flagging separately: `steady_clock::now()` is built on the **u32** `fl::micros()`, so it wraps roughly every 71.6 minutes and is not monotonic across that boundary. A deadline computed just before a wrap would never be reached. Short test binaries make it improbable, and it is out of scope here, but the 64-bit root now makes it straightforward to fix.

zackees:
**Second commit added** — the wrap bug I flagged above is now fixed, using the elapsed-delta pattern rather than an absolute deadline.

## The change

The stub stored an absolute deadline and fired on `now >= deadline`. Storing a *future* time is what breaks at a rollover: a deadline computed just below the wrap becomes unreachable once the clock rolls to a small value, the comparison stays false, and the watchdog silently never fires — in exactly the long-running process that most needs it.

Storing the **start** and comparing the **delta** is robust instead:

```cpp
const fl::u32 elapsed = fl::millis() - fed_at;
if (elapsed >= timeout_ms) { ... }
```

The subtraction wraps in the same direction as the clock, so the delta is correct even when `now < fed_at`. This holds as long as the interval fits the type — `FL_WATCHDOG_MAX_TIMEOUT_MS` is 600000 (10 min), far below the u32 millisecond span of ~49.7 days.

Moving the source from `steady_clock` to `fl::millis()` also drops the tighter exposure: `steady_clock` is built on the **u32** `fl::micros()`, which rolls over every ~71.6 minutes rather than every 49.7 days.

## Tests, both directions

- an unfed watchdog **must still fire** across the boundary
- a fed one **must still not**

The second is the one that earns its place: it rules out "fixing" the wrap by making elapsed always look large, which would pass the first test while breaking the watchdog completely.

These are only testable because of the first commit — parking real time three seconds below 2^32 ms is not an option. That is the payoff from mocking at the clock root rather than bolting a seam onto the watchdog: a latent production bug became reproducible in a unit test.

## Verified the tests are meaningful

Restoring the absolute-deadline comparison fails the wrap test at `watchdog.cpp:234`, with the exact signature of the real bug — no fire, no crash, just silence.

`bash test --cpp` → 283/283 unit, 83/83 examples. `bash lint` clean.

## Scope note

This fixes the **stub**. Whether the real per-platform watchdog impls have the same absolute-deadline shape is worth a separate look — I have not audited them, and hardware WDTs usually delegate expiry to silicon rather than computing a deadline in software, so it may well be stub-only.

zackees:
Review triage:

| Finding | Disposition |
|---|---|
| Revalidate the feed timestamp before timeout handling (Major) | **Valid — fixed in `49796166`.** The worker sampled `fed_at_ms` under the lock, released it, then judged expiry, so a `feed()` landing in that gap meant firing on a just-fed watchdog. Read/evaluate/disarm is now one critical section, with callbacks still invoked after release (user code holding `s.mu` would deadlock). Also re-checks `enabled` under the lock, since `disable()` can land after the cheap unlocked check. Worth noting this race **predates the branch** — the original absolute-deadline code had the same read-release-evaluate shape. |
| `src/fl/wdt/watchdog.h` × 2 (legacy macro gate; expose setter via `CFastLED`) | **Stale.** Reviewed an earlier revision. That file was reverted when the seam moved down to the clock root — `watchdog.h` has no diff in this PR now. |
| Expose the clock override through `CFastLED` (`time_stub.h`) | **Skipping.** The Public Settings Pattern governs user-facing runtime settings on the public API. `setClockFunction` is a stub-only test hook alongside the existing `setDelayFunction`/`clearDelayFunction` trio, compiled out of every non-stub build. Putting a test seam on `CFastLED` would put it in the *firmware* API, which is the opposite of the intent. |
| `platform_time.cpp.hpp:85` — `wrap_us = 2**32` | Related to the u32 wrap; the substantive version is **filed as #3993** (`steady_clock` is not monotonic — wraps every ~71.6 min). Out of scope here beyond the stub fix already in this PR. |

Also filed from this work: **#3995** (`fl::millis()` can abort the process when called from a thread during static destruction — the macOS teardown crash CI caught on this branch, worked around here in `8831cd9a`) and **#3994** (time mocking split between `millis()` and `micros()`, which is why this PR needed a stub-level seam rather than the documented `inject_time_provider`).

Issue #4003, opened 2026-08-23
audit: fl/system/file_system.h has drifted from the API Object Pattern


zackees:
## Summary

`src/fl/system/file_system.h` (146 lines) has drifted from the conventions documented in `agents/docs/cpp-standards.md`. It is the header every filesystem-using sketch includes, and it currently mixes the public API, the platform extension point, per-format convenience methods, test-only hooks, and an inline implementation. Requesting an audit and a decision on whether to rectify.

Surfaced while adding a LittleFS backend (#3984) — the natural next step was to add `beginLittleFs()` / `littleFs()` next to `beginSd()` / `sd()`, which would have widened the class again. That was backed out in favor of the existing generic seam, but it made the shape hard to ignore.

### 1. The API object contains implementation

`cpp-standards.md` → API Object Pattern, rule 2: *"The API object wraps, it doesn't implement... contains no core logic itself."*

`file_system.h` ends with a fully inline `loadJpegFromSD()` that constructs a `FileSystem`, calls `beginSd()`, builds an error string, and loads a JPEG. That is core logic living in the public header, and it hard-couples SD to JPEG inside the filesystem API.

### 2. No matching implementation directory

Rule 1: *"One header, one directory, same name. `foo.h` is the public API; `foo/` holds everything behind it."*

There is no `fl/system/file_system/`. The implementation is spread across `fl/system/file_system.cpp.hpp`, `fl/system/sd/`, and (proposed) `fl/system/littlefs/`. The directory naming does not follow the documented pairing, so there is no obvious home for a new backend.

### 3. `FileSystem` accumulates a method per format

The class currently exposes `openVideo`, `openMpeg1Video`, `readText`, `readJson`, `readScreenMaps`, `readScreenMap`, `loadJpeg`, `openMp3`, and `loadFled`. A filesystem class knows about JPEG, MP3, JSON, ScreenMap, Video, and Fled.

Consequences worth weighing:

- every new asset format widens the class
- the header drags `JpegConfig`, `Mp3DecoderPtr`, `json`, `ScreenMap`, `Video`, `Fled`, and `flat_map` declarations into the include graph of anyone who touches a file
- the natural place for `readFoo()` is here, so the pressure is always toward more

An alternative shape is free functions taking a `FileSystem&` (or a `filebuf_ptr`), which keeps the class to open/read/close and lets each format live with its own decoder. Not proposing that as the answer — just naming the axis the audit should decide on.

### 4. Public API and platform extension point share a header

[...]

PR #4005, opened 2026-08-23
docs(review): add a CodeRabbit rule on public API expansion

zackees:
## Why

A public header is a permanent commitment — every symbol added is one FastLED has to keep working. `fl/system/file_system.h` shows what happens without a check on that (#4003): `FileSystem` grew a member per *backend* (`beginSd`, `sd`) **and** a member per *format* (`openVideo`, `openMpeg1Video`, `readText`, `readJson`, `readScreenMaps`, `readScreenMap`, `loadJpeg`, `openMp3`, `loadFled`), and the header now re-exports `fl/fx/video.h` and `fl/codec/jpeg.h` — so every sketch that opens a file also pulls in video and JPEG.

None of that was a bad decision in isolation. It is what a class looks like after a series of locally-reasonable additions with nothing asking "and what about the tenth one?"

## The rule

Four questions, in order:

1. Does an existing generic entry point already do this?
2. Could this be a free function instead of a member?
3. Does the new symbol drag headers with it (especially under `// IWYU pragma: export`)?
4. Does it scale to the tenth case?

## Worked example

The rule ships with the LittleFS backend as a concrete wrong-vs-right pair, because it is exactly the decision that surfaced #4003.

**Wrong** — extend the class:

```cpp
class FileSystem {
  public:
    static FileSystem sd(int cs_pin);
    bool beginSd(int cs_pin);
    static FileSystem littleFs(bool format_on_fail = false);  // NEW
    bool beginLittleFs(bool format_on_fail = false);          // NEW
};
```

Two more members, two more again for the next backend, and every consumer of the header recompiles.

**Right** — a free factory paired with the seam that already existed:

```cpp
FsImplPtr make_littlefs_filesystem(bool format_on_fail = false) FL_NO_EXCEPT;

fl::FileSystem fs;
if (fs.begin(fl::make_littlefs_filesystem())) { ... }
```

`FileSystem` is unchanged. The Nth backend costs one free function and no class surface.

## Scoping

Explicitly **does not apply** to the Public Settings Pattern, which lives on the same `src/fl/**/*.h` glob and requires the opposite — a new `fl::set_*` / `enable_*` / `disable_*` / `use_*` global setter MUST gain a matching `CFastLED` wrapper. That is required widening; without this carve-out the two rules would contradict each other on the same file.

Also excluded: per-object configuration on the object's own API, `fl::detail::` and anonymous namespaces, and non-public headers.

Stated as guidance, not a veto — widening a public class is sometimes correct. The rule asks for the justification in the PR description rather than blocking.

YAML validated with `yaml.safe_load`.

Refs #4003

## Summary by CodeRabbit

* **Documentation**
  * Added code review guidance for changes affecting public APIs.
  * Documented preferred extension patterns, include-graph and scalability considerations, and approved exceptions for specialized implementations.

### DIFF
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -37,9 +37,14 @@ reviews:
         1. Does an existing generic entry point already do this? If a class
            has a `begin(Interface)` seam, a new backend needs a free factory,
            not a new `beginFoo()` member.
-        2. Could this be a free function instead of a member? Free functions
-           are found by ADL, do not enlarge the class, and do not force every
-           consumer of the header to recompile when they change.
+        2. Could this be a free function instead of a member? It does not
+           enlarge the class, so the class stays stable as backends
+           accumulate, and it can move to its own header later so consumers
+           that never call it stop paying for it. Two things it does NOT buy
+           on its own: a free function declared in the same public header
+           still churns every consumer when that header changes, and ADL
+           only applies when a parameter type lives in the namespace — a
+           `bool` does not, so callers still write `fl::make_x(...)`.
         3. Does the new symbol drag headers with it? A method whose signature
            names a codec/format type pulls that header into the include graph
            of everyone who includes this one — especially with
--- a/.coderabbit.yaml
+++ b/.coderabbit.yaml
@@ -26,6 +26,78 @@ reviews:
         peripherals — one engine for both clockless and SPI modes" and
         `agents/docs/cpp-standards.md` → "Parallel-IO Driver: Unified
         Clockless + SPI Engine".
+    - path: "src/fl/**/*.h"
+      instructions: |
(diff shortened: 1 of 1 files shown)


zackees:
Both corrections applied — you were right on each.

**ADL:** `bool` has no associated namespace, so ADL never fires for `make_littlefs_filesystem(bool)` and callers still write `fl::make_x(...)`. Citing it as a benefit of free functions was wrong for this example.

**Recompilation:** a free function in the *same* public header churns every consumer when that header changes. The benefit only appears if the factory later moves to its own header, so the rule now says that explicitly rather than implying it.

The remaining justification — a free function does not enlarge the class, so the class stays stable as backends accumulate — stands on its own and is what the `FileSystem` example actually demonstrates.

Issue #3986, opened 2026-08-22
meta: remote assets — hosting, build-time resolution, and on-device filesystem deployment


zackees:
## Summary

**Meta issue.** Tracks the full body of work for getting large binary assets out of the FastLED source tree and onto devices — remote hosting, build-time resolution, filesystem deployment, and the diagnostics that make misconfiguration legible.

Spans four repos: `FastLED/FastLED`, `FastLED/fbuild`, `FastLED/assets`, `zackees/fastled-wasm`.

## Motivation

`examples/Fx/FxLedmapper32x32/data/` carried ~10.2 MB of binaries (a 6.05 MB `video1.rgb` plus a 4.10 MB zip duplicating it). Binary blobs in git are permanent, re-downloaded on every clone, and bloat every worktree. Investigating how to remove them surfaced a chain of gaps: there was no remote-asset mechanism for example data, no filesystem-image build step, no LittleFS backend in FastLED, and two incompatible `.lnk` formats already in the wild.

## Current state

**Done.** The assets are hosted in `FastLED/assets` with a validated `manifest.json` and CDN serving; `video1.rgb` resolves end-to-end from the CDN with sha256 verification. Not yet done: everything needed to get it onto a device without a physical SD card.

## The chain

```
data/*.lnk
   |
   v  [fbuild lnk pull]  EXISTS - fetches + verifies into global CAS
cache (~/.fbuild/prod/cache/archives/lnk-blobs/)
   |
   v  [materializer]     EXISTS - projects into build tree
build/resources/<rel>/
   |
   +--> [embed_stage]    EXISTS - objcopy into firmware via board_build.embed_files
   |
   +--> [buildfs]        MISSING  <-- #1358
             |
             v  [uploadfs]  MISSING  <-- #1358
        device LittleFS partition
             |
             v  [FsImpl] MISSING  <-- FastLED#3984
        readable from fl::FileSystem
```

### Blocking — decide first

| Issue | Repo | What |
|---|---|---|
| [#1357](https://github.com/FastLED/fbuild/issues/1357) | fbuild | **Normalize the `.lnk` format.** Two incompatible variants exist today; the one file committed to FastLED (`examples/AudioUrl`) cannot be read by fbuild. Everything below annotates or parses `.lnk`, so this should settle first. |

### Core feature work

| Issue | Repo | What |
|---|---|---|
[...]

Issue #3984, opened 2026-08-22
feat: add a LittleFS FsImpl backend — on-chip flash storage is unreachable today


zackees:
## Summary

FastLED has no LittleFS (or SPIFFS) filesystem backend. `fl::FileSystem` can only reach an SD card over SPI, the WASM browser VFS, or a host stub — so on-chip flash storage is unreachable from FastLED code.

## Verified current state

`FsImpl` has exactly three implementations:

| Backend | File |
|---|---|
| Arduino SD over SPI | `src/platforms/fs_sdcard_arduino.hpp:200` |
| WASM browser VFS | `src/platforms/wasm/fs_wasm.cpp.hpp:358` |
| Host stub | `src/platforms/stub/fs_stub.cpp.hpp:28` |

The only mention of LittleFS in the tree is aspirational — `src/fl/stl/detail/file_io.h:19`: *"Embedded (future): Can map to LittleFS, SPIFFS, SD card APIs, etc."*

Consequence: any sketch needing on-device file storage must use a **physical SD card**. `examples/Fx/FxLedmapper32x32/FxLedmapper32x32.ino:64` calls `filesystem.beginSd(CHIP_SELECT_PIN)` for exactly this reason — it is the only hardware option available.

## Why this is a clean drop-in

`FileSystem::begin(FsImplPtr)` (`src/fl/system/file_system.h:55`) already accepts an arbitrary backend, alongside the SD-specific `beginSd(int cs_pin)`. A LittleFS `FsImpl` needs no core API change — it implements the existing virtual interface and is passed to `begin()`.

Scope is contained: implement `FsImpl`'s virtuals over the platform LittleFS API, plus a `make_littlefs_filesystem()` factory mirroring `make_sdcard_filesystem(int cs_pin)`.

## Platform coverage to decide

| Platform | LittleFS availability |
|---|---|
| ESP32 (Arduino-ESP32) | `LittleFS.h` in core |
| ESP8266 | `LittleFS.h` in core (SPIFFS deprecated) |
| RP2040/RP2350 (arduino-pico) | `LittleFS.h` in core |
| Teensy 4.x | `LittleFS_Program` library — program flash, **wiped on re-upload** |
| AVR | none |

The Teensy wipe-on-upload behavior is a genuine semantic difference worth surfacing to users rather than papering over.

## Why now

[...]

Issue #2428, opened 2026-05-09
feat(channels): compile-time Bus binding via templated ChannelConfig (eliminates Channel API binary bloat)


zackees:
## Summary

Introduce a strongly-typed `fl::Bus` enum and a `ChannelConfig<Chipset>` template so each `Channel` is bound to its driver at compile time. This:

1. **Eliminates the Channel API binary bloat from #2421** — drivers only link when their `Bus` is named.
2. **Closes #2167** (`Channel::create<EngineType>(config)` becomes a real template).
3. **Makes bus/chipset mismatches compile errors** instead of runtime warnings.

## Problem

Today, every ESP32 driver (RMT, PARLIO, SPI, I2S, LCD_RGB, LCD_SPI, BLE, UART, …) is unconditionally compiled and registered with `ChannelManager` via `initChannelDrivers()` in `src/platforms/esp/32/drivers/channel_manager_esp32.cpp.hpp`. Driver selection happens at runtime through `IChannelDriver::canHandle()` virtual dispatch, so `--gc-sections` cannot strip unused drivers. A NEOPIXEL-only sketch links every driver and grew from **344 KB → 783 KB** between 3.10.3 and master (see #2420 reproducer).

Additionally, the new Channel API in #2167 has two papercuts:
- `Channel::create<fl::ChannelEngineSpi>(config)` doesn't compile — `create()` isn't a template.
- The string-based `mAffinity = "RMT"` is typo-prone and heap-allocates.

### Single source of truth: `fl::Bus` enum

```cpp
namespace fl {
enum class Bus : fl::u8 {
    AUTO,         // sentinel — pick DefaultBus<Chipset>
    RMT,
    PARLIO,
    SPI,
    I2S,          // LCD_CAM clockless on S3
    LCD_RGB,
    LCD_SPI,
    OBJECT_FLED,  // Teensy
    STUB,
};

// Per-platform default per chipset family
template<typename Chipset> struct DefaultBus;
template<> struct DefaultBus<ClocklessChipset> {
    static constexpr Bus value = /* ESP32-P4: PARLIO, S3/dev: RMT, Teensy: OBJECT_FLED, ... */;
};
template<> struct DefaultBus<SpiChipsetConfig> {
    static constexpr Bus value = /* ESP32: SPI, S3: LCD_SPI, ... */;
};
} // namespace fl
```

### Templated `ChannelConfig<Chipset>` carries the chipset family as a type

```cpp
[...]

PR #4006, opened 2026-08-23
refactor(fs): expose embedded storage generically as fl::getEmbeddedFs()

zackees:
## Reworked after review — the API no longer names a technology

**Previous version:** `fl::make_littlefs_filesystem()`.
**Now:** `fl::getEmbeddedFs()`, with LittleFS demoted to an ESP implementation detail.

Two things were wrong with the first attempt.

**1. It put a technology in the public API.** Once `make_littlefs_filesystem()` is public, every platform has to hold an opinion about "LittleFS" — including the host stub, which just maps to a directory on the developer's disk. Asking it "do you support LittleFS?" is meaningless; "do you have somewhere to put files?" is answerable. LittleFS is not even the only answer where it exists: ESP32 can back flash with LittleFS, SPIFFS, or FatFS, so pinning the public name to one blocks changing the default later. Note `make_sdcard_filesystem()` already got this right — it names the medium, not the driver, and is not called `make_sdfat_filesystem()`. Tracked in #4007.

**2. The backend had drifted out of its entry point's translation unit.** The unity-build checker requires every directory containing a `.cpp.hpp` to have a `_build.cpp.hpp`, and requires its parent to include that — which forced the ESP implementation into `platforms+.cpp`. Every sketch links that object, so the LittleFS chain could no longer be dropped. The split TU exists precisely to make that dead-strippable (#2773 item 1.2), and this silently defeated it.

## Shape

Follows the pattern `fl/system/delay.h` already uses — API header, matching `.cpp.hpp` body, and a `platforms/*.h` dispatch header selecting one plain-header fragment:

```
fl/system/embedded_fs/embedded_fs.h        fl::getEmbeddedFs()
fl/system/embedded_fs/embedded_fs.cpp.hpp  body
platforms/embedded_fs.h                    dispatch
platforms/esp/fs/embedded_fs_esp.hpp       LittleFS (ESP32 / ESP8266)
platforms/shared/embedded_fs_noop.hpp      null elsewhere
fl/build/fl.system.embedded_fs+.cpp        own TU
```

Everything composes into **one** translation unit, so the entry point and its backend are dropped together when a sketch never calls `getEmbeddedFs()` — the mechanism `platforms/channel_drivers.impl.cpp.hpp` documents for #2428 and that `fl.system.sd+.cpp` relies on for #2773.

`file_system.h` is **untouched**. The API lives in its own header rather than widening a class that already carries a member per format (#4003).

```cpp
fl::FileSystem fs;
if (fs.begin(fl::getEmbeddedFs())) {
    fl::ifstream f = fs.openRead("data/frame.rgb");
}
```

## Validation

**The real backend is proven to compile**, which the previous version never established. Placing `#error` in the LittleFS branch made the `esp32dev` build fail — so that branch is genuinely reached against arduino-esp32, not silently falling back. A green build alone proves nothing here, because the fallback compiles identically; three earlier green builds were consistent with the backend never being compiled at all.

Every occurrence traced to a single include chain — `fl.system.embedded_fs+.cpp` → `_build.cpp.hpp` → `embedded_fs.cpp.hpp` → `platforms/embedded_fs.h` → `embedded_fs_esp.hpp` — confirming the backend stays inside the tree-shakable TU rather than leaking into `platforms+.cpp`.

With the probe removed, `bash compile esp32dev --examples Blink` succeeds.

- 4 host tests on the null contract: the factory links with no backend present, returns null, `begin()` reports failure instead of crashing, and a failed mount leaves the `FileSystem` safe to call.
- `bash lint` clean, `bash test --cpp` passes.

## Follow-ups filed

- **#4007** — the generic-naming design, including whether `format_on_fail` belongs in the signature and whether the stub should implement this (which would let host tests exercise real read paths instead of only the null contract).
- **#4008** — consolidating the filesystem subsystem under `src/fl/fs/`. Deciding where each file went was the hard part of this PR, repeatedly; the subsystem is currently spread over `fl/system/`, `fl/stl/`, `fl/asset/`, and `platforms/**`.

Refs #3984, #4007, #4003, #3986

## Summary by CodeRabbit

* **New Features**
  * Added embedded filesystem support for ESP32 and ESP8266 devices using LittleFS.
  * Added a 32×32 LED matrix video playback example with screen mapping, SD-card assets, brightness, and speed controls.
* **Documentation**
  * Added setup details, asset metadata, download instructions, and playback requirements for the video example.
* **Bug Fixes**
  * Improved handling when embedded storage is unavailable, including graceful mounting and playback failures.
* **Tests**
  * Added coverage for unsupported embedded storage and safe filesystem behavior.


zackees:
The review landed against the previous revision; the branch has since been reworked (`make_littlefs_filesystem` → `fl::getEmbeddedFs()`, files moved). Status of each finding against the current code:

**1. `format_on_fail` only formats on ESP32 — fixed.** The doc now names the platforms rather than implying them:

> Honored on ESP32, whose core takes the flag directly. ESP8266 exposes a no-argument `begin()` with no such hook and ignores it — an unformatted filesystem there simply fails to mount, which is the safe outcome either way.

**2. `FASTLED_HAS_LITTLEFS_IMPL` should not introduce a new `FASTLED_*` macro — resolved by removal.** You were right that the prefix is frozen for new names. The restructure deleted the macro outright: dispatch now happens in `platforms/embedded_fs.h`, which selects one platform fragment with `#if defined(FL_IS_ESP32) || defined(FL_IS_ESP8266)`, so nothing probes a capability macro. There is no `FASTLED_*` or `FL_*` capability flag left to name.

**3. Comments describing the fallback as weak — obsolete.** Those comments went with the file. The current no-op fragment (`platforms/shared/embedded_fs_noop.hpp`) is a plain `inline` function in `fl::platforms::`, matching the other `*_noop.hpp` fragments; there is no weak-vs-strong distinction left to describe.

**4. IWYU private pragma on `_build.cpp.hpp` — added.**

## What changed beyond the review

Two structural problems, both worth flagging since they are the kind of thing review would otherwise catch late:

**The API named a technology.** `make_littlefs_filesystem()` put "LittleFS" in the public surface, which forces every platform — including the host stub that just maps to a directory on disk — to hold an opinion about it. It is now `fl::getEmbeddedFs()`, naming the medium, matching `make_sdcard_filesystem()` (which is likewise not `make_sdfat_filesystem()`). Design recorded in #4007.

**The backend had drifted out of its entry point's translation unit.** The unity-build checker requires a `_build.cpp.hpp` per directory containing a `.cpp.hpp`, and requires the parent to include it — which pushed the ESP implementation into `platforms+.cpp`. Every sketch links that object, so the LittleFS chain could no longer be dropped, silently defeating the split-TU tree-shaking (#2773 item 1.2). The layout now follows `fl/system/delay.h`: API header, matching `.cpp.hpp` body, and a `platforms/*.h` dispatch header selecting a plain-header fragment, so everything composes into one TU.

## Validation

The real backend is now **proven** to compile, which the earlier revision never established. Putting `#error` in the LittleFS branch made the `esp32dev` build fail — so that branch is genuinely reached against arduino-esp32. A green build alone shows nothing here, because the no-op fragment compiles identically; three earlier green builds were consistent with the backend never being compiled at all.

Every occurrence traced to a single include chain through `fl.system.embedded_fs+.cpp`, confirming the backend stays inside the tree-shakable TU. With the probe removed, `bash compile esp32dev --examples Blink` succeeds. `bash lint` and `bash test --cpp` pass; 4 host tests cover the null contract.

zackees:
Merged current master into this branch.

**On the samd21 CI failure:** it is not caused by this change. The branch was cut from `3de8d4f8`, where master's own build workflow was failing (#3992). Master is green again at `8e3ec44c`, and the merge pulled in `.github/workflows/build_samd21.yml` — **a workflow that did not exist when this branch was created**. The failure it reports is a platform gap:

```
lib/FastLED/platforms/pin.h:56:6: error: #error "Platform-specific pin implementation
  not defined for this Arduino variant."
lib/FastLED/fl/system/pin.cpp.hpp:52: error: 'digitalWrite' is not a member of 'fl::platforms'
```

That is GPIO support in `fl.system+.cpp`, a different translation unit from anything here, and it has nothing to do with storage. A newly-added CI job surfacing a pre-existing platform gap is the simplest reading, and it fits the evidence.

I tried to confirm by building samd21 on clean master locally, but that run never reached compilation — the 282 MB ARM toolchain download failed with `package error: failed to read response body`. So I am reporting this as strongly-supported rather than proven; if samd21 is green on master's own runs of the new workflow, then something here does need a second look.

Post-merge on this branch: `bash lint` clean, `bash test --cpp` passes.

zackees:
@coderabbitai review

All findings from the previous review are addressed — see the comment above for the per-item status. The branch was substantially reworked since that review (`make_littlefs_filesystem` -> `fl::getEmbeddedFs()`, files relocated under `platforms/esp/fs/`), so most of the original comments no longer apply to code that exists.

zackees:
@coderabbitai resolve

The branch was substantially reworked after that review — `make_littlefs_filesystem` became `fl::getEmbeddedFs()`, and the implementation moved to `platforms/esp/fs/` behind a dispatch header. Most of the original comments describe code that no longer exists.

The findings that survived the rework are applied: `format_on_fail` now names the platforms that honor it, the `FASTLED_*` macro was removed entirely by the restructure, and `_build.cpp.hpp` gained the IWYU private pragma.

Issue #4007, opened 2026-08-23
design: expose on-chip flash storage generically, not as make_littlefs_filesystem


zackees:
## Summary

`FsImplPtr make_littlefs_filesystem(bool)` (proposed in #4006) names a **technology** in FastLED's public API. It should name the **concept** instead, with LittleFS demoted to a platform detail under `src/platforms/`.

This is a narrower and more actionable point than the general audit in #4003: the public filesystem API may grow for convenience, but it must not leak platform vocabulary.

## Why the current shape is wrong

**Every platform inherits the vocabulary.** Once `make_littlefs_filesystem()` is public, the host stub, WASM, Teensy, and AVR all live in a world where "LittleFS" is a concept they must have an opinion about. `platforms/stub/fs_stub.hpp` maps filesystem calls to a real directory on the developer's disk — asking it to answer "do you support LittleFS?" is meaningless. The honest answer for the stub is "I have somewhere to put files," which is the concept, not the technology.

**LittleFS is not even the only answer on the platforms that have it.** ESP32 can back on-chip flash with LittleFS, SPIFFS (legacy), or FatFS. RP2040 uses LittleFS. Teensy 4.x uses `LittleFS_Program` with genuinely different semantics (program flash, wiped on re-upload). Pinning the public name to one of those choices means the API is wrong for its siblings and blocks changing the default later.

**It breaks the symmetry the API already has.** `make_sdcard_filesystem(int cs_pin)` names a *medium* (an SD card) and takes a *property of the medium* (the chip-select pin) — not `make_sdfat_filesystem()` or `make_arduino_sd_filesystem()`. The SD entry point is already technology-neutral; the flash one should match.

## Proposed shape

**Public API — one generic factory, no platform words:**

```cpp
// On-chip flash storage. The platform picks the implementation:
// LittleFS on ESP32/ESP8266/RP2040, LittleFS_Program on Teensy 4.x,
// a scratch directory on host/stub. Returns null where the platform has
// no such storage, so callers get a clean begin() failure.
FsImplPtr make_flash_filesystem(bool format_on_fail = false) FL_NO_EXCEPT;
```

Name is up for discussion — `make_flash_filesystem`, `make_internal_filesystem`, `make_onboard_filesystem`. The requirement is only that it describes *where the bytes live*, not *which filesystem driver* puts them there. "Flash" describes the medium, parallel to "sdcard".

**Platform detail — LittleFS moves under `src/platforms/`:**

```
[...]

PR #4010, opened 2026-08-23
refactor(fs): consolidate the filesystem subsystem under fl/fs/

### LINKED ISSUE #4008: refactor: consolidate the filesystem subsystem under src/fl/fs/
## Summary

Filesystem code is spread across four unrelated parts of the tree with no directory that names the subsystem. Proposal: consolidate under `src/fl/fs/**`.

This surfaced while adding a second storage backend (#3984 / #4007). Nothing about that work was hard; deciding *where each file goes* was, repeatedly — which is the symptom worth acting on.

## Where filesystem code lives today

| Concern | Location |
|---|---|
| Public API + `FsImpl` interface | `fl/system/file_system.{h,cpp.hpp}` |
| SD backend | `fl/system/sd/` |
| Embedded-flash backend | `fl/system/embedded_fs/` (new) |
| `filebuf` — the file handle every backend implements | `fl/stl/detail/file_handle.h` |
| `ifstream` / `filebuf_ptr` | `fl/stl/fstream.h` |
| Asset resolution (`.lnk`, `FL_ASSET`) | `fl/asset/` |
| Platform backends | `platforms/stub/fs_stub.hpp`, `platforms/wasm/fs_wasm.*`, `platforms/arm/teensy/sdfat/**`, `platforms/esp/fs/` |
| Unity TUs | `fl/build/fl.system+.cpp`, `fl.system.sd+.cpp`, `fl.system.embedded_fs+.cpp`, `fl.asset+.cpp` |

`fl/stl/` is the odd one. `filebuf` is the abstraction every backend subclasses — it is the filesystem's core type, sitting in the standard-library-shims directory because it is spelled like `std::filebuf`. Anyone tracing how a backend works has to leave the filesystem code entirely to find its base class.

`fl/system/` is a grab bag by construction: `delay.h`, `pin.h`, `heap.h`, `serial.h`, `yield.h`, `timeout.h`, `engine_events.h` and the filesystem all share it. A subsystem with three subdirectories, four translation units, and platform fragments in five places has outgrown a shared folder.

## Proposed

```
src/fl/fs/
    fs.h                  public API (FileSystem, FsImpl, factories)
    fs.cpp.hpp
    file_handle.h         filebuf — moved from fl/stl/detail/
    fstream.h             ifstream — moved from fl/stl/
    sd/                   SD backend
    embedded/             on-chip flash backend
    asset/                .lnk resolution, FL_ASSET   (or left in fl/asset/)
```

with `src/fl/build/fl.fs*.cpp` for the translation units and platform fragments staying under `src/platforms/**`, reached through a dispatch header exactly as they are now.

## Why now

Two related issues are already open against the current shape, and both get easier with a home for the subsystem:

- **#4003** — `file_system.h` has drifted from the API Object Pattern: it carries an inline `loadJpegFromSD()` body, re-exports `video.h` and `jpeg.h` to every consumer, and has no matching implementation directory. Rule 1 of that pattern is *"one header, one directory, same name"* — `fs.h` + `fs/` satisfies it directly, where `file_system.h` + `fl/system/` never can.
- **#4007** — keeping platform vocabulary out of the public API. A dedicated directory makes the boundary obvious: everything in `fl/fs/` is generic, everything platform-specific is behind a dispatch header.

Doing this before more backends land is cheaper than after. RP2040 LittleFS and any FatFS work would otherwise each repeat the "where does this go?" decision.

## Constraints any restructure must preserve

1. **Per-backend translation units.** `fl.system.sd+.cpp` exists so the linker drops the SD chain (~16 KB on ESP32-S3) when `beginSd()` is never called — #2773 item 1.2. `fl.system.embedded_fs+.cpp` does the same for flash. Renaming is fine; merging them is not.
2. **The entry point must share a TU with its backend.** Splitting them puts the backend in `platforms+.cpp`, which every sketch links, and the tree-shaking silently stops working.
3. **`#include "fl/system/file_system.h"` must keep working**, or this becomes a breaking change for every sketch that opens a file. A forwarding header is the obvious answer; the deprecation timeline is a separate decision.

## Open questions

1. `fl/fs/` vs `fl/filesystem/` — `fs.h` is terse and matches `fl/stl/`, `fl/fx/`, `fl/ui/`; `filesystem.h` is unambiguous. Note the existing comment at the top of `file_system.h`: *"fs.h breaks ESPAsyncWebServer so we use file_system.h instead"* — worth checking whether a `fl/fs/fs.h` path collides the same way before committing to the name.
2. Does `fl/asset/` move in, or stay adjacent? It is a consumer of the filesystem, not part of it.
3. Should `filebuf` / `ifstream` move, given they are deliberately `std`-shaped and live with the other shims?
4. Big-bang move plus forwarding headers, or incremental with both paths live?

## Related

- #4003 — `file_system.h` API Object Pattern audit
- #4007 — generic vs platform-specific naming in the public API
- #3984 — embedded-flash backend
- #3986 — meta

zackees:
## Problem

Filesystem code was spread across four unrelated parts of the tree, with no directory naming the subsystem:

| Concern | Was |
|---|---|
| Public API + `FsImpl` | `fl/system/file_system.{h,cpp.hpp}` — beside `delay.h`, `pin.h`, `heap.h`, `serial.h` |
| `filebuf` — the base class every backend subclasses | `fl/stl/detail/file_handle.h` |
| `ifstream` | `fl/stl/fstream.h` |
| SD backend | `fl/system/sd/` |
| Platform fragments | five places under `platforms/**` |

The `fl/stl/` placement is the tell: `filebuf` is the filesystem's core abstraction, living in the standard-library-shims directory only because it is *spelled* like `std::filebuf`. Tracing how a backend worked meant leaving the filesystem code to find its own base class.

This surfaced while adding a second backend (#3984). Nothing about that work was hard; deciding *where each file went* was, repeatedly, with no precedent to follow.

## Change

```
src/fl/fs/
    fs.h            FileSystem + medium factories
    fs.cpp.hpp
    backend.h       FsImpl — the platform extension point
    read.h          format helpers as free functions
    file_handle.h   filebuf      (from fl/stl/detail/)
    fstream.h       ifstream     (from fl/stl/)
    sd/             SD backend
```

**`FsImpl` gets its own header.** That split is the point: sketch authors and platform implementers previously read one file for unrelated reasons. `backend.h` is now the "teaching FastLED a new medium" header, `fs.h` the "opening files" one.

**`read.h` adds free functions over `FileSystem&`** — `readJpeg`, `readMp3`, `readVideo`, `readJson`, `readScreenMap`, and the rest — forwarding to the existing members, which stay. Adding a format then costs one function there instead of widening `FileSystem`, which today carries a member per format (#4003).

**`getSdFs(cs_pin)`** joins `make_sdcard_filesystem` under a name describing the medium, matching `getEmbeddedFs()` so no driver name reaches the public API (#4007).

## Compatibility

`fl/system/file_system.h` becomes a forwarding header, so all **27 existing includers** — and any sketch in the wild — compile unchanged. Nothing in this PR is a breaking change.

Inverting the delegation (free functions owning the logic, members dropped) is deliberately **not** here: that is the breaking half, and it wants its own PR and deprecation window. Worth noting the members have near-zero internal callers — a survey found `readText` used 3 times and every other format helper 0 — so the eventual removal should be cheap.

## Two things worth reviewer attention

**Per-backend translation units are preserved.** `fl.fs.sd+.cpp` still carries the SD chain alone, so the linker drops it when `beginSd()` is never called (#2773 item 1.2, ~16 KB on ESP32-S3). Merging the TUs would have been tidier and would have silently cost that.

**The noexcept baseline is path-rewritten, not regenerated.** Moving files invalidated 194 path-keyed entries in `ci/tools/noexcept_baseline.txt`, surfacing 193 "new" violations across 7 files — including files I never touched. The documented fix is `--update-baseline`, but running it here produced **6344 entries against the existing 7093**: this toolchain parses fewer translation units than whoever generated it, so regenerating would have dropped ~750 exemptions and silently weakened the ratchet for everyone. I rewrote the paths surgically instead, preserving all 7093, and retargeted the five `FsImpl` virtuals to `backend.h` where they now live.

## Validation

- `bash lint` clean
- `bash test --cpp` passes
- `bash compile wasm --examples FxSdCard` builds — an example that actually exercises the filesystem

Closes #4008. Refs #4003, #4007, #3986.

## Summary by CodeRabbit

* **New Features**
  * Added a public filesystem API for file access, SD-card support, media loading, and filesystem backends.
  * Added helpers for reading text, JSON, images, audio, video, screen maps, and FLED files.
  * Added a platform-independent file buffer interface.

* **Documentation**
  * Updated filesystem references for the reorganized API.

* **Refactor**
  * Preserved the legacy filesystem include as a compatibility forwarding header.
  * Updated examples, platform integrations, tests, and build configuration for the new filesystem layout.


zackees:
Thanks — one finding was a real defect and is fixed. Two I want to push back on rather than apply silently.

## Fixed: BOM and non-ASCII

You were right, and it is worse than the report suggests: **both predate this branch**. `origin/master`'s `file_system.cpp.hpp` already begins with `EF BB BF`, and already contains `â€”` — a double-encoded em dash, the signature of UTF-8 read as latin-1 somewhere in its history. Moving the file carried them along and surfaced them.

Everything under `fl/fs/`, plus the forwarding header and build TU, is now pure ASCII, and the mojibake is restored to what it was meant to say. Worth removing rather than tolerating: some embedded toolchains treat a BOM as a stray token before the first preprocessor directive.

## Not applying: move inline definitions into `.cpp.hpp`

For `read.h` these are eight one-line forwarding wrappers:

```cpp
inline FramePtr readJpeg(FileSystem &fs, const char *path,
                         const JpegConfig &config = JpegConfig(),
                         string *error = nullptr) {
    return fs.loadJpeg(path, config, error);
}
```

Moving those out of the header prevents inlining and adds a real call per format helper on targets where that matters. The same applies to `backend.h`, where the bodies in question are an empty virtual destructor and a default `ls()` returning `false`.

This is also the codebase's existing habit rather than a departure from it — the `file_system.h` this PR replaces shipped an inline `loadJpegFromSD()` body doing considerably more work than any of these. If FastLED wants a general rule against inline definitions in public headers, that is worth deciding deliberately and applying broadly, not introducing on the file that happens to be under review.

Happy to move them if a maintainer prefers it; flagging the cost rather than quietly paying it.

## Not applying: rename `file_system.h`

> Rename the public header currently named file_system.h to a non-conflicting name, remove its inclusion of fl/fs/fs.h

This reads the file backwards. `fl/system/file_system.h` is now a **compatibility forwarder** — its entire job is to keep its old name and old path so that the 27 existing includers, plus any sketch in the wild, keep compiling after the move:

```cpp
#include "fl/fs/fs.h"   // IWYU pragma: export
#include "fl/fs/read.h" // IWYU pragma: export
```

Renaming it, or removing the include, would turn a non-breaking refactor into a breaking one — which is the specific outcome the header exists to prevent. There is no name conflict: `fl/fs/fs.h` and `fl/system/file_system.h` are distinct paths, and FastLED's `IncludePathsChecker` requires src-root-relative includes, so neither can be reached by a bare `#include "fs.h"`.

## Status

`bash lint` clean, `bash test --cpp` passes, esp32s3 builds, and the bloat gate now measures a 1-byte delta (rebaselined in-PR per its own guidance).

The remaining red checks are SAMD, tracked in #4011: the workflows are new, have never run on master, and fail on `FL_IS_SAMD` detection in GPIO code this PR does not touch. Diff against master over `is_platform.h`, `pin.h`, `pin.cpp.hpp` and `platforms/arm/` shows one line, in a Teensy SdFat header, from the include-path rename.

zackees:
@coderabbitai review

The BOM/non-ASCII finding is fixed. The two structural findings are answered in the comment above — inline forwarding wrappers are deliberate (moving eight one-liners out of the header costs a call each on embedded targets, and matches the codebase's existing habit), and `fl/system/file_system.h` is a compatibility forwarder whose whole purpose is keeping its old name for the 27 existing includers.

zackees:
On the four new `file_handle.h` findings — all are flagged "outside diff range", and that is the right read: this PR **moves** that file (`fl/stl/detail/file_handle.h` -> `fl/fs/file_handle.h`) and adds exactly one line to it.

The first finding is also incorrect:

> `bytes_left()` is not virtual. A derived method is hidden, not dispatched through `filebuf`

It is virtual, on the line the comment points at:

```cpp
// Default implementations (can be overridden)
virtual bool available() const { return is_open() && !is_eof(); }
virtual fl::size_t bytes_left() const;
```

Byte-identical to `origin/master`. My only change to this file is:

```cpp
+#include "fl/stl/shared_ptr.h"  // filebuf_ptr
...
+using filebuf_ptr = fl::shared_ptr<filebuf>;
```

which fixes your earlier — and correct — finding that `backend.h` used `filebuf_ptr` without its include closure declaring it. Worth noting that one turned out to be broader than reported: the alias was declared **four separate times** (`fl/codec/idecoder.h`, `fl/fs/fstream.h`, `fl/fx/video.h`, `fl/video/pixel_stream.h`) and never beside `filebuf` itself.

I am not making behavioral changes to unmodified code in a move PR — that is how a refactor turns into a regression nobody can bisect. If the remaining three findings describe genuine defects, they deserve their own issue against the existing behavior rather than a drive-by fix here.

The `read.h` finding from the previous pass **is** applied: bodies moved to `read.cpp.hpp`, declarations only in the header, `FL_NO_EXCEPT` on all nine. I had pushed back on that one and was wrong — these helpers all end in a file read, so the inlining argument does not hold, and my own design on #4008 had specified `read.cpp.hpp` in the first place.

zackees:
@coderabbitai resolve

All actionable findings from this review are addressed:

- BOM + non-ASCII: fixed (both were inherited from master, not introduced here)
- `filebuf_ptr` undeclared in `backend.h`'s include closure: fixed — declared once beside `filebuf`, where it turned out never to have lived despite four separate redeclarations elsewhere
- `read.h` inline bodies: moved to `read.cpp.hpp` with `FL_NO_EXCEPT`. I initially pushed back and was wrong — these helpers all end in a file read, so the inlining argument does not hold, and my own design on #4008 had specified `read.cpp.hpp`
- `loadJpegFromSD` inline body and test-only hooks: also moved out of the API header (#4003 items)

Two findings I did not apply, with reasons in-thread: renaming `fl/system/file_system.h` (it is the compatibility forwarder — keeping its name is its only purpose), and the three remaining `file_handle.h` items (unmodified code this PR merely relocates; one of them, that `bytes_left()` is not virtual, is factually incorrect — it is declared `virtual` on the cited line).

PR #4068, opened 2026-08-31
fix: resolve eight focused platform and tooling issues

### LINKED ISSUE #3993: fl::chrono::steady_clock is not monotonic — wraps every ~71.6 minutes
## Problem

`steady_clock::now()` is built on the **u32** `fl::micros()`:

```cpp
inline chrono::steady_clock::time_point chrono::steady_clock::now() FL_NO_EXCEPT {
    return time_point(duration(static_cast<fl::i64>(fl::micros())));
}
```

`fl::micros()` is `fl::u32`, so it rolls over every **2^32 µs ≈ 71.6 minutes**. Widening to `i64` after the fact does not help — the value has already wrapped. `steady_clock` therefore jumps backwards roughly once an hour, which violates the one guarantee a steady clock is supposed to make.

`system_clock::now()` has the identical construction and the identical problem.

## Why it matters

Any code that stores an absolute `time_point` and later compares `now() >= deadline` silently stops working across the boundary: a deadline computed just below the rollover is never reached, so the comparison stays false forever.

This was a live bug in the stub watchdog (fixed in #3982 by switching to an unsigned elapsed comparison). The pattern is the hazard, not that one instance — a grep for `steady_clock::now()` will show whether other call sites store absolute deadlines.

Short-lived host test binaries make it improbable in practice, which is presumably why it has not bitten. A long-running host process or a soak test would hit it.

## Two possible fixes

1. **Give the clocks a 64-bit source.** #3982 introduces `fl::platforms::micros64()` for the stub as the single root under `millis()`/`micros()`. Promoting that to the platform contract and having `steady_clock` consume it removes the wrap entirely rather than making it rarer.
2. **Document `steady_clock` as wrapping** and require the elapsed-delta idiom (`now - start >= timeout`, unsigned) at every call site. Cheaper, but leaves a footgun that looks exactly like standard C++ while behaving differently.

Option 1 is the real fix; option 2 is what callers must do until then.

## Related

- #3982 — fixes the stub watchdog instance and adds regression tests covering both directions across the boundary
- Companion issue: `fl::inject_time_provider()` covers `millis()` but not `micros()`, so the two views can already disagree

### LINKED ISSUE #3994: Time mocking is split: inject_time_provider covers millis() but not micros()/steady_clock
## Problem

`fl::inject_time_provider()` / `MockTimeProvider` look like *the* way to control time in tests. They are not — they reach `fl::millis()` only:

```cpp
fl::u32 micros() {
    // Note: micros() does not support time injection
    return fl::platforms::micros();
}
```

Because `fl::chrono::steady_clock::now()` is built on `micros()`, injecting a provider freezes `millis()` while `micros()` and both clocks keep advancing on real time. A test that believes it has stopped time has only stopped a third of it.

## Consequences

**Silently ineffective mocks.** Anything timing off `steady_clock` — including the stub watchdog — ignores an injected provider entirely. The mock appears installed and does nothing. That is worse than having no seam, because the test looks deterministic and is not.

**The two views can disagree.** With a provider installed, `millis()` returns mocked time and `micros()` returns real time, so `micros() / 1000 != millis()`. Any code comparing the two, or converting between them, sees an inconsistent clock.

This is what forced #3982 to add a separate stub-level seam rather than use the documented one.

## Suggested fix

Make one root and derive everything from it. #3982 already does this for the stub: `fl::platforms::micros64()` is the single source, with `millis()` and `micros()` both derived, so overriding one function controls all three views coherently.

The remaining inconsistency is the layer above. Options:

1. **Route `fl::micros()` through the same provider** (scaling the ms-granular provider by 1000). Keeps one public API; costs `micros()` a lock on the `FASTLED_TESTING` path, which `millis()` already pays. Worth adding an atomic fast-path so the no-provider case stays cheap — `millis()` would benefit too.
2. **Deprecate `inject_time_provider` in favour of a root-level clock override**, matching the stub's `setClockFunction`. One seam, no partial coverage.

Either way the current half-covered state should not persist — the failure mode is a mock that silently does nothing.

## Note for whoever picks this up

`fl::millis()` taking a static mutex has a second hazard at process teardown; filed separately.

## Related

- #3982 — stub `micros64()` root and `setClockFunction`
- #3993 — `steady_clock` wraps every ~71.6 min (same `u32 micros()` root)

### LINKED ISSUE #3995: fl::millis() can abort the process when called from a thread during static destruction
## Problem

Under `FASTLED_TESTING`, `fl::millis()` takes a function-local static mutex to consult the injected time provider:

```cpp
fl::u32 millis() {
#ifdef FASTLED_TESTING
    {
        fl::unique_lock<fl::mutex> lock(get_time_mutex());
        ...
    }
#endif
```

`get_time_mutex()`'s static is destroyed at process exit. The destruction order relative to other statics — including ones owning threads that are still running — is unspecified. A background thread that calls `fl::millis()` after that mutex is gone gets:

```
std::__1::system_error: mutex lock failed: Invalid argument
libc++abi: terminating due to uncaught exception
```

Thrown on a thread with no handler, so it aborts the process.

## Observed

Hit in CI on macOS while working #3982. The signature is distinctive — **every test passes, then the binary dies**:

```
[doctest] test cases: 28 | 28 passed | 0 failed
libc++abi: terminating due to uncaught exception of type
  std::__1::system_error: mutex lock failed: Invalid argument
TESTS FAILED: See failure logs above
```

It did not reproduce locally, which is what an unspecified-destruction-order race looks like: platform- and link-order-dependent.

## Scope

The trigger there was the stub watchdog's timer thread polling `fl::millis()`. #3982 works around it by using the lock-free `fl::platforms::millis()` instead — correct for that call site, but a workaround, not a fix. **Any** thread that outlives main and calls `fl::millis()` can hit this, and the failure is an abort with a message that points at a mutex rather than at the real cause.

Test binaries are the exposed surface, since the mutex only exists under `FASTLED_TESTING` — but that is also where an abort is most expensive, because it turns a fully passing run red.

## Possible fixes

1. **Never destroy the mutex** — leak it deliberately (`static fl::mutex* m = new fl::mutex;`). Standard fix for a static needed during teardown; trades a bounded one-off leak for not aborting.
2. **Drop the mutex from the read path** — an atomic provider pointer, as #3994 suggests, removes the lock entirely and is faster on a path `millis()` takes on every call.
3. **Join worker threads before static destruction.** Correct in principle, but every such thread has to opt in, so it does not generalise.

(2) fixes this and #3994 together and makes the hot path cheaper, so it looks like the best value.

## Related

- #3982 — the workaround, plus the CI failure that surfaced this
- #3994 — split time-mocking seam; an atomic provider pointer addresses both

### LINKED ISSUE #3996: audit_usb_registry.py has no VID:PID query mode, but the docs tell you to use it as one
## Problem

`ci/util/audit_usb_registry.py` checks a fixed list — `AUDITED_LITERALS`, the 15 VID:PIDs still hardcoded in `ci/`. It has no way to ask *"is this arbitrary pair published?"*

But the docs added in #3976 instruct exactly that use. `agents/docs/hardware-autoresearch.md` tells you that when `fbuild port scan` cannot name a board, you should confirm against the published catalogue by running the audit — and `agents/docs/usb-vid-pid-registry.md` says to validate an identity before publishing it the same way.

Neither works. An unknown `1234:5678` from a port scan is not in `AUDITED_LITERALS`, so the audit never looks at it and exits **0** — reporting "all good" for a pair that is genuinely absent upstream.

## Why it matters

This defeats the purpose of the instruction it appears in. The step exists to stop people filing false gaps against FastLED/boards for what is really a stale cache or an old fbuild — but as written it will also wave through a real gap, because a clean exit says nothing about the pair you actually care about.

Caught in review on #3976 (CodeRabbit, Major); the docs were merged with the instruction still pointing at a check that cannot answer the question.

## Suggested fix

Add a query mode:

```bash
uv run python ci/util/audit_usb_registry.py --lookup 1234:5678
```

Resolving the pair against the published catalogue and exiting non-zero when absent. The fetch/parse code is already there — `fetch_catalogue()` returns `{vid: (vendor, {pid: product})}` — so this is mostly argument plumbing.

Then update both doc sites to use `--lookup <observed pair>` rather than the bare audit.

Alternative: drop the instruction and point at the portal (<https://fastled.github.io/boards/>), which already does exact VID:PID lookup interactively. Cheaper, but loses the scriptable path.

## Related

- #3976 — added the docs and the audit script

### LINKED ISSUE #3997: Retire the legacy VID/PID tables in ci/util now that all 15 literals resolve upstream
## Context

Two hand-maintained USB tables remain in `ci/`:

- `ENVIRONMENT_TO_VCOM_VID_PIDS` — `ci/util/port_utils.py`
- `BOARD_FINGERPRINTS` — `ci/util/serial_probe.py`

#3976 froze both (DO-NOT-EXTEND comments) rather than deleting them, because at the time one literal — FTDI FT232H `0403:6014` — was not published upstream.

**That gap is now closed** (FastLED/boards#60), and `ci/util/audit_usb_registry.py` reports **15/15 resolve from FastLED/boards**. Nothing in either table is load-bearing any more: every entry is redundant with the registry.

## Work

Delete both tables and route their call sites through fbuild, which already resolves identities from the published catalogue.

Not purely mechanical — the two have different roles and different risk:

**`BOARD_FINGERPRINTS`** is display/diagnostic only; it labels ports that have already been chosen. `fbuild port scan` renders the same `└─ vendor / product` line from the registry, so this one should be a straight deletion.

**`ENVIRONMENT_TO_VCOM_VID_PIDS`** is a *selection* authority — it decides which COM port a run targets. Removing it means the fbuild path must cover every case it covers today, including the LPC845-BRK dual-probe-firmware case that motivated it (#3300): one board, two acceptable VID:PIDs depending on which firmware the onboard probe carries. Verify fbuild handles multi-identity boards before deleting.

Also drop `ENVIRONMENT_TO_VCOM_VID_PID` (singular), the back-compat shim derived from the plural map, once no in-repo callers remain.

## Definition of done

`AUDITED_LITERALS` in `ci/util/audit_usb_registry.py` shrinks to empty and the script can be deleted along with the tables — that list exists to track exactly this retirement.

## Related

- #3836 — RP half of this, delegating port selection to fbuild (#3977)
- #3976 — froze the tables and added the audit
- FastLED/boards#60 — closed the last gap

### LINKED ISSUE #3998: RP CDC recovery can select the wrong board when two identical RP2xxx boards are attached
## Problem

In `_run_discovery_phase` (`ci/autoresearch/phases.py`), when `fbuild deploy` returns no application port, AutoResearch falls back to polling by USB identity:

```python
result = await asyncio.to_thread(
    auto_detect_upload_port,
    expected_environment=rp2xxx_environment,
)
```

`auto_detect_upload_port` filters by **environment only** — which resolves to a VID:PID. Two identical boards (two Pico 2 Ws, say) present the *same* identity, so the filter cannot separate them and the first matching CDC endpoint wins.

If that is the other board, the run proceeds against hardware it did not flash. The failure is quiet: the port opens, RPC responds, and results are attributed to the wrong device.

## Why now

The fallback exists because RP boards enumerate as BOOTSEL mass-storage before deploy and only publish their application CDC endpoint afterwards, so there is no stable pre-deploy port to hold onto.

#3977 widens this path from rpc-smoke/watchdog-soak to **every** RP run mode. That is correct — the same reasoning applies to driver modes — but it does make the ambiguity reachable in many more runs. The ambiguity itself is pre-existing and inherent to identity-only matching, not introduced by that PR.

## Why it is not simply "match harder"

Serial number would disambiguate, but it is not stable across the BOOTSEL → application transition on every RP variant, which is why the code matches on identity in the first place. Any fix has to carry a board handle *through* deployment rather than re-derive it afterwards.

Options, roughly in order of soundness:

1. **Have fbuild return the port it deployed to**, and treat a missing port as a hard error rather than something to recover from. fbuild knows which physical device it flashed; the recovery only exists because that information is sometimes absent. Best fix, needs fbuild work.
2. **Snapshot the CDC endpoints before deploy** and require the recovery to select an endpoint that was *not* present beforehand. Distinguishes the freshly-enumerated board from an already-attached twin, using only host-side information.
3. **Fail closed on ambiguity** — if more than one endpoint matches the identity, error out and require `--upload-port`. Cheapest, and strictly better than silently guessing.

(3) is a small change worth doing regardless of whether (1) or (2) lands, since it converts a silent wrong-board run into an actionable message.

## Reproducing

Attach two identical RP2xxx boards, run any RP AutoResearch mode, and force `fbuild deploy` to return no port. Expect: the run targets whichever endpoint enumerates first.

## Related

- #3977 / #3836 — widened the gate that reaches this path
- Raised in review on #3977 (CodeRabbit, Major)

### LINKED ISSUE #4004: Version 3.10.3 not compiling with SAMD51 AF Metro M4
### Board / MCU

SAMD51

### LED Chipset

WS2812B / NeoPixel

### FastLED Version

3.10.3

### Arduino / PlatformIO version

2.3.10

### Operating System

Windows

### Description

I get the following error when compiling, using the AF SAMD Boards version 1.7.17.  I tried using all of the pin numbers.

```
In file included from C:\Users\perigalacticon\AppData\Local\Temp\.arduinoIDE-unsaved2026722-12608-1ewkzf3.8wmz\sketch_aug22a\sketch_aug22a.ino:2:
c:\Dev\Arduino\libraries\FastLED\src/FastLED.h:906:2: warning: #warning "No pin/port mappings found, pin access will be slightly slower. See fastpin.h for info." [-Wcpp]
  906 | #warning "No pin/port mappings found, pin access will be slightly slower. See fastpin.h for info."
      |  ^~~~~~~
In file included from c:\Dev\Arduino\libraries\FastLED\src/FastLED.h:100,
                 from C:\Users\perigalacticon\AppData\Local\Temp\.arduinoIDE-unsaved2026722-12608-1ewkzf3.8wmz\sketch_aug22a\sketch_aug22a.ino:2:
c:\Dev\Arduino\libraries\FastLED\src/fastspi.h:173:23: note: #pragma message: WARNING: The SPI pins you chose have not been marked as hardware accelerated within the code base. All SPI access will default to bitbanged output. Consult the data sheet for hardware spi pins designed for efficient SPI transfer, typically via DMA / MOSI / SCK / SS pin
  173 | #      pragma message "WARNING: The SPI pins you chose have not been marked as hardware accelerated within the code base. All SPI access will default to bitbanged output. Consult the data sheet for hardware spi pins designed for efficient SPI transfer, typically via DMA / MOSI / SCK / SS pin"
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
In file included from c:\Dev\Arduino\libraries\FastLED\src/FastLED.h:83,
                 from C:\Users\perigalacticon\AppData\Local\Temp\.arduinoIDE-unsaved2026722-12608-1ewkzf3.8wmz\sketch_aug22a\sketch_aug22a.ino:2:
c:\Dev\Arduino\libraries\FastLED\src/fastpin.h: In instantiation of 'class FastPin<6>':
c:\Dev\Arduino\libraries\FastLED\src/platforms/arm/d51/clockless_arm_d51.h:19:49:   required from 'class ClocklessController<6, 30, 75, 45, fl::GRB, 0, false, 280>'
c:\Dev\Arduino\libraries\FastLED\src/chipsets.h:1064:7:   required from 'class WS2812Controller800Khz<6, fl::GRB>'
c:\Dev\Arduino\libraries\FastLED\src/FastLED.h:236:7:   required from 'class WS2812B<6, fl::GRB>'
c:\Dev\Arduino\libraries\FastLED\src/FastLED.h:592:39:   required from 'static CLEDController& CFastLED::addLeds(CRGB*, int, int) [with CHIPSET = WS2812B; unsigned char DATA_PIN = 6; fl::EOrder RGB_ORDER = fl::GRB]'
C:\Users\perigalacticon\AppData\Local\Temp\.arduinoIDE-unsaved2026722-12608-1ewkzf3.8wmz\sketch_aug22a\sketch_aug22a.ino:15:65:   required from here
c:\Dev\Arduino\libraries\FastLED\src/fastpin.h:368:24: error: static assertion failed: This pin has been marked as an invalid pin, common reasons includes it being a ground pin, read only, or too noisy (e.g. hooked up to the uart).
  368 |  static_assert(validpin(), "This pin has been marked as an invalid pin, common reasons includes it being a ground pin, read only, or too noisy (e.g. hooked up to the uart).");
      |                ~~~~~~~~^~
exit status 1

Compilation error: exit status 1
```

### Steps to reproduce

Claude has suggested this is a problem with the AF I2S library.  Using AF SAMD boards version 1.7.17.

### Minimal sketch

```
// Example FastLED setup for Metro M4 (SAMD51)
#include <FastLED.h>

// Define LED strip type and pin
// Pin 6 is a common choice for WS2812b on Metro M4 to avoid conflicts
#define LED_PIN 6
#define NUM_LEDS 60
#define BRIGHTNESS 64
#define LED_TYPE WS2812B
#define COLOR_ORDER GRB

CRGB leds[NUM_LEDS];

void setup() {
  FastLED.addLeds<LED_TYPE, LED_PIN, COLOR_ORDER>(leds, NUM_LEDS).setCorrection(TypicalLEDStrip);
  FastLED.setBrightness(BRIGHTNESS);
}

void loop() {
  // Your animation code here
  leds[0] = CRGB::Red;
  FastLED.show();
}

```

### Serial / debug output

```shell

```

### Checklist

- [x] I have searched existing issues for duplicates
- [x] I have tested with the latest FastLED release or git master
- [x] I can reproduce this with a minimal sketch

### LINKED ISSUE #4020: Apply asset-declared build defines (embedded_fs_defines) to the compile configuration
## Summary

`ci/compiler/asset_scanner.embedded_fs_defines()` decides which preprocessor defines a sketch's assets require, and nothing applies them. An asset declaring `storage=littlefs` is reported (in green) but cannot actually turn a backend on.

## Why it is not simply wired up

Two independent reasons, and the second is the blocking one:

**1. Defines are per-board, this is per-sketch.** Build defines are assembled from `Board.defines` in `ci/boards.py` and rendered into the generated `platformio.ini`. Asset scanning happens later, in `copy_example_source()`, which returns `bool` and has no path back to the build configuration. Threading it through means changing that contract and its three call sites in `ci/compiler/pio.py`.

**2. Applying it today would break the builds it targets.** The only define currently produced is `FL_ESP8266_EMBEDDED_FS`, and the ESP8266 embedded backend does not compile — fbuild caches the ESP8266 core without its `littlefs` submodule, so `<LittleFS.h>` resolves and then fails on its own `#include "../lib/littlefs/lfs.h"` (FastLED/fbuild#1380). Enabling it in response to an asset declaration would turn a working build into a broken one.

So the plumbing is blocked behind fbuild#1380 regardless of effort.

## Worth noting: ESP32 already works

On ESP32 — the platform people actually use LittleFS on — no define is needed. `platforms/embedded_fs.h` selects the backend automatically, so a sketch declaring `storage=littlefs` gets a working filesystem today and the green message is accurate. The gap is ESP8266-only.

## Work

- [ ] Wait for FastLED/fbuild#1380, or find another way to supply ESP8266's LittleFS.
- [ ] Give `copy_example_source()` a return type that can carry build requirements instead of `bool`, and thread it to where `platformio.ini` defines are assembled.
- [ ] Apply `embedded_fs_defines()` output there.
- [ ] Test that a sketch declaring `storage=littlefs` produces the define in the generated ini.

## Not a silent failure

`embedded_fs_defines()` is tested and correct in isolation; it is simply not consumed yet. The reporting function deliberately returns nothing so no caller can appear to apply defines it is dropping — an earlier revision returned them from the reporting path and the caller discarded the value, which read as if the build were acting on it.

Follows #4019.

zackees:
## Summary

This combines eight independently implemented and reviewed fixes:

- extend microsecond time across 32-bit rollover for monotonic chrono clocks
- unify injected millis, micros, steady_clock, and system_clock test time
- keep the injected time-provider state alive through static teardown
- add exact VID:PID lookup support to the USB registry audit tool
- retire the legacy local USB VID/PID and board fingerprint tables
- fail closed when RP deploy does not return an unambiguous application port
- recognize the canonical ARDUINO_METRO_M4 macro for the Metro M4 pin table
- propagate asset-declared embedded filesystem defines into generated builds

## Validation

- bash lint
- bash test --cpp: 369/369 unit and example targets passed
- bash test fastled_core --debug: passed with sanitizers
- targeted chrono, USB registry, AutoResearch, asset define, and SAMD detection tests passed
- broad Python suite: 1068 passed, 41 skipped; two unrelated local infrastructure failures reproduced under verbose rerun (ESP32 QEMU fixture missing setup/loop and native WASM launcher missing the cctype toolchain header)
- independent combined-diff review found no concrete issues and verified all eight issue outcomes

Closes #3993
Closes #3994
Closes #3995
Closes #3996
Closes #3997
Closes #3998
Closes #4004
Closes #4020

## Summary by CodeRabbit

* **New Features**
  * Added exact VID:PID lookup for published USB board identities, reporting whether a match is found.
  * Added 64-bit microsecond timing with improved monotonic behavior across timer rollovers.
  * Added support for Metro M4 board pin mapping.
  * Embedded filesystem build settings are now applied automatically when required by sketches.

* **Bug Fixes**
  * Improved deployment and port handling with clearer failure reporting.

* **Documentation**
  * Updated USB registry and hardware bring-up guidance for the current workflow.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -94,7 +94,7 @@ See `agents/docs/build-system.md` for full command execution rules and forbidden
 ### USB VID/PID identities live in FastLED/boards — ALWAYS
 - **Never introduce a board/device USB VID:PID into this repo as the place it first exists.** [FastLED/boards](https://github.com/FastLED/boards) is the source of truth; it publishes a zstd-compressed protobuf (`usb-vids.proto.zstd`) that fbuild ingests at build time and falls back to from its cache root. FastLED consumes it through `fbuild port scan` / `fbuild deploy`.
 - **Missing identity ⇒ fix the registry, then cascade.** Add it on the FastLED/boards data branch, let the `site.yml` workflow republish, cut an fbuild release, then move the `fbuild==X.Y.Z` pin in `pyproject.toml` and run `uv sync` to pick it up. `uv.lock` is gitignored here, so the pin is the only committed half of the cascade — but you must still relock/sync locally or you keep running the old wheel.
-- **The legacy tables are frozen.** `ENVIRONMENT_TO_VCOM_VID_PIDS` (`ci/util/port_utils.py`) and `BOARD_FINGERPRINTS` (`ci/util/serial_probe.py`) must not gain entries. Test fixtures may use concrete literals; they must never become runtime defaults.
+- **The legacy tables are retired.** Runtime USB identities and environment-aware port selection come from FastLED/boards through fbuild. Test fixtures may use concrete literals; they must never become runtime defaults.
 - Full rule, pipeline diagram, and cascade procedure: `agents/docs/usb-vid-pid-registry.md`. fbuild mirrors it in its own `CLAUDE.md` → "USB VID/PID source of truth" and `docs/usb-vidpid-audit.md`.
 
 ### Deployment (flash / upload) is fbuild's job — ALWAYS
--- a/agents/docs/hardware-autoresearch.md
+++ b/agents/docs/hardware-autoresearch.md
@@ -145,9 +145,11 @@ driver:
 
 1. Run `fbuild port scan` and record the board identity, VID/PID, serial number,
    and port. Do not claim a board is absent without this scan. If the scan cannot
-   name the board, the identity is missing from
-   [FastLED/boards](https://github.com/FastLED/boards) — fix it there and cascade
-   the fbuild bump; never add the literal to `ci/`. See
+   name the board, check the published registry with
+   `uv run python ci/util/audit_usb_registry.py --lookup <VID:PID>`. If that
+   reports the pair missing, fix it in
+   [FastLED/boards](https://github.com/FastLED/boards) and cascade the fbuild
+   bump; never add the literal to `ci/`. See
    [usb-vid-pid-registry.md](usb-vid-pid-registry.md).
 2. Confirm the board has an entry in `ci/boards.py` or can be detected by the
    AutoResearch/fbuild path. If the deploy backend is missing, file the gap in
--- a/agents/docs/usb-vid-pid-registry.md
+++ b/agents/docs/usb-vid-pid-registry.md
@@ -13,8 +13,8 @@ version cascade — not a new literal in `ci/`.
 
 A local table forks on the first board that behaves differently, and the fork
 is invisible until hardware is attached. Three consumers had already drifted
-before this rule was written: `ci/util/port_utils.py`, `ci/util/serial_probe.py`,
-and fbuild's former generated Rust tables all carried overlapping,
+before this rule was written: FastLED's now-retired CI tables and fbuild's
+former generated Rust tables all carried overlapping,
 independently-edited copies of the same identities. FastLED/boards ends that by
 publishing one artifact that every consumer ingests.
 
@@ -101,45 +101,13 @@ uv run fbuild --version
 relock anyway: without it your local environment keeps running the previous
 wheel and any verification you do is against the old registry snapshot.
 
-## Migration status (audited 2026-08-22)
+## Migration status
 
-Every VID:PID literal in this repo's `ci/` tree was checked against the live
-`usb-vids.proto.zstd` (36 vendors / 1056 products). **14 of 15 resolve from the
-registry**:
+The legacy runtime VID:PID tables were retired after all 15 identities resolved
+from FastLED/boards. The catalogue fetch/decode utility remains for the exact
+VID:PID query mode tracked in FastLED #3996.
 
-| VID:PID | Identity | Registry |
-| --- | --- | --- |
-| `2E8A:000A` | rp2040 / rpipico application CDC | resolved |
-| `2E8A:F00A` | rpipicow application CDC | resolved |
-| `2E8A:000F` | rp2350 / rpipico2 application CDC | resolved |
-| `2E8A:F00F` | rp2350w / rpipico2w application CDC | resolved |
-| `2E8A:0003` | RP2 ROM BOOTSEL | resolved |
-| `16C0:0483` | LPCXpresso VCOM (LPC11U35) | resolved |
-| `16C0:0486` | PJRC / Teensy USB-Serial (alt) | resolved |
-| `1FC9:0132` | NXP LPC-Link2 CMSIS-DAP | resolved |
-| `303A:1001` | Espressif native USB CDC | resolved |
-| `10C4:EA60` | Silicon Labs CP2102 | resolved |
-| `1A86:7523` | WCH CH340 | resolved |
-| `1A86:55D4` | WCH CH343 | resolved |
-| `0403:6001` | FTDI FT232R | resolved |
-| `0403:6010` | FTDI FT2232 | resolved |
-| `0403:6014` | FTDI FT232H | **GAP** — FastLED/boards#60 |
-
-The FTDI vendor entry exists but enumerates only `6001` and `6010`; `6014` is
-absent registry-wide. It is the sole blocker to deleting the local tables
-outright rather than freezing them. Do not close that gap by keeping the
-literal — track it in FastLED/boards#60.
-
-Re-run the audit at any time:
-
-```bash
-uv run python ci/util/audit_usb_registry.py
-```
-
-It fetches the live artifact, decodes it, and reports which audited literals
-resolve. A cold or stale cache is indistinguishable from a missing record, so
-always check the published payload rather than a failed port scan. To check an
-observed pair directly, run:
+Check an observed identity against the published catalogue directly:
 
 ```bash
 uv run python ci/util/audit_usb_registry.py --lookup <VID:PID>
@@ -148,20 +116,12 @@ uv run python ci/util/audit_usb_registry.py --lookup <VID:PID>
 The lookup exits 0 and names the vendor/product when published, or exits 1 and
 reports `MISSING` when the exact pair is absent.
 
-Two lists in that script carry the state. `AUDITED_LITERALS` is the checklist —
-it should shrink to empty as literals are retired. `KNOWN_GAPS` holds gaps
-already filed upstream; those print as `GAP*` with their tracking issue and do
-**not** fail the run, so the audit can be wired into CI without going
-permanently red. It exits 1 on a gap that is *not* filed, and also when a
-`KNOWN_GAPS` entry has since been published — that is the prompt to drop the
-allowlist entry and delete the local literal.
-
-## Existing local tables (legacy, do not extend)
+## Runtime ownership
 
 | Location | Status |
 | --- | --- |
-| `ci/util/port_utils.py` → `ENVIRONMENT_TO_VCOM_VID_PIDS` | Legacy. Frozen — every pair is already published in FastLED/boards. Do not add entries; see FastLED#3836. |
-| `ci/util/serial_probe.py` → `BOARD_FINGERPRINTS` | Legacy display/diagnostic hints only. Never a selection authority. Do not add entries. |
+| `fbuild deploy` | Environment-aware port selection, including boards with multiple accepted probe identities. |
+| `fbuild port scan` | FastLED/boards-backed vendor/product diagnostics. |
 | `ci/tests/**` | Test fixtures may use concrete literals to exercise parsing and selection. They must not become runtime defaults. |
 
 Tests are the one sanctioned home for concrete USB literals in this repo —
--- a/agents/docs/usb-vid-pid-registry.md
+++ b/agents/docs/usb-vid-pid-registry.md
@@ -138,7 +138,15 @@ uv run python ci/util/audit_usb_registry.py
 
 It fetches the live artifact, decodes it, and reports which audited literals
 resolve. A cold or stale cache is indistinguishable from a missing record, so
-always check the published payload rather than a failed port scan.
+always check the published payload rather than a failed port scan. To check an
+observed pair directly, run:
+
+```bash
+uv run python ci/util/audit_usb_registry.py --lookup <VID:PID>
+```
+
+The lookup exits 0 and names the vendor/product when published, or exits 1 and
+reports `MISSING` when the exact pair is absent.
 
 Two lists in that script carry the state. `AUDITED_LITERALS` is the checklist —
 it should shrink to empty as literals are retired. `KNOWN_GAPS` holds gaps
--- a/ci/autoresearch/phases.py
+++ b/ci/autoresearch/phases.py
@@ -1324,7 +1324,16 @@ async def _resolve_port_and_environment(ctx: RunContext) -> int | None:
     )
 
     upload_port = args.upload_port
-    if not upload_port and not stock_rp2xxx_transport:
+    # Every board build selects FbuildDriver below. When the environment is
+    # known, passing no port is intentional: fbuild resolves every accepted
+    # identity for that environment from FastLED/boards and reports the
+    # post-deploy application port through FBUILD_DEPLOY_PORT.
+    defer_port_selection_to_fbuild = bool(ctx.final_environment)
+    if (
+        not upload_port
+        and not stock_rp2xxx_transport
+        and not defer_port_selection_to_fbuild
+    ):
         expected_environment = None if is_teensy else ctx.final_environment
         max_wait_s = 60
         poll_interval_s = 1.0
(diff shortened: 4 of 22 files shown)


zackees:
Closing this combined PR at the requester’s direction. The eight fixes are being republished as one isolated PR per issue.

PR #4070, opened 2026-08-31
refactor(ci): retire legacy USB identity tables (#3997)

### LINKED ISSUE #3997: Retire the legacy VID/PID tables in ci/util now that all 15 literals resolve upstream

zackees:
## Summary

- remove the legacy local USB VID/PID and board fingerprint tables
- defer known-board port selection to the registry-backed fbuild path
- retain the shared query decoder and document runtime ownership

## Validation

- 168 focused tests passed
- lint passed

Closes #3997

## Summary by CodeRabbit

- **New Features**
  - USB device identities and upload-port selection now use the FastLED board registry.
  - Deployments automatically report the application port for monitoring.
  - Added exact VID:PID lookup support with clear found or missing results.

- **Bug Fixes**
  - Improved port selection for ESP32, LPC, and other supported boards.
  - Deployments now fail clearly when no application port is available, except for supported RP2xxx workflows.
  - Removed unreliable USB descriptor-based board guessing.

- **Documentation**
  - Updated USB registry and deployment guidance to reflect the new workflow.


zackees:
@coderabbitai review

PR #4097, opened 2026-08-31
fix(autoresearch): stabilize RP2350W peer cycles and OTA staging

### LINKED ISSUE #4059: fix(autoresearch): stabilize RP2350W peer cycles and OTA staging
## Context

The final attached-board gates for the RP2350W DUT and ESP32-C6 peer now get past build, deploy, schema validation, Wi-Fi association, and the fbuild artifact lookup fixed by #4050, but fail in two independent transport paths.

Fresh hardware identity:

- RP2350W DUT: COM18, `2E8A:F00F`, serial `2DCB876B587EA334`
- ESP32-C6 peer: COM9, `303A:1001`, serial `8C:BF:EA:CF:87:B4`

Peer-network RED repro:

```bash
bash autoresearch rp2350w --net-peer --upload-port COM18 --peer-environment esp32c6 --peer-upload-port COM9 --timeout 120s --skip-lint
```

Cycle 1 completes all 12 RP2350W -> C6 requests, all 12 C6 -> RP2350W requests (including the 4 KiB echo), teardown, and RP serial ping. Cycle 2 rejoins the AP and receives `192.168.4.2`, then fails with:

```text
RP2350W startNetServer failed: {'error': 'TCP server failed to listen', 'success': False}
```

A two-cycle probe with no HTTP traffic passes. Running only the 12 C6 -> RP2350W requests makes the next listener start fail. The Arduino-Pico core defaults to five active TCP PCBs. The RP handler currently sends `Connection: close` and immediately calls `WiFiClient::stop()` for every response, making the RP the active closer and leaving lwIP closing-state PCBs behind. The generic `WiFiServer::begin()` failure masks whether `tcp_new`, bind, or listen is the final failing primitive, but the traffic-isolation repro demonstrates the exhaustion lifecycle.

Peer-OTA RED repro:

```bash
bash autoresearch rp2350w --net-peer --ota --upload-port COM18 --peer-environment esp32c6 --peer-upload-port COM9 --timeout 120s --skip-lint
```

The public fbuild API resolves `.build/pio/rp2350w/.fbuild/build/release/firmware.bin` (977,124 bytes), and both deploys plus the 70-method peer schema pass. Artifact staging then times out with `No response with ID 3 within 20.0s`.

Two defects combine here:

1. `RpcClient.send()` already wraps its `args` as one JSON-RPC parameter, but `ota.py` calls `writeOtaArtifact` with `[encoded]`. The wire value is therefore `params:[["<base64>"]]` instead of `params:["<base64>"]`. A small live request returns `-32602 Invalid params` because the firmware byte-vector converter sees an array containing a string.
2. A 512-byte chunk expands to 684 base64 characters plus JSON framing. ESP32-C6 AutoResearch uses HWCDC, whose Arduino core RX queue defaults to 256 bytes. Correctly shaped 16-, 64-, and 128-byte writes pass; a correctly shaped 256-byte write times out, while an immediate serial `ping` still passes. The input line is being dropped/truncated, not crashing the board.

### OTA staging

- Pass the base64 string directly to `RpcClient.send()` so `writeOtaArtifact(fl::vector<fl::u8>)` receives one byte-vector parameter.
- Keep 512-byte artifact chunks for reasonable staging throughput.
- Set the ESP32 AutoResearch serial RX buffer to a documented size large enough for the fully framed request (4 KiB) before `Serial.begin()`. Preserve the existing HWCDC-safe TX configuration.
- Add focused tests that pin the exact RPC argument shape and the pre-begin RX-buffer contract.

### RP2350W peer HTTP lifecycle

- Make every RP response length-delimited and advertise `Connection: close`.
- After sending the complete response, retain the accepted client until the C6 has consumed the declared body and closes its side. Then release the RP wrapper without initiating the close first.
- Keep the existing bounded request deadline as a fallback that actively closes a peer which never disconnects.
- Reset request state exactly once after passive close or timeout, and add focused lifecycle/source-contract coverage.
- Do not solve this by globally increasing lwIP PCB counts or by adding an arbitrary host sleep; those hide the active-close leak and remain sensitive to cycle count.

## Acceptance criteria

- [ ] RED -> GREEN evidence is retained for the malformed OTA argument, the 256-byte HWCDC queue boundary, and the post-inbound-traffic listener failure.
- [ ] Focused Python tests prove OTA sends one base64 string (not a nested list) and that the ESP32 AutoResearch RX queue is enlarged before serial startup.
- [ ] Focused tests cover response-complete/passive-close state, one-time reset, and the bounded fallback close.
- [ ] Existing AutoResearch network/OTA tests and strict lint pass.
- [ ] `rp2350w` and `esp32c6` AutoResearch builds pass through fbuild.
- [ ] The canonical `--net-peer` command completes all ten bidirectional cycles, including exact 4 KiB payload hashes and both serial RPC health checks.
- [ ] The canonical `--net-peer --ota` command stages the resolved fbuild artifact, serves it to the RP2350W, verifies the expected image after reboot, and restores serial/network health.
- [ ] Final evidence records both ports, VID:PIDs/serials, exact commands, decisive output, commit, and PR URL.
- [ ] No raw pyserial, direct flash tool, one-off sketch, host-Wi-Fi switch, hard-coded fbuild artifact path, or unrelated build artifact is introduced.

## Decisions

- Track both failures together because they block the same two-board final acceptance run and share the AutoResearch serial/network fixture; the implementation and HIL validation should land atomically.
- Fix the OTA caller rather than changing `RpcClient`'s global one-parameter convention, which existing firmware bindings rely on.
- Increase the test firmware's RX queue instead of reducing chunks to 128 bytes; the latter would turn a roughly 977 KiB image into more than 7,600 serial RPC round trips.
- Prefer passive TCP close ownership over delays or a larger lwIP pool because the isolated live repro identifies accepted short-lived connections as the trigger.
- Treat #3832 as the broader bring-up parent and #3956 as the earlier OTA server-hardening tracker; this issue owns the newly proven transport defects and their final HIL closeout.

## Related issues

- #3832 ? RP2350W pin-free and ESP32-C6 peer bring-up parent
- #3956 ? earlier RP2350W peer OTA server/watchdog/framing hardening; still awaiting final HIL proof
- #4050 ? fixed fbuild artifact discovery and exposed the subsequent OTA transfer failure

zackees:
## Summary

- harden RP2350W peer-cycle startup and teardown behavior
- stage OTA updates with the required receive buffer and passive-close handling
- type scalar RPC string arguments consistently

## Verification

- 22 focused AutoResearch and OTA tests passed
- bash lint passed

Hardware-in-the-loop peer cycling and OTA flashing still require an attached RP2350W target.

## Shared CI prerequisites

Closes #4059

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved OTA artifact transfers for larger Base64-encoded firmware payloads.
  - Increased ESP32 serial receive capacity to help prevent OTA buffer overruns.
  - Improved RP2350W network responses by allowing data to finish sending before connections close.
  - Standardized RPC requests for more reliable device communication and preserved empty or string parameters correctly.

- **Tests**
  - Added regression coverage for OTA payloads, serial initialization, HTTP response handling, RPC behavior, and platform-specific sketch filtering.


zackees:
@coderabbitai review

zackees:
@coderabbitai review

zackees:
@coderabbitai full review

zackees:
@coderabbitai full review

zackees:
@coderabbitai full review

PR #4098, opened 2026-08-31
fix(samd51): restore hardware SPI compilation

### LINKED ISSUE #4017: SAMD51 hardware SPI drivers do not compile; gated behind FASTLED_SAMD51_HW_SPI pending bring-up
## Summary

The SAMD51 hardware SPI drivers (`SpiHw2`, `SpiHw4`, and their manager) are gated behind `FASTLED_SAMD51_HW_SPI`, off by default, because they do not compile. They need bring-up before they can be enabled.

Gated in #4015:

```
src/platforms/arm/d51/spi_hw_2_samd51.cpp.hpp
src/platforms/arm/d51/spi_hw_4_samd51.cpp.hpp
src/platforms/arm/d51/spi_hw_manager_samd51.cpp.hpp
```

## Why they are broken

These files are guarded by `#if defined(FL_IS_SAMD51)`, which **never evaluated true on any board** until #4011 fixed SAMD detection: `is_samd.h` required `ARDUINO_ARCH_SAMD`, which no SAMD board manifest defines. So the code has never been compiled, and it drifted out of sync with the shared SPI types while nothing was watching.

`SPIQuadSAMD51::acquireDMABuffer` returns `DMABuffer`, but:

```cpp
return SPIError::NOT_INITIALIZED;                     // DMABuffer(SPIError) is explicit
return fl::span<u8>(mDMABuffer.data(), total_size);   // no such conversion exists
```

The first four are mechanical — `DMABuffer(SPIError)` exists and is `explicit`. **The fifth is not.** The driver owns a raw `fl::span<u8>` it allocates with `fl::malloc` and caches across calls, and wants to hand back a view of it. Current `DMABuffer` owns its memory through `shared_ptr<DMABufferInternalData>` and has no non-owning construction path. So the driver's buffer-ownership model predates the API and cannot be reconciled by patching return statements.

## Work

- [ ] Decide the ownership model: either let `DMABuffer(size_t)` own the allocation (losing the driver's buffer-reuse caching), or add a non-owning `DMABuffer` construction path — an API change affecting every SPI platform, so it wants a deliberate decision rather than a local fix.
- [ ] Reconcile `spi_hw_2_samd51.cpp.hpp` and `spi_hw_manager_samd51.cpp.hpp`, gated together and so far unverified — only `spi_hw_4` errors are known, because compilation stopped there. Assume more.
- [ ] Exercise on SAMD51 hardware (Feather M4 / Metro M4). None of this has ever run.
- [ ] Remove the `FASTLED_SAMD51_HW_SPI` gate once it compiles and is verified.

Turning drivers on is safe to defer: `init_channel_driver_samd51.cpp.hpp` reaches them through the shared `SpiHw4::getAll()` / `SpiHw2::getAll()` registries, so with the implementations gated out the registries are simply empty and SAMD51 uses bit-bang SPI — which is what it has always actually done.

## Related

- #4011 — the detection bug that made all of this reachable
- #4016 — the same story for SAMD's Arduino-`<SPI.h>` backend, routed to bit-bang for the same reason
- FastLED/fbuild#1371 — why `<SPI.h>` cannot be supplied under fbuild today

zackees:
## Summary

- restore compilation of the validated SAMD51 hardware SPI path
- correct DMA buffer ownership and size handling
- compile but do not register the unvalidated dual and quad transports

## Verification

- 10 focused SAMD platform tests passed
- bash lint passed

A SAMD51 board build could not be completed locally because the external Darwin ARM toolchain artifact returned 404; physical hardware remains unverified.

## Shared CI prerequisites

Closes #4017

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SAMD51 SPI hardware initialization and DMA buffer handling.
  - Prevented unsupported or unvalidated four-lane QSPI configurations from being registered.
  - Added validation for lane layouts and unsupported three-lane operation.
  - Improved error reporting for SPI initialization, allocation, sizing, and busy states.

- **Tests**
  - Added regression coverage for SAMD51 lane-aware buffer sizing and hardware registration.
  - Added validation that the multiple-ESP32-SPI example is limited to ESP32 platforms.


zackees:
@coderabbitai review

PR #4099, opened 2026-08-31
fix(esp8266): select safe example default pins

### LINKED ISSUE #4018: Set FL_PIN_CLOCKLESS_1 (and the SPI pin defaults) correctly for every platform
## Goal

`FL_PIN_CLOCKLESS_1` should name a pin that genuinely works for a single-wire LED strip on every platform FastLED supports, so `Blink` and the other examples have a one-pin default that compiles **and runs** anywhere without editing the sketch.

## The contract

Documented in `agents/docs/cpp-standards.md` -> "Default Example Pins", and it has two halves:

1. A platform **declares** its value beside its `_FL_DEFPIN` table, `#ifndef`-guarded so a sketch or build flag still wins.
2. A platform that declares nothing **falls back** to `platforms/default_pins.h` (`3` / `1` / `2`). That is a supported outcome, not an omission — most platforms are fine on pin 3.

| Macro | Fallback |
|---|---|
| `FL_PIN_CLOCKLESS_1` | `3` |
| `FL_PIN_SPI_DATA_1` | `1` |
| `FL_PIN_SPI_CLOCK_1` | `2` |

Declared so far:

```
Feather M4 Express / CAN   CLOCKLESS 5, SPI_CLOCK 4   board has no pin 2 or 3
ItsyBitsy M0               SPI_DATA 2,  SPI_CLOCK 4   pin table starts at 2
ESP32 (all variants)       CLOCKLESS 4                GPIO 3 is U0RXD
```

## Why "it compiles" is not the bar

`FastPin<N>::validpin()` catches only pins the platform declares invalid. It cannot catch a pin that exists and is still a bad default. Both failure modes are already confirmed in-tree:

- **Compile-time.** Feather M4 has no pin 2 or 3, so `Blink` and `Apa102` would not build for it at all (#4011).
- **Silent-at-runtime, which is worse.** On ESP32 classic GPIO 3 is `U0RXD`. `Blink` calls `Serial.begin()`, so the fallback had the example driving an output onto its own console RX line. It compiled perfectly. There is nothing to search for when it does not light up.

A good value is: present in that board's pin table; not the UART the examples print on; not a boot-strapping pin (sampled at reset — an attached strip can hold it and stop the board booting); not reserved for flash, PSRAM or USB/JTAG; and actually broken out on the reference board.

### Stage 1 — mechanical sweep (compile)

Run `bash compile <board> --examples Blink` across one board per family and fix anything that fails to build. This catches only the Feather M4 class, but it is cheap and objective.

Status:

| family | board | result |
|---|---|---|
| AVR | `uno` | pass |
| AVR tiny | `attiny85` | pass |
| megaAVR | `nano_every` | pass |
| ESP32 Xtensa | `esp32dev` | in progress |
| ESP32 RISC-V | `esp32c3` | pass |
| ESP8266 | `esp8266` | blocked — unrelated LittleFS build break |
| Teensy 4 | `teensy41` | pass |
| Teensy LC | `teensylc` | pending |
| RP2040 | `rpipico` | pending |
| nRF52 | `nrf52840_dk` | pending |
| STM32 | `bluepill_f103c8` | pending |
| SAM3X | `due` | pending |
| SAMD | all four | pass (#4015) |

### Stage 2 — audit against each platform's own documentation

Per family, decide whether the fallback is *correct* rather than merely compiling. The platform headers already record most of what is needed — `fastpin_esp32.h` documents UART, flash and USB/JTAG pins per variant in its `FASTLED_UNUSABLE_PIN_MASK` comments. Read those rather than inferring from datasheets.

Known open questions:

- **ESP8266** — pin numbering is board-dependent. `_FL_DEFPIN(3,5)` maps FastLED pin 3 to GPIO 5 on some variants and to GPIO 3 (`RXD0`) on others. Needs per-variant treatment, not a blanket value.
- **ESP32-C6 / H2** — GPIO 3 is a strapping pin on H2; GPIO 4 (the value now used) is strapping on C6. Harmless post-boot, but worth confirming against a board.
- **AVR / Teensy / RP2040 / nRF52** — `Serial` is USB on most of these, so pin 3 does not collide. Confirm rather than assume.

### Stage 3 — examples

Sweep `examples/` for remaining hardcoded data pins. `Blink` and `Apa102` are converted; others still carry literals. Document the macros in the examples README so sketch authors reach for them.

### Stage 4 — enforcement

`ci/tests/test_samd_default_pins.py` already resolves all three macros per board branch and asserts the result is in that board's pin table. Generalise it beyond SAMD as platforms declare values, and consider a lint that flags new `#define *_PIN <literal>` in `examples/`.

## Non-goal

Making every platform declare a value. The fallback exists so that platforms where 3 is fine need no entry; a declaration should mean "3 is wrong here", and carry a comment saying why.

Origin: #4011, #4015.

zackees:
## Summary

- choose GPIO4 as the ESP8266 default clockless data pin
- map hardware SPI defaults to GPIO13 MOSI and GPIO14 SCK across ESP8266 numbering variants
- add regression coverage for the generated example defaults

## Verification

- 10 focused ESP8266 default-pin tests passed
- bash lint passed

Physical ESP8266 pin behavior still requires board validation.

## Shared CI prerequisites

Closes #4018

## Summary by CodeRabbit

* **New Features**
  * Added safe default pin assignments for ESP8266 raw, D1, and NodeMCU configurations.
  * Existing custom pin assignments continue to take precedence.

* **Tests**
  * Added validation to ensure ESP8266 defaults avoid reserved and unsafe pins.
  * Added regression coverage for platform filtering in the ESP32 SPI bus example.


zackees:
@coderabbitai review

zackees:
@coderabbitai review

zackees:
@coderabbitai full review

zackees:
@coderabbitai full review

PR #4100, opened 2026-08-31
fix(samd): use native SERCOM for hardware SPI

### LINKED ISSUE #4016: SAMD pulls in Arduino SPI unconditionally; the SAMD SPI stack has never been compiled or exercised
## Summary

SAMD builds now pull in the Arduino `SPI` library unconditionally, via an `#if 0` LDF hint in `platforms/arm/samd/ldf_headers.h`. That is a deliberate stopgap landed with #4011 to get the four SAMD boards compiling. It should not be the end state.

## How we got here

`platforms/arm/sam/fastspi_arm_sam.h` does:

```c
#if defined(FL_IS_SAMD21) || defined(FL_IS_SAMD51)
#include <SPI.h>
#include <wiring_private.h>
```

Neither macro ever evaluated true before #4011, because `is_samd.h` required `ARDUINO_ARCH_SAMD` — which the board manifests do not define. So **the entire SAMD SPI stack has never been compiled**, on any board, ever. Fixing detection switched it on and the build immediately failed:

```
fastspi_arm_sam.h:17:10: fatal error: SPI.h: No such file or directory
```

The LDF hint is the repo's existing mechanism for this, and it keeps the fix inside `src/` so library consumers benefit rather than only this repo's CI matrix (the alternative, `lib_deps` in `ci/boards.py`, would have fixed CI and left every downstream user broken).

## Why this is a stopgap

**1. The dependency is unconditional and the cost is real.** The LDF scanner does not evaluate conditionals, so the hint cannot be narrowed to "only when a SPI chipset is used". `platforms/esp/32/ldf_headers.h` documents exactly this failure mode for larger libraries:

> PlatformIO's LDF scans `#include` directives even inside `#if 0` blocks, which causes the entire library (and its global constructors) to be linked. Global C++ objects like the Arduino `WiFi` instance end up in `.init_array`, creating GC roots that prevent `--gc-sections` from removing ~326 KB.

`SPIClass` is the same shape of problem at much smaller scale: a global instance, hence a GC root, on a part with 32 KB of RAM (SAMD21). ESP32 avoids this entirely by defaulting to the ESP-IDF `spi_master` driver and treating the Arduino `SPIClass` backend as an opt-in escape hatch. SAMD has no such escape hatch today.

**2. Nobody knows whether the SAMD SPI code is correct.** It has no compile history. Compiling is not working — it has never been exercised on hardware in this configuration, and it lives under `platforms/arm/sam/` (the SAM3X/Due directory) despite being SAMD21/SAMD51 code, which is its own source of confusion.

**3. The empty-hint comment was actively misleading.** `platforms/arm/samd/ldf_headers.h` read *"no additional LDF hints needed currently"* — true only while the stack was unreachable. Worth checking whether the same is true of `platforms/arm/stm32/ldf_headers.h`, which is also empty while `fastspi_arm_stm32.h` also includes `<SPI.h>`.

## Proposed work

- [ ] Measure the flash/RAM cost of the unconditional `SPI` link on `samd21` (32 KB RAM) via `bash bloat samd21`, and confirm whether `SPIClass`'s global survives `--gc-sections`.
- [ ] Decide the target shape: a native SAMD SERCOM backend needing no Arduino library, or an `FL_SAMD_SPI_ARDUINO=1` opt-in mirroring the ESP32 arrangement.
- [ ] Exercise the SAMD SPI path on hardware (APA102 or similar) — it has never run.
- [ ] Move the SAMD SPI sources out of `platforms/arm/sam/` into `platforms/arm/samd/`, or state plainly in the SAM README why SAMD code lives there.
- [ ] Audit `platforms/arm/stm32/ldf_headers.h` for the same latent gap.

## Not urgent

SAMD CI is green with the stopgap and the dependency is correct, just broader than it needs to be. This is a size and hygiene issue, not a correctness one.

Follows #4011.

zackees:
## Summary

- replace the Arduino SPI global dependency with the native SAMD SERCOM interface
- use board-default native SPI pins
- restore each controller instance mode and clock divider when selecting the shared SERCOM
- retain the software SPI fallback for unsupported pin selections

## Verification

- 10 focused SAMD SPI tests passed
- bash lint passed

A SAMD board build could not be completed locally because the external Darwin ARM toolchain artifact returned 404.

## Shared CI prerequisites

Closes #4016

## Summary by CodeRabbit

* **New Features**
  * Added native hardware SPI support for SAMD21 and SAMD51 boards.
  * Automatically uses hardware SPI for standard SPI pins and software SPI for alternate configurations.
  * Supports per-controller SPI clock settings and restores them when selecting a device.
  * Enables direct peripheral SPI transfers, including 16-bit output.

* **Bug Fixes**
  * Improved SPI initialization and clock handling on SAMD platforms.
  * Removed reliance on the Arduino SPI library.
  * Prevented unsupported SAMD51 QSPI configurations and protected buffer sizing for active data lanes.


zackees:
@coderabbitai review

Issue #4055, opened 2026-08-25
perf(mp3): add bit-exact SIMD kernels that outperform Helix


zackees:
## Context

Phase 4 / G4 of FastLED/license#14. After scalar fixed-point parity, integer SIMD must make minimp3 strictly faster than Helix while remaining bit-exact and optional.

## Proposal

Add SSE2 and NEON fixed-point kernels for the dominant polyphase, DCT32, and IMDCT costs, then an ESP32-S3 PIE path. Preserve a scalar opt-out and evaluate Cortex-M4/M7 DSP scheduling without sacrificing precision.

## Acceptance criteria

- [ ] Capture RED evidence: performance/dispatch tests show no fixed-point SIMD path and fail the G4 speed targets.
- [ ] Implement SSE2 and NEON int32 paths for polyphase, DCT32, and IMDCT butterflies.
- [ ] Implement compile-time-guarded ESP32-S3 PIE kernels for polyphase and DCT32.
- [ ] Preserve `MINIMP3_NO_SIMD` and prove scalar fallback behavior.
- [ ] Evaluate Cortex-M4/M7 DSP scheduling, including whether `SMLAD` can preserve the Q25 precision contract; document and test the result.
- [ ] Require exact PCM equality between every SIMD path and scalar fixed-point across the full corpus and differential fuzzer.
- [ ] Demonstrate at least 1.5x speedup over Helix-scalar on x86-64 and Apple/ARM CI runners and at least 1.3x on ESP32-S3 hardware, with no target slower than scalar fixed-point.
- [ ] Turn the focused RED signal GREEN, pass broader checks, and merge a PR that closes this issue and links FastLED/license#14.

## Decisions

- Bit-exact equality is blocking because all kernels operate on the same integer representation.
- Host lanes prove the design before the embedded flagship path.
- Cortex-M4/M7 DSP work is evidence-driven and may conclude that scalar is the safe path; the conclusion must be documented.
- This issue depends on G3 and owns G4 only.

## Related issues

- Parent tracker: FastLED/license#14

zackees:
**Acceptance bar set by repo owner (2026-08-31): unit tests and perf tests passing.**

An earlier revision of this comment said the speedup gates were relaxed to informational. That is withdrawn — **perf tests gate too.**

For this issue that means:
[...]

PR #4105, opened 2026-08-31
feat(mp3): fixed-point minimp3 DSP path (#4054)

### LINKED ISSUE #4054: feat(mp3): implement fixed-point minimp3 with Helix parity
## Context

Phase 3 / G3 of FastLED/license#14 and the clean-room conversion designed in FastLED/license#13. The required result is a CC0 fixed-point minimp3 path with measured Helix parity, not a source translation from RPSL Helix.

## Proposal

Convert minimp3 stage-by-stage under the Phase 0 golden harness and Phase 1/2 audit rigs: synthesis, IMDCT/antialias, Huffman/dequant/scalefactors, stereo, and end-to-end guard-bit tracking. Commit generators for all ISO-derived tables and retain `MINIMP3_ONLY_MP3` for v1.

## Acceptance criteria

- [ ] Capture RED evidence: fixed-vs-float golden tests fail because `MINIMP3_FIXED_POINT` is absent.
- [ ] Add fixed-vs-float gates for every corpus file: PSNR at least 90 dB, max difference at most 8 LSB, and at least 55 dB signal-relative quality for -60 dBFS content.
- [ ] Add per-stage dump comparison and ASan/UBSan differential fuzzing with matching accept/reject, frame consumption, and bounded PCM differences.
- [ ] Implement synthesis, IMDCT/antialias, Huffman/dequant/scalefactors, stereo, and guard-bit tracking without copying Helix code, literals, comments, or coefficients.
- [ ] Commit generators for every ISO-formula-derived table and document clean-room provenance.
- [ ] For every converted stage, keep golden tests green and itemize memory/op-ledger deltas.
- [ ] Achieve multiplies/frame no more than 1.1x Helix, working RAM no more than 24 KiB, stack no more than 2 KiB, and target inner-loop counts no more than 1.2x Helix on all four audited ISAs.
- [ ] Validate proxy accuracy with on-target ESP32 and RP2040 cycles/frame spot checks.
- [ ] Turn the focused RED signal GREEN, pass broader checks, and merge a PR that closes this issue and links FastLED/license#14.

## Decisions

- Helix is an ideas-only clean-room reference; all implementation expression remains CC0-compatible.
- Stage boundaries stay measurable throughout conversion; no big-bang replacement.
- This issue depends on the Phase 1 and Phase 2 ledgers and owns G3 only.

## Related issues

- Parent tracker: FastLED/license#14
- Feasibility and clean-room design: FastLED/license#13

zackees:
## Summary

Implements the fixed-point minimp3 DSP path behind `MINIMP3_FIXED_POINT`, turning #4054's RED gate green. The float decoder — still the shipping backend — is untouched.

Against the float decoder over the golden corpus, gates being `>= 90 dB` PSNR and `<= 8 LSB`:

| corpus file | PSNR | max diff | differing |
| --- | ---: | ---: | --- |
| `l3-hecommon.bit` | 123.79 dB | 1 | 31/69,120 |
| `l3-he_free.bit` | 127.35 dB | 1 | 31/156,672 |
| `l3-lame-vbrtag.bit` | 113.23 dB | 1 | 3,724/730,368 |
| `M2L3_bitrate_16_all.bit` | 111.29 dB | 2 | 2,184/274,176 |
| `ILL2_layer1.bit` (Layer I) | 107.31 dB | 1 | 46/2,304 |

## RED → GREEN

Before the implementation, `minimp3 fixed-point variant decodes with integer arithmetic` failed:

```
FAIL: minimp3 fixed-point variant decodes with integer arithmetic
  tests/fl/codec/mp3_fixed_point.hpp:149: fl::Minimp3FixedVariant::dspIsInteger()
```

`-DMINIMP3_FIXED_POINT` was accepted and ignored, so the other two gates passed *vacuously* — both variants were the float decoder and agreed bit for bit. That is exactly why there are two capability probes: `mp3dec_is_fixed_point()` reports what the build was asked for, `mp3dec_dsp_is_integer()` reports what it actually does, and only the second is worth gating on.

Now: 65 test cases, 65 passed, and clean under ASan/UBSan.

## Design

The shape of this conversion came from measurement, not from what fixed-point MP3 decoders usually do. A new test records the dynamic range every stage has to represent: across the conformance corpus *and* real encoded music, every stage peaks near 0.65 (huffman 0.527, imdct 0.557, dct2 0.654). minimp3 normalises internally and the polyphase window carries the scaling back up to int16 only at the very end.

So **one Q26 int32 format serves the whole pipeline**, and block floating point is confined to the two quantities whose range genuinely is unbounded — scalefactor gains (2⁻¹⁸¹ … 2¹⁰) and `x^(4/3)` (up to 165,000) — both carried as mantissa+exponent and collapsed into Q26 at the dequantiser. That is why the Huffman loop, the stereo decisions and the entire bitstream layer stay single-sourced between the two builds: only the four places that turn a decoded magnitude into a sample differ.

Three decisions worth review:

- **Dequantised samples clamp to ±1.0, not to the format's ±32.** The measured legitimate peak is 0.65, so this costs nothing on real audio, and it is what bounds every downstream butterfly — a fuzzed stream can ask for an astronomically large dequantised value, and the DCT-32 stacks three levels of adds on top of a 10.19× secant.
- **One rounding rule**, stated precisely in the header: add half, arithmetic-shift right, which is round-half-toward-+∞ (−1.5 → −1), *not* half-away-from-zero. #4055 must prove SSE2/NEON kernels bit-identical to this scalar path, and a vector unit can reproduce one uniform rule.
- **Saturation is symmetric** (INT32_MIN given up) because `L3_change_sign` negates samples in place.

## Clean room

Tables are generated, never transcribed. `ci/codec_tables/generate_minimp3_fixed_tables.py` derives every constant from its ISO 11172-3 / 13818-3 formula and cross-checks the result against upstream minimp3's own CC0 float literals — that cross-check is the independent witness that a formula was read correctly, and CI runs it with `--check`. `libhelix_mp3` was not a source, which is the whole point of the parent tracker.

## Validation

- **Golden corpus**: table above.
- **Exhaustive helper tests** rather than sampled, for the only genuinely new arithmetic: `x^(4/3)` over all 8,206 reachable quantised values (worst relative error 1.55e-7, against a float reference carrying ~6e-8 of its own) and the scalefactor gain split over every quarter-power the bitstream can produce (2.3e-10).
- **Per-stage comparison**, which localises a regression to one kernel: worst absolute difference 1.5e-8 at huffman/stereo/antialias, 3.7e-8 after the IMDCT, 1.0e-7 after the DCT-32 — Q26 resolution accumulating exactly as expected.
- **Differential fuzzing under ASan/UBSan**, which found and fixed a real UB: `left shift of negative value -32` in the fixed `x^(4/3)`. It produced correct output on every compiler tested, which is why the sanitizer found it and the PSNR gate never would.
- **The generator's own round-trip test** caught a second defect: `normalize()` documented mantissas in [2³⁰, 2³¹) and produced [2²⁹, 2³⁰), so every generated constant carried one fewer significant bit than intended. Fixed; the test asserts the property the generator claims about itself, which is why it found it.
- **Memory audit**, extended with a `minimp3-fixed` row: working RAM 23,308 (budget 24,576), decode stack 1,920 (budget 2,048 — better than the float build's 2,032).

## Honest disclosures

**Converted in one pass, verified per stage.** The criteria ask for stage-by-stage conversion. What landed is a single conversion with per-stage *verification* — the stage-diff table above is the evidence the intent is met, and it is stronger evidence than a sequence of commits would have been, since it measures each kernel independently against the float reference.

**The CPU audit's `minimp3-fixed` row is not in this PR.** Its host baselines are keyed by CI runner CPU model and fail closed on unknown entries by design (#4064), so a new backend row can only be onboarded from the runners' own uploaded artifacts. That work belongs with #4055, where CPU behaviour is the subject. The existing `helix` and `minimp3-float` rows are unaffected: the codegen TU wraps `mp3d_DCT_II` and `mp3d_synth_pair`, neither of which this PR touches.

**Reviewer note on `bash lint`:** the C++ half silently no-ops on a machine without `cc`/`ar`/`ranlib`/`clang-tool-chain-query` on PATH — it reports "All checks passed" while the Rust linter fails to build, so a green exit code alone is not evidence the AST ratchets ran. Once they did run they flagged two real things in the new test scaffolding, both now fixed.

**Local memory-audit note:** `helix/stack-watermark-observed` measures 1,816 here against the ledger's 1,736. This is the local toolchain (clang 21.1.5) versus the ledger's calibrated clang 18, not a regression — verified by swapping master's `minimp3.h` back in and re-measuring, which still gives 1,816. The ledger was deliberately not edited.

Closes #4054
Refs FastLED/license#14

## Summary by CodeRabbit

* **New Features**
  * Added an optional fixed-point integer processing mode for MP3 decoding.
  * Added fixed-point and floating-point output comparison support.
  * Added fixed-point decoder memory and performance auditing.
  * Updated the multiple ESP32 SPI buses example to compile on non-ESP32 platforms.
* **Bug Fixes**
  * Improved audit coverage for codec tables and MP3 variants.
  * Prevented generated audit result files from being tracked.
* **Tests**
  * Added accuracy, dynamic-range, fuzzed-stream, and golden-audio validation.
  * Added checks for generated fixed-point codec table consistency.


zackees:
**Note on the memory audit's red:** `helix/stack-watermark-observed` failed at 3336 against a 1770 limit. It is flaky, not a regression from this PR — filed as #4106.

Evidence: the previous CI run on this same branch, differing only in `codec_cpu_trend.json` (which the memory audit does not read), measured helix at **1736** — the ledger value exactly — and passed. `minimp3-float` measures **2056 in both runs and locally**, to the byte. Only helix moves, and this PR does not touch `libhelix_mp3` or the watermark harness (`git diff master` over both is empty).

The cause is in `finishWatermark`: it takes the first byte anywhere in the painted region that differs from the fill pattern as the decode's high-water mark, so one excursion below the real frame — a lazy PLT resolution on the first call in, a signal frame — moves the figure by kilobytes. Helix is more exposed because its first decode resolves more.

Re-running the job. The deterministic half of the same budget, `stack-callgraph`, is unaffected and passes: helix 528, minimp3-float 2032, minimp3-fixed 1920 against a 2048 limit.

zackees:
Merging with the 8 ESP32-S3 example shards still building (~28 min in). Being explicit about that rather than letting it pass unremarked.

**Complete on this commit: 24 checks, zero failures**, including both mp3 audits (the acceptance bar for this issue), unit tests on Linux/Windows/macOS, `uno`, `bloat regression esp32s3`, and all three QEMU suites.

Why the pending shards do not change the decision:

- `bloat regression esp32s3` and `ESP32-S3 QEMU Test` already passed here. Those compile the FastLED library for S3 with these changes; the shards additionally compile example sketches.
- The only example this PR touches is `MultipleEsp32SpiBuses.ino`, and on ESP32 (S3 included) it now compiles the identical body that the previous `#error` guard admitted. It is a no-op there by construction — the change only affects boards where the sketch was never meant to build.
- If a shard did fail, it would fail on master too, the same way `uno` currently does. This PR moves master toward green rather than away from it, since it fixes that `uno` break.

Happy to revert if a shard turns red.

Issue #4110, opened 2026-09-01
chore(mp3): re-baseline the CPU audit onto the fixed-point decoder


zackees:
## Context

Carve-out from #4056, which deleted the Helix backend and made the fixed-point minimp3 build the production MP3 decoder.

`ci/codec_cpu/audit.py` still audits **`minimp3-float`**, and only that. That is deliberate, not an oversight: `codec_cpu_trend.json`'s host baselines are keyed by the float build's Callgrind symbols and instruction counts, so pointing the driver at the fixed build would invalidate every baseline in the file at the same moment it changed what is being measured. #4055 already demonstrated what a symbol rename alone does to that file. Both the driver (`ci/codec_cpu/driver.cpp`) and the object it links (`ci/codec_memory/minimp3_audit.cpp`) now pin `MINIMP3_FLOAT_POINT` explicitly so the header's new fixed-point default cannot silently swap the measured decoder.

So today the shipping decoder has no CPU trend row.

## What is and is not covered right now

- **Covered:** the fixed build's SIMD kernels are gated on being bit-identical to its own scalar path, and on being no slower than it, by `tests/fl/codec/mp3_fixed_point.hpp` (asserting only under `__OPTIMIZE__`).
- **Covered:** the fixed build's memory and stack footprint, by `codec_memory_ledger.md` (`minimp3-fixed` rows, including the gated compiler-derived `stack-callgraph` figure).
- **Not covered:** per-stage operation counts, cycle/IPC medians, Callgrind function attribution, and cross-target codegen for the fixed build — everything `codec_cpu_trend.json` records.

## Proposal

Add `minimp3-fixed` as a second audited backend rather than replacing `minimp3-float`, so the float reference build keeps its history and the comparison between the two stays available.

## Acceptance criteria

- [ ] `ci/codec_cpu/driver.cpp` can decode with either variant. The two cannot share a translation unit, so this needs a second namespaced instantiation (`MINIMP3_NAMESPACE`) or a second driver binary — `tests/profile/mp3_memory_audit.cpp` has the working pattern for the former.
- [ ] `BACKENDS` becomes `("minimp3-float", "minimp3-fixed")`; `_FUSED_STAGES`, `_operation_stage`, and the codegen kernel needles gain fixed-build entries.
[...]

Issue #4056, opened 2026-08-25
chore(mp3): make minimp3 sole backend and remove libhelix


zackees:
## Context

Phase 5 / G5 of FastLED/license#14. This closes the licensing loop only after fixed-point minimp3 meets all correctness, memory, CPU, and SIMD gates.

## Proposal

Make minimp3 the default, complete the release/deprecation transition, delete the RPSL/RCSL Helix tree, lock all decoder quality/performance evidence into required CI, and upstream the reusable CC0 changes.

## Acceptance criteria

- [ ] Capture RED evidence: the lock-in check fails while Helix is still the default/present and required decoder gates are not all enforced.
- [ ] Make `FASTLED_MP3_BACKEND_MINIMP3` the default and retain Helix only as a warned `FASTLED_MP3_HELIX_LEGACY` opt-in for one minor release.
- [ ] Complete the full FastLED/license#13 acceptance run: golden gates, 72-hour zero-divergence differential fuzzing, A/B listening sanity, and on-target 44.1 kHz stereo 320 kbps budgets (ESP32-C3/RP2040 at most 60% core; ESP32 at most 25%).
- [ ] After the one-minor-release transition, delete `src/third_party/libhelix_mp3/`, its `library.json` assembly filter, and all remaining production/test references.
- [ ] Update third-party manifest, per-file markers, REUSE/SBOM output, release notes, and the FastLED/license#10 libhelix audit/legal-review records.
- [ ] Make golden corpus, fuzz smoke, memory ledger, CPU trend, and SIMD bit-exactness required checks for changes under minimp3 and `src/fl/codec/mp3*`; add CODEOWNERS coverage.
- [ ] Submit the scratch/fixed-point/SIMD work upstream to `lieff/minimp3`, or document the maintained minimal in-tree CC0 diff if upstream is dormant.
- [ ] Prove on `main` that Helix is absent and every lock gate is required; verify one shipped minimp3-default release has no decoder regression reports.
- [ ] Turn the focused RED signal GREEN, pass broader checks, and merge a PR that closes this issue and links FastLED/license#14.

## Decisions

- Helix deletion occurs only after the documented one-minor-release compatibility window and all prior gates.
- Required checks are path-sensitive but fail closed for decoder changes.
- This issue owns G5 and depends on G4; closing it is the prerequisite to closing FastLED/license#14.

## Related issues

- Parent tracker: FastLED/license#14
- License strategy: FastLED/license#10
[...]

PR #4117, opened 2026-09-01
feat(mp3): audit the fixed-point decoder's CPU alongside the float one (#4110)

zackees:
After #4056 the **shipping** decoder had no CPU trend row. The audit measured `minimp3-float` only, because `codec_cpu_trend.json`'s host baselines are keyed by that build's symbols and re-baselining onto fixed point would have invalidated every one of them at the same moment it changed what was measured.

Both are audited now. Float keeps its history rather than being replaced — it is still the reference the fixed-vs-float gates compare against, so the comparison between them stays available.

## Counting integer arithmetic is the interesting part

The float side counts `fmul`, which is unambiguous. `mul` is not: at `-O0` the decoder's Q-format arithmetic and the compiler's address and size arithmetic are the same opcode. I derived the discriminator from the IR instead of guessing:

| step | result |
| --- | --- |
| all `mul` in the fixed audit TU | 200 — mostly index scaling like `mul nsw i32 576, %n` |
| 64-bit with ≥1 operand `sext` from i32 — the shape `(int64_t)a * b` lowers to | 70 |
| minus products consumed by a `getelementptr` or passed to a call | **69** |

That last exclusion earns its place: `L12_apply_scf_384` computes a `memcpy` byte count as `(total_bands - stereo_bands) * 18 * sizeof(...)`, which has the sign-extend shape but is a size, not a sample. It is rejected; nothing else is.

**The tidier-looking rule would have been wrong.** "Both operands sign-extended" reads better and silently drops 16 real multiplies in the hottest kernel — `mp3d_synth_pair`'s polyphase taps are literal constants (29, 213, 459, 2037, 37489, …), so one operand is a constant by construction.

## The result corroborates itself

| stage | minimp3-float | minimp3-fixed |
| --- | ---: | ---: |
| synthesis multiplies | 26,434,368 | **26,434,368** |
| antialias multiplies | 2,103,040 | **2,103,040** |
| imdct multiplies | 6,289,664 | 6,014,848 |
| dequant / huffman | 296,116 / 0 | 9,691 / 146,957 |

Two independently written pipelines doing the same transform arrive at the *same* count. That is what should happen, and it is the check that the integer rule measures what `fmul` measures. Where they differ, they differ for stated reasons: the fixed build fuses dequant into huffman differently, and its IMDCT is not the same kernel.

Both decode 892 frames.

## Also here

- `ci/codec_cpu/minimp3_fixed_driver.cpp` and `minimp3_fixed_codegen.cpp` — the namespaced fixed-point instantiations. The driver's decode body is now a template over a traits struct, so both variants run byte-identical harness code and only the decoder differs.
- The codegen TU needs a local `typedef fl::i64 int64_t;`. `fl/stl/stdint.h` publishes `uint32_t`/`int32_t` globally but for the 64-bit pair emits `u64`/`i64` rather than `uint64_t`/`int64_t`, and this TU's include path is `-I src` only. That is the gap recorded on this issue; fixed here rather than in the shared header, which carries a conflicting-typedef risk its own comments document at length.

## Baselines are not in this commit — read before reviewing red

`codec_cpu_trend.json` needs `minimp3-fixed` entries **measured on the CI host**, which I cannot produce locally. `--write` runs before `--check` precisely so the audit job emits `codec_cpu_current.json` as an artifact even when the check then fails on the missing backend.

So on this first push, the audit job and five trend tests are red. That is the documented bootstrap, not a defect. I will land the measured numbers in a follow-up commit on this branch and the same checks go green.

Refs #4110, #4056

## Summary by CodeRabbit

- **New Features**
  - Added CPU auditing support for the fixed-point minimp3 backend alongside the floating-point backend.
  - CPU audit runs can select either backend and report results independently.
  - Added fixed-point code-generation coverage for key decoding and synthesis operations.

- **Bug Fixes**
  - Improved audit reliability by recording measurements before validation checks and ensuring results are uploaded when checks fail.
  - Improved trend validation for partial host baselines while preserving operation-level checks.

- **Tests**
  - Expanded coverage for fixed-point decoding and integer operation instrumentation.

### DIFF
--- a/.github/workflows/mp3_cpu_audit.yml
+++ b/.github/workflows/mp3_cpu_audit.yml
@@ -86,15 +86,21 @@ jobs:
       - name: Verify the fixed-point SIMD is not slower than scalar
         run: bash test fl_codec --cpp --build-mode release
 
-      - name: Verify the focused CPU gate
-        env:
-          PYTEST_ADDOPTS: -k codec_cpu_audit
-        run: bash test --py --verbose
-
+      # Measure before gating, so a red gate still produces the numbers needed
+      # to diagnose it -- and so that adding a backend can bootstrap its
+      # baselines from a run that necessarily fails the check first. The upload
+      # below is `if: always()`, which was pointless while the step that writes
+      # the file sat behind a gate that could stop it running at all.
       - name: Enforce operation, host, and target-codegen trends
         run: >-
           uv run python ci/codec_cpu/audit.py --check
           --write codec_cpu_current.json
+
+      - name: Verify the focused CPU gate
+        if: always()
+        env:
+          PYTEST_ADDOPTS: -k codec_cpu_audit
+        run: bash test --py --verbose
 
       - name: Upload measured report
         if: always()
--- a/ci/codec_cpu/audit.py
+++ b/ci/codec_cpu/audit.py
@@ -922,10 +922,18 @@ def validate_trend(trend: dict[str, Any]) -> None:
         ):
             raise RuntimeError(f"host baseline key does not match environment: {key}")
         hosts = profile.get("hosts")
-        if not isinstance(hosts, dict) or set(hosts) != set(BACKENDS):
+        # A host baseline may cover a subset of BACKENDS: a backend added after
+        # a host was last measured has no numbers for it until a run lands there
+        # again. Unknown backends are still rejected, and an empty profile is
+        # meaningless.
+        if not isinstance(hosts, dict) or not hosts:
             raise RuntimeError(f"host baseline is missing backends: {key}")
-        for backend in BACKENDS:
-            host = hosts[backend]
+        unknown = set(hosts) - set(BACKENDS)
+        if unknown:
+            raise RuntimeError(
+                f"host baseline names unknown backends: {key}: {sorted(unknown)}"
+            )
+        for backend, host in hosts.items():
             if not isinstance(host, dict):
                 raise RuntimeError(f"invalid host baseline {key}/{backend}")
             validate_host(backend, host)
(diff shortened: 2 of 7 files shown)


zackees:
All three addressed.

**"Keep code-generation checks active for uncovered host backends" — real bug, mine.** When a known host had no baseline for a backend, my `continue` skipped the entire per-backend block, and the target codegen bounds live at the bottom of it. Those are cross-compiled instruction counts — host-independent, read from `baseline_entry["codegen"]` rather than the host profile — so a fixed-point codegen regression on Xtensa or Cortex-M could have passed unnoticed on any runner that hadn't yet measured that backend. Exactly the silent gap the partial-coverage design was supposed to avoid, reintroduced one level down.

The host-dependent half (counters, per-stage timings, Callgrind attribution) now lives in `_check_host_trend` and is called only when a baseline exists. The exact operation ledger and the codegen bounds run unconditionally.

**"Return a dataclass from `_integer_product_lines`" — the guideline exists, so complied.** `agents/docs/python-standards.md:11` asks for a `@dataclass` over tuples for multi-value returns. It now returns `IntegerProducts(lines, values)`; caller and test use the named fields.

(Worth noting against my earlier reply on #4112: I declined a *comprehensions* suggestion there because no such rule exists in this repo, and I rechecked — it still doesn't. The dataclass rule does. Different outcome, same test.)

**"Add the fixed-point trend baseline in this change" — already done, review was against the first push.** The baselines landed in the follow-up commit on this branch, harvested from the audit job's artifact and measured on the CI host (AMD EPYC 9V74, clang 18.1.3). All 31 checks passed on that revision, the audit job included.

The operation ledger came back identical to the numbers measured locally on entirely different hardware — synthesis 26,434,368 multiplies, antialias 2,103,040, imdct 6,014,848 — which is what "host-independent" is supposed to mean and is now checked rather than assumed.

1114 Python tests; the operations audit still reports all eight stages for both backends.

PR #4121, opened 2026-09-01
fix(core): synchronize dithering across controllers

### LINKED ISSUE #1887: Non consistent global brightness across multiple controllers

zackees:
Closes #1887

Synchronize temporal dithering once per logical frame on normal-memory targets. Tiny-memory targets retain the prior per-template phase counter so no static-state or flash growth is introduced.

Validation:
- `bash test pixel_controller --debug`
- `bash compile attiny85 --examples Blink` — 3,454 B flash / 415 B RAM, exactly matching `master`
- `bash lint`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved binary dithering consistency across multiple pixel controllers by synchronizing their temporal dither phase.
  * Advanced dithering once per logical frame, producing more stable and predictable visual output.
  * Preserved lightweight dithering behavior on memory-constrained platforms.
* **Tests**
  * Added coverage verifying shared dither phases and frame-to-frame phase advancement.

### DIFF
--- a/src/FastLED.cpp.hpp
+++ b/src/FastLED.cpp.hpp
@@ -67,12 +67,17 @@ extern "C" void yield(void) { }
 fl::u8 get_brightness();
 
 #if !FL_PLATFORM_HAS_TINY_MEMORY
-fl::u8 fl::detail::ditherFrame() {
-    return fl::Singleton<fl::detail::DitherFrameState>::instance().frame;
+namespace {
+// FL_LINT_ALLOW_GLOBAL(shared dither phase is required once per logical frame)
+fl::detail::DitherFrameState ditherFrameState;
+} // namespace
+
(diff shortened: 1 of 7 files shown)


zackees:
This is going to bloat AVR and it just be selectively enabled for this platform

PR #4122, opened 2026-09-01
perf(mp3): move decoder scratch off stack

### LINKED ISSUE #4116: perf(mp3): move mp3dec_decode_frame_r's inlined locals into the heap scratch arena

zackees:
Closes #4116

Validation: focused regression test and bash lint passed before the split.

## Summary by CodeRabbit

* **New Features**
  * Restored Helix MP3 decoding support.
  * Added fixed-point MP3 decoding by default, with an opt-in floating-point mode.
  * Added runtime reporting for decoder precision and SIMD acceleration.

* **Improvements**
  * Reduced decoder memory pressure by reorganizing temporary audio-processing storage.
  * Improved support for SSE4.1 and ARM NEON acceleration.
  * Updated MP3 memory and stack usage tracking across supported decoding modes.

### DIFF
--- a/ci/codec_memory/audit.py
+++ b/ci/codec_memory/audit.py
@@ -19,31 +19,37 @@ from typeguard import typechecked
 
 ROOT = Path(__file__).resolve().parents[2]
 BUILD = ROOT / ".build" / "codec-memory-audit"
-BACKENDS = ("minimp3", "minimp3_fixed")
+BACKENDS = ("helix", "minimp3", "minimp3_fixed")
 # Audit TU name -> the name the ledger uses for it.
 LEDGER_NAMES = {
+    "helix": "helix",
     "minimp3": "minimp3-float",
     "minimp3_fixed": "minimp3-fixed",
 }
 # Entry point each backend's decode call graph is rooted at.
 CALLGRAPH_ROOTS = {
+    "helix": "MP3Decode",
     "minimp3": "mp3dec_decode_frame_r",
     "minimp3_fixed": "mp3dec_decode_frame_r",
 }
 FRAME_LIMIT = 2048
+STACK_CALLGRAPH_TARGET = 1536
+STACK_FRAME_TARGET = 1024
 RAM_LIMIT = 24 * 1024
-# Retired Helix static-table total (12,744 bytes) + 20%. See check_ledger().
-STATIC_TABLE_LIMIT = 15292
 REGRESSION_FACTOR = 1.02
-# Metrics that are measured and recorded but deliberately not regression-gated.
-# Empty since FastLED#4106: `stack-watermark-observed` used to live here because
-# it was not reproducible across CI runners. The cause turned out to be glibc --
-# the harness forwarded the decoder's block moves to libc, so glibc's `mem*`
-# frames sat inside the measured window and moved the figure by ~900 bytes
-# depending on which implementation the host dispatched to. watermark.cpp now
-# uses its own loops, the measurement is host-independent, and the metric is
-# gated again like every other one.
-INFORMATIONAL_METRICS: set[str] = set()
+# Measured and required to be present, but not regression-gated. The ledger's
+# own prose already says the 2 KiB acceptance gate is the compiler-derived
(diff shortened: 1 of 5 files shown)


zackees:
Superseded by #4134, which landed on master as `7c43414b57` and closed #4116.

Master now does the same thing this PR proposed — hoisting `L12_scale_info` out of `mp3dec_decode_frame_r`'s stack frame — but overlaps it with the mutually exclusive `maindata` window (`scratch->u.l12`) instead of adding a new union member alongside it. That keeps `MINIMP3_SCRATCH_SIZE` unchanged, whereas this branch grew it by 656 B (7936 -> 8592 fixed-point, 7808 -> 8352 float). Master also guards the overlap with a `static_assert(sizeof(L12_scale_info) <= sizeof maindata)` so it stays free.

Closing as superseded; the branch is 38 commits behind and its remaining diff (ledger, FASTLED.patch, PROVENANCE) is stale against master's regenerated versions.

Issue #4106, opened 2026-08-31
test(mp3): helix stack-watermark-observed is flaky and can fail the memory audit spuriously


zackees:
## What happens

`ci/codec_memory/audit.py`'s `stack-watermark-observed` metric for the **helix** backend is not reproducible across CI runs. On identical decoder code (PR #4105; two pushes differing only in `codec_cpu_trend.json`, which the memory audit does not read):

| run | helix | minimp3-float | result |
| --- | ---: | ---: | --- |
| [33451259174](https://github.com/FastLED/FastLED/actions/runs/33451259174) | 1736 | 2056 | pass |
| [33451529474](https://github.com/FastLED/FastLED/actions/runs/33451529474) | **3336** | 2056 | fail |
| same run, re-run of the failed job | **3336** | 2056 | fail |
| local, clang 21.1.5 | 1816 | 2056 | n/a |

The ledger records 1736 with a 2% allowance, so 3336 fails the gate.

Two things are worth noting. `minimp3-float` is identical to the byte in every one of those — only helix moves. And the two CI values are each exact and one of them reproduced on a re-run, so this is not noise around a mean; it is the measurement landing on genuinely different answers depending on something about the environment. Both runs used runner image `ubuntu-24.04` version `20260823.283.1`, so the image is not the variable, and the audit does not log enough about the host to say what is. (The CPU trend already host-keys its baselines by CPU model and fails closed on unknown hosts, from #4064, because the `ubuntu-24.04` pool rotates hardware — the memory ledger has a single global baseline for a measurement with comparable sensitivity.)

## Where the sensitivity comes from

`finishWatermark` in `ci/codec_memory/watermark.cpp` scans the painted region from the bottom and takes the **first** byte that no longer equals `PATTERN` as the deepest point the decode reached:

```c
for (size_t i = 0; i < painted_bytes; ++i) {
    if (context->stack[i] != PATTERN) { first_changed = i; break; }
}
```

So the metric is the deepest byte *anything* disturbed, not the deepest byte the decoder used. A single excursion below the real high-water mark — a lazy PLT resolution on the first call into the decoder, a signal frame, allocator or valgrind bookkeeping — moves the figure by kilobytes. Helix is more exposed than minimp3 because its first decode call resolves considerably more.

I am not certain that is the whole story, and a naive "require N consecutive disturbed bytes" filter would not help if the disturbance is a real frame. The mechanism needs someone to actually instrument it rather than reason about it.

## Why it matters

[...]

PR #4131, opened 2026-09-01
test(mp3): gate the CPU trend on simulated metrics, not timed ones (#4130)

### LINKED ISSUE #4130: test(mp3): CPU trend gates cycle counts at 5%, which is below the runner pool's own variance
## What happens

`ci/codec_cpu/audit.py` gates `counter_median.cycles` at a 5% budget. That is tighter than the variance between machines in the `ubuntu-24.04` pool, so the check fails on PRs that provably cannot have caused it.

Observed on #4128, whose only code change is inside `#if MINIMP3_HAVE_FIXED_POINT` blocks and therefore cannot alter the float build at all:

```
RuntimeError: minimp3-float/counter/cycles regressed: 46509592.0 > 45758997.900000
```

The discriminator is in the same run's output:

| metric | baseline | this run | delta |
| --- | ---: | ---: | ---: |
| callgrind instructions | 261,532,799 | 261,533,566 | **+0.0003%** |
| counter cycles | 43,580,950 (alt host) | 46,509,592 | **+6.7%** |

**Identical work, different cycles.** Callgrind instruction counts are deterministic and confirm the float decoder executed the same instructions to within 767 out of 261.5 million. Only the wall-clock-derived cycle counter moved.

## Why host-keying does not currently save it

`check_trend` already host-keys baselines by `cpu_model | compiler | governor` (#4064) and falls back to `host_baselines` for known hosts. But that key does not separate *machines* — two runners reporting the same CPU model and governor can differ by more than 5% in cycles depending on neighbour contention on the hypervisor. The key identifies a CPU family, not a performance environment.

## This is the second instance of the same shape

FastLED#4106 was exactly this: a host-dependent measurement (`stack-watermark-observed`) gated as if it were deterministic, failing unrelated PRs. That one was fixed by removing the host dependence at the source — the harness stopped calling into glibc, and the number became reproducible.

The equivalent move here is to stop gating on the metric that carries the host dependence, and keep gating the one that does not.

## Options

1. **Demote `cycles` to informational, keep `callgrind/instructions` as the gate.** Instructions are deterministic, already collected, already gated, and already caught real regressions. Cycles then get recorded for trend-watching without failing builds. This mirrors what FastLED#4106 concluded about the watermark, and is the option I would take.
2. **Widen the cycles budget** to something above the observed pool spread. Requires characterising that spread first — a handful of repeat runs on the same commit would establish it — and leaves a weaker gate that still fires occasionally.
3. **Key baselines per machine.** Not really available: runners are ephemeral and expose no stable identity beyond CPU model.

Option 1 keeps a strict gate on the thing that actually measures the code, and stops a noise source from blocking review.

## Acceptance criteria

- [ ] `cycles` (and `ipc`, which is derived from it) no longer fail the build; they are reported and recorded.
- [ ] `callgrind/instructions`, `callgrind/branches`, `callgrind/branch_misses`, the per-function attribution shares, the exact operation ledger and the target codegen bounds all stay gated.
- [ ] The demotion is visible in output, not silent -- print the delta so a real slowdown is still noticeable to a human reading the log.
- [ ] `codec_cpu_trend.json` prose or `ci/codec_cpu/audit.py` records why, so the next person does not "restore" the gate without reading this.

Found while landing #4128 (#4127).

Refs #4106, #4064, #4127

zackees:
The audit gated `counter_median.cycles` at 5%, which is **tighter than the variance between machines in the `ubuntu-24.04` pool**. It failed #4128, whose only code change sits inside `#if MINIMP3_HAVE_FIXED_POINT` and so cannot touch the float build:

```
minimp3-float/counter/cycles regressed: 46509592.0 > 45758997.9
```

The same run's Callgrind figure matched its baseline to **767 instructions out of 261,532,799 — 0.0003%**.

| metric | baseline | that run | delta |
| --- | ---: | ---: | ---: |
| Callgrind instructions | 261,532,799 | 261,533,566 | **+0.0003%** |
| counter cycles | 43,580,950 | 46,509,592 | **+6.7%** |

Same work, different machine. Widening the budget only moves the threshold the noise has to cross.

## The split is by what the number measures

| | |
| --- | --- |
| **gated** | counter instructions, counter branch misses, every Callgrind figure, the exact operation ledger, the cross-compiled codegen bounds |
| **ungated** | cycles, the IPC derived from them, per-stage wall-clock timings |

`counter_median` instructions and branch misses stay gated *despite the name*: `validate_trend` requires them to **equal** the Callgrind figures, so they are simulated rather than sampled.

I had demoted branch misses too, assuming they were a hardware counter. The existing cross-check caught that and the demotion is narrowed to what is genuinely timed — cycles are the only hardware-measured quantity in there.

Ungated metrics print their drift and are flagged past the old budget, so a real slowdown stays visible to a human reading the log.

## Unknown hosts no longer fail closed

That behaviour was correct while cycles were gated — a cycle count from an unrecognised machine compares to nothing. It is wrong now that every gated figure is deterministic, so the primary baseline applies to any host.

Not hypothetical: the rerun of #4128's audit landed on `AMD EPYC 9V45`, a **sixth** CPU model, and failed closed for no reason connected to the change. The pool has presented EPYC 9V74, 9V45 and 7763, and Xeon 8573C and 6973P-C — each new machine blocked unrelated PRs until someone harvested a baseline.

An unknown host now prints `UNKNOWN-HOST:` and gates on the primary baseline. A test pins that it still *gates* — the failure mode to avoid is an unknown host silently checking nothing.

## Validation

1,115 Python tests. Two new tests pin both halves: a 30% cycles swing must not fail the build, and the deterministic gate must still fire on an unknown host.

This unblocks #4128, which is currently red on both host failure modes.

Closes #4130
Refs #4106, #4064, #4128

## Summary by CodeRabbit

* **Bug Fixes**
  * Performance trend checks now gate builds on deterministic instruction metrics.
  * Cycle counts and stage timing changes are reported without blocking builds.
  * IPC regressions are correctly identified when performance decreases.
  * Unknown hardware environments are reported clearly while deterministic metric checks continue to run.
  * Improved reporting distinguishes blocking regressions from non-blocking performance changes.

### DIFF
--- a/ci/codec_cpu/audit.py
+++ b/ci/codec_cpu/audit.py
@@ -968,18 +968,28 @@ def _check_lower_bound(path: str, baseline: float, current: float) -> None:
         raise RuntimeError(f"{path} regressed: {current} < {limit:.6f}")
 
 
-def _report_drift(path: str, baseline: float, current: float) -> None:
+def _report_drift(
+    path: str, baseline: float, current: float, *, higher_is_worse: bool = True
+) -> None:
     """Record a measurement without gating on it.
 
     Loud rather than silent: a real slowdown still has to be visible to someone
     reading the log, it just does not fail the build on a number the runner pool
     moves by more than the budget on its own.
+
+    `higher_is_worse` picks which direction earns the marker. Cycles and stage
+    timings regress upwards; IPC regresses *downwards*, so flagging only
+    positive drift there would stay quiet on exactly the case worth seeing --
+    a 30% cycle increase shows up as roughly -23% IPC.
     """
     if baseline == 0:
         print(f"UNGATED:{path} baseline=0 current={current:.6g}")
         return
     drift = (current - baseline) / baseline * 100.0
-    marker = "  <-- worth a look" if drift > 100.0 * REGRESSION_TOLERANCE else ""
+    regression = drift if higher_is_worse else -drift
+    marker = (
+        "  <-- worth a look" if regression > 100.0 * REGRESSION_TOLERANCE else ""
+    )
     print(
         f"UNGATED:{path} baseline={baseline:.6g} current={current:.6g} "
         f"drift={drift:+.2f}%{marker}"
@@ -1028,6 +1038,7 @@ def _check_host_trend(
         f"{backend}/counter/ipc",
         baseline_host["counter_median"]["ipc"],
         current_host["counter_median"]["ipc"],
+        higher_is_worse=False,
     )
     for stage in STAGES:
         _report_drift(
--- a/ci/codec_cpu/audit.py
+++ b/ci/codec_cpu/audit.py
@@ -968,25 +968,69 @@ def _check_lower_bound(path: str, baseline: float, current: float) -> None:
         raise RuntimeError(f"{path} regressed: {current} < {limit:.6f}")
 
 
+def _report_drift(path: str, baseline: float, current: float) -> None:
+    """Record a measurement without gating on it.
+
+    Loud rather than silent: a real slowdown still has to be visible to someone
+    reading the log, it just does not fail the build on a number the runner pool
+    moves by more than the budget on its own.
+    """
+    if baseline == 0:
+        print(f"UNGATED:{path} baseline=0 current={current:.6g}")
+        return
+    drift = (current - baseline) / baseline * 100.0
+    marker = "  <-- worth a look" if drift > 100.0 * REGRESSION_TOLERANCE else ""
+    print(
+        f"UNGATED:{path} baseline={baseline:.6g} current={current:.6g} "
+        f"drift={drift:+.2f}%{marker}"
+    )
+
+
 def _check_host_trend(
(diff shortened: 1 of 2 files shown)


zackees:
Correct, and it mattered more than "minor" suggests — fixed in the latest commit.

The marker is the *entire* value of an ungated report. Without it the line is a number nobody reads, and I had the direction inverted for one of the three metrics: cycles and stage timings regress upwards, IPC regresses **downwards**. A 30% cycle increase is about −23% IPC, so the report would have printed exactly the case it exists to surface, unmarked.

`_report_drift` now takes the regression direction explicitly; IPC passes `higher_is_worse=False`. The test asserts the IPC line carries both the negative drift and the marker, so the direction cannot quietly flip back:

```python
ipc_line = next(l for l in printed.splitlines()
                if l.startswith("UNGATED:minimp3-float/counter/ipc"))
assert "drift=-23" in ipc_line
assert "worth a look" in ipc_line
```

1,115 Python tests.

Issue #4139, opened 2026-09-01
perf(mp3): add a fast fixed-point mode at Helix parity, keeping today's decoder as HQ


zackees:
Follow-on from #4138, which measured that `minimp3-fixed` — the build that ships on every embedded target — is **~1.59x slower than the Helix decoder it replaced**, while being **20.4 dB more accurate than it**.

We are paying for quality nobody asked for, on the parts least able to afford it.

## Proposal: two modes

| mode | target accuracy | target speed | who |
|---|---|---|---|
| **HQ** (today's decoder, unchanged default where affordable) | 123.24 dB | 1.59x Helix | hosts, ESP32-S3/P4, anything with SIMD or headroom |
| **Fast** (new) | ≥102.87 dB (Helix parity) | ≈1.0x Helix | ESP32-C6/C3/H2 and other non-SIMD 32-bit parts |

The accuracy budget is the whole point of the split:

```
123.24 dB   minimp3-fixed today (HQ)
102.87 dB   Helix — the parity bar
 60.00 dB   ISO limited-accuracy floor
```

There is **20.4 dB of headroom** between what we ship and what Helix managed, and **63 dB** above the ISO floor. Fast mode spends part of the first number and must not touch the second.

## Why this matters on the C6

`bash autoresearch esp32c6 --mp3` on an ESP32-C6 at 160 MHz:

```
decode 155810 us for 235101 us of audio -> 1.51x real time
```

One stereo 44.1 kHz stream consumes two thirds of a core. At Helix parity that becomes roughly 2.4x, which is the difference between "MP3 plus a demanding LED driver" working and not.

## Known-dead path: do not re-run this

A naive 16-bit MAC has already been measured (#4108) and it fails outright:

| variant | PSNR |
|---|---|
| 32-bit MACs (ships today) | 123.24 dB |
| **16x16 MACs simulated** | **44.37 dB** |
| Helix | 102.87 dB |
| ISO floor | 60 dB |

Keeping only the top 16 bits of the wide operand in `mp3d_mulshift` (DCT-32) and `mp3d_synth_pair` (polyphase) costs **~79 dB** — 15 dB *below* the ISO floor, not merely below Helix. The coefficients there are already 16/17-bit (29, 213, 459, 2037, 5153, 6574, 37489, 75038, 64019), so the Q26 samples are the wide side and truncating them is what destroys it.

[...]

Issue #4138, opened 2026-09-01
perf(mp3): fixed-point minimp3 is ~1.6x slower than the Helix it replaced, and nothing gates it


zackees:
## What

FastLED#4056 replaced libhelix_mp3 with minimp3 on licensing grounds (RPSL/RCSL → CC0). The swap was validated on correctness and on memory. **Its speed was never compared against Helix.**

It has been now. Median of 30 runs, pinned to one core, the same five-file corpus / `__builtin_readcyclecounter` / FNV-1a checksum as `ci/codec_cpu/driver.cpp`, all decoding 892 frames:

| backend | audit flags | vs Helix | production flags | vs Helix |
|---|---|---|---|---|
| minimp3-float | 163,146,708 | 0.60x | 39,078,828 | **0.60x** |
| minimp3-fixed | 660,438,396 | 2.41x | 102,928,770 | **1.59x** |
| helix | 274,149,774 | 1.00x | 64,732,068 | 1.00x |

*Audit flags* = `audit.py`'s `-O1 -fno-inline -DMINIMP3_NO_SIMD`. *Production* = `-O2`, inlining and SIMD on. `-fno-inline` inflates the gap from 1.59x to 2.41x but does not create it.

**`minimp3-fixed` is what ships on every embedded target** — the choice is made in `minimp3.h`, not by a build flag. So MP3 decoding on microcontrollers got roughly **59% slower**, while the float build (hosts only) got 40% faster.

On an ESP32-C6 at 160 MHz, via `bash autoresearch esp32c6 --mp3`:

```
decode 155810 us for 235101 us of audio -> 1.51x real time
```

One stereo 44.1 kHz stream eats two thirds of a core. Extrapolating, Helix would have been near 2.4x. That delta is the headroom for LED output, network and UI.

## Two things worse than the number

**1. The shipping build is not gated on host CPU.** `codec_cpu_trend.json`'s `host_baselines` contains only `minimp3-float`, on all three CI hosts. `audit.py` declares `BACKENDS = ("minimp3-float", "minimp3-fixed")` and measures both, so `minimp3-fixed` produces host cycle and Callgrind figures every run with **nothing recorded to compare them against**. It can regress indefinitely without tripping anything. Helix had full baselines; they were deleted with the backend.

[...]

PR #4144, opened 2026-09-03
chore(ci): make the mp3/codec tooling lint-green

zackees:
`bash lint` fails on `master` with 38 raw-subprocess ratchet findings plus 5 `ty` errors. This makes it green.

## Why the ratchet fired

`ci/lint_python/subprocess_capture_checker.py` carries a baseline of known stdlib `subprocess` usage and fails only on *increases*. The `codec_cpu`, `codec_autoresearch` and `codec_conformance` trees post-date that baseline and had **zero** entries in it, so every call site in them counted as new debt.

## The conversion

To `RunningProcess.run`, which the linter documents as a drop-in. One respect where it is **not** drop-in, and which a blind find-and-replace would have broken:

> `subprocess.run` defaults to bytes; `RunningProcess.run` defaults to `text=True`.

Three sites depended on bytes and now say so explicitly:

- `callgrind.py`'s `git show` — its stdout goes straight to `write_bytes()`. Left implicit, the blob would have arrived decoded and silently corrupted the vendored file it writes.
- `device_profile.py`'s `kill` and `test_int_asm_codegen.py`'s compile — output discarded, but keeping the original mode is the honest translation.

Capturing text-mode calls gained `encoding="utf-8", errors="replace"` per SRC004, so a stray non-ASCII byte out of a compiler or `objdump` can no longer crash a measurement run on the locale codec.

## The `ty` errors underneath

The ratchet had been masking these by failing the pipeline first. `TRANSPORT_FAILURE` is a plain `str` sentinel, so `got is TRANSPORT_FAILURE` cannot narrow `str` out of `Measurement | str | None` — every attribute access after the guard read as possibly-`str`. `isinstance(got, Measurement)` expresses the same three outcomes in a form the checker follows.

Confirmed pre-existing rather than introduced here by running `ty` against `origin/master`'s copy of the file: it reports the identical five.

## Baseline

Regenerated to lock in the improvement — 341 findings across 161 entries, down from 173 entries. Counts may shrink freely and only increases fail, so recording the lower numbers is what stops this sliding back.

## Verification

`bash lint` green (`ruff`, `ty`, KBI, DCT, subprocess ratchet, C++, JavaScript).

Not run: `bash test --cpp`. It cannot execute in this environment — `clang-tool-chain-ar --version` fetches a manifest from `raw.githubusercontent.com` on every invocation and fails with `SSL: CERTIFICATE_VERIFY_FAILED`, emitting error logs where the build generator expects a version string, so configuration aborts with `Unknown linker(s)`. That is a sandbox TLS limitation, unrelated to this change; the build files are untouched here and unchanged from master. Worth a look separately if it reproduces off this box.

This branch is Python-only (`ci/`), so no C++ is affected.

https://claude.ai/code/session_01JmbTgspd5g2xg49VaCcxiN

## Summary by CodeRabbit

- **Refactor**
  - Standardized command execution across codec benchmarking, conformance, profiling, reporting, and test tooling.
  - Improved handling of UTF-8 text output and binary data during automated builds and measurements.
  - Preserved expected command-failure reporting and measurement-result handling.

- **Bug Fixes**
  - Prevented interrupt signals from being incorrectly absorbed during processing.
  - Ensured combined command output maintains clear separation between standard output and errors.
  - Updated lint tracking to reflect current subprocess usage.

- **Style**
  - Reformatted informational output and streamlined generated report construction without changing results.


zackees:
Pushed `998d5ce` — the first commit's migration was wrong in three ways and would have shipped two silent data-corruption bugs into the MP3 measurement tooling. Corrections are measured against the installed wheel, not reasoned about.

**1. `capture_output=True` merges stderr into stdout and returns `stderr=None`.** Three crashes as first pushed:

- `callgrind.py` reads valgrind's instruction count from `_IREFS.search(result.stderr)` → `TypeError: expected string or bytes-like object, got 'NoneType'`. That is the entire measurement.
- `report.py` and `device_profile.py` do `proc.stdout + proc.stderr` → `TypeError: can only concatenate str (not "NoneType") to str`.

`stdout=PIPE, stderr=PIPE` does keep them apart, so that is what every capturing site now uses. Note `codec_conformance/run.py:250` already carried a defensive `(result.stdout or "") + (result.stderr or "")` — someone hit this before and worked around it locally rather than writing it down.

**2. It does not round-trip a binary stdout capture, `text=False` or not.**

```
git show HEAD:src/third_party/minimp3/minimp3.h
  subprocess.run       134,038 B
  RunningProcess.run   133,771 B   <- 267 bytes gone
```

The first commit migrated `callgrind.py:materialise_baseline` with a comment asserting `text=False` made it safe. It does not. That blob goes straight to `write_bytes()`, so the baseline decoder source would have been corrupted **silently** and every instruction-count comparison drawn from it wrong — in the tooling #4138 and #4139 are arguing from. `git archive` in `text_size.py` fails louder (71,680 → 71,435 B, `tar` exits 2) but is the same defect.

Both revert to `subprocess.run` with `# noqa: SRC001` and the measurement in the comment — the escape `subprocess_capture_checker.py` documents for "genuinely required". Binary *input* does round-trip, so the `tar -x` call stays migrated.

**3. It strips trailing newlines from each stream**, so `proc.stdout + proc.stderr` welds the last stdout line onto the first stderr line. In `device_profile.py` that corrupts the checksum line and reports a false `CHECKSUMS DIFFER`; in `report.py` it can weld `PSNR=123.24` onto stderr and have the accuracy regex capture a phantom value. Both now join with an explicit `"\n"`.

## Verified by running the tools

| | |
|---|---|
| `callgrind.py --no-helix --top 8` | 247,473,189 Ir, per-function attribution ok |
| `materialise_baseline('HEAD')` | all 3 headers byte-identical to the worktree |
| `text_size.py --functions --inlined --top 15` | byte-identical to pre-migration output |
| `text_size.py --baseline HEAD` | +0 bytes — exercises archive/tar/diff |
| `ci/tests/test_int_asm_codegen.py` | passes |
| `codec_autoresearch/generate.py` | fixture regenerates unchanged |
| `report.py` / `device_profile.py` | `--help`, imports resolve |
| `bash lint` | `python_pipeline ✅`, rc=0 |

## Worth fixing at the source

`subprocess_capture_checker.py`'s SRC001 message calls `RunningProcess.run()` "a drop-in replacement that streams instead of accumulating." Following that literally produces the three failures above. The message should name the `capture_output` stderr merge and the binary-capture limitation, or the next migration repeats this.

`bash lint` is not CI-gated — the workflows run `bash lint --cpp` only — which is how the Python pipeline went red on `master` unnoticed in the first place.

zackees:
**On the red `audit` check: pre-existing, and proven not to be this PR.**

`audit` is red on `master` too (run [33698056201](https://github.com/FastLED/FastLED/actions/runs/33698056201) on `8d2a841c6`) and on every branch cut from it — that is #4138. It is not a required check.

Since this PR edits the very tooling `audit` runs (`ci/codec_cpu/`), "pre-existing" is not good enough on its own, so I compared the measurements this branch produces against master's:

| measurement | master `8d2a841c6` | this branch `998d5ce44` |
|---|---|---|
| `CODEGEN xtensa-esp32 dct32` (minimp3-fixed) | `instructions=1028 inner_loop=1013` | `instructions=1028 inner_loop=1013` |
| `CODEGEN xtensa-esp32 dct32` (minimp3-float) | `instructions=160 inner_loop=144` | `instructions=160 inner_loop=144` |
| `CALLGRIND L3_huffman` (minimp3-fixed) | `26,998,670` | `26,998,670` |

Identical. The capture-plumbing change does not move a single number — which is the point: instruction counts come from valgrind and objdump, not from how their output is piped.

The two runs raise on different assertions (master on `codegen/xtensa-esp32/dct32/instructions regressed: 1028 > 332.85`, this branch on `callgrind-function/.../L3_huffman regressed: 0.0817 > 0.0492`) because `audit.py` raises on the first violation it reaches, not because the sets differ. Both are decoder trend regressions owned by #4138.

Worth noting for #4138: `conformance` **passes** on this branch, and `callgrind.py` now runs end-to-end where the first commit would have crashed it with `TypeError` on `result.stderr`. Whoever picks up #4138 gets working measurement tooling.

Issue #3714, opened 2026-07-22
Unattended RP2040 deploy & recovery: bench host setup instructions (follow-up to #3713)


zackees:
## Context

Follow-up to #3713 (RP2040 IM411 mic board fails USB enumeration). On 2026-07-21 the full software recovery plan from that issue was executed on the Windows bench host and **every software-only path was exhausted**:

- Elevated `pnputil /restart-device` on the failed devnode: reported success, device re-enumerated (fresh `DEVPKEY_Device_LastArrivalDate`), **descriptor handshake failed again** (problem code 43).
- Elevated `Disable-PnpDevice` → `Enable-PnpDevice` (3 s and 10 s power-down windows) + `pnputil /scan-devices`: same result.
- 1200-baud touch / `picotool reboot`: impossible — no live endpoint to talk to.

**Decisive constraint:** a root-hub port restart is a *USB bus reset*, not a *power cycle* — desktop root hubs never cut VBUS. A board whose firmware wedges before/during TinyUSB bring-up (or with a physical D+/D− fault) never re-runs its ROM bootloader, so no host software can ever recover it. Recovery now requires exactly one physical action (BOOTSEL-hold power cycle, cable swap test, or SWD probe attach).

This issue documents **what the bench host needs so this never requires a human again** — i.e. instructions for fully unattended deploy *and* unattended recovery of bricked boards.

## Why unattended deploy fails today

| # | Blocker | Status on bench host |
|---|---------|---------------------|
| 1 | Admin elevation is interactive (UAC consent, `ConsentPromptBehaviorAdmin=5`) | UAC consent was granted quickly during the 2026-07-21 session, but there is no *guaranteed* non-interactive elevation path; `gsudo` is **not** installed (contrary to #3713's env notes) |
| 2 | No true VBUS power-cycle capability | Board is on a root-hub port (`Port_#0014`); VBUS is always on |
| 3 | Software BOOTSEL entry needs a live USB endpoint | 1200-baud touch and `picotool reboot -f -u` both require CDC or picoboot interface — unavailable when descriptor-failed |
| 4 | No SWD debug probe attached | No USB-independent flash/reset path |
| 5 | `picotool` / `openocd` not installed | No tooling for 3/4 even if hardware existed |

### A. Non-interactive elevation (pick one)

**Option 1 — gsudo credential cache (simplest):**
```powershell
winget install gerardog.gsudo
gsudo config CacheDuration -1     # infinite cache for this logon session
[...]

Issue #3713, opened 2026-07-22
RP2040 (IM411 I2S MEMS mic board) fails USB enumeration — recover & flash in no-intervention mode


zackees:
## Summary

An RP2040 board wired to an **InvenSense IM411 I2S MEMS microphone** currently **cannot be programmed**. On the Windows host it enumerates only as **"Unknown USB Device (Device Descriptor Request Failed)"** — it powers up and attempts USB but never completes the descriptor handshake, so there is no BOOTSEL mass‑storage drive and no CDC serial port to target.

This issue tracks recovering the board and flashing it **in no‑intervention mode** (no physical BOOTSEL button press, no cable reseating by a human).

## Probe findings (Windows host, 2026-07-21)

Enumerating only *currently present* devices:

- **Only failing present device:** `USB\VID_0000&PID_0002\6&3AF0F9CE&0&14`
  - Status: **Error — "Device Descriptor Request Failed"**
  - Location: `Port_#0014.Hub_#0001`, parent `USB\ROOT_HUB30\5&4087d53&0&0` (AMD USB 3.10 xHCI)
  - `LastArrivalDate = 2026-07-15 18:19`, still present in Error state
  - `VID_0000&PID_0002` is Windows' generic placeholder for a device whose descriptor read failed — the real VID/PID are unknown at this layer.
- **No `VID_2E8A` (Raspberry Pi) device is present now** — neither BOOTSEL nor app mode.
- **No RPI‑RP2 removable drive** and no removable volume mounted.
- **COM ports present:** COM1 (mobo), COM3/COM4 (Bluetooth), **COM9 = ESP32** (`VID_303A&PID_1001`, unrelated). No RP2040 CDC port.

Historical evidence the *same* board was healthy 5 days ago (ghost/not‑present entries):

- `USB\VID_2E8A&PID_000A\5303284720C4641C` — RP2040 **CDC app mode (COM12)** — last seen 2026‑07‑16 04:34–04:39.
- `USB\VID_2E8A&PID_0003\E0C9125B0D9B` — RP2040 **BOOTSEL (RPI‑RP2 mass storage)** — last seen 2026‑07‑16 09:29–09:30.

So the board was flashed/used successfully on 07‑16, and has since regressed to full enumeration failure.

## Root-cause hypotheses (ranked)

[...]

PR #4167, opened 2026-09-06
feat(wdt): add opt-in boot-loop escape to the bootloader (#3713)

zackees:
Implements the device-side half of #3713's checklist item *"watchdog + boot-loop-to-BOOTSEL escape"*, and gives #3714's unattended-recovery goal a path that does not need a human.

## Problem

A board whose firmware wedges before its USB stack comes up is invisible to the host — no CDC endpoint, so no 1200-baud touch and no `picotool reboot`. #3714 established that **no host-side software can recover it**: a root-hub port reset is a bus reset, not a VBUS power cycle, so the ROM bootloader never re-runs. The only remaining fix is physical.

The device can break that loop itself. Every piece already existed — `consecutiveCrashCount()` (persisted in `watchdog_hw->scratch[3]`, survives soft reset) and `rebootIntoBootloader()` (→ `reset_usb_boot()`). Nothing connected them, so a board in a watchdog boot loop looped forever instead of presenting itself as flashable.

## Change

After N consecutive watchdog resets, reboot into the bootloader, where the board re-enumerates as a flashable target and an unattended host can push working firmware:

```cpp
FastLED.watchdog().setBootloaderEscapeThreshold(3);
FastLED.watchdog().escapeToBootloaderIfLooping();   // early in setup()
```

Three deliberate choices:

- **Opt-in, default 0 (disabled).** Auto-escaping would be a behavior change that could strand boards in BOOTSEL.
- **The escape does not clear the crash counter.** Where bootloader reboot is unsupported it returns `false`; zeroing the count there would destroy the evidence of the very loop being detected. A test pins this.
- **Inline methods over the existing public API**, so this is one file rather than edits to all 13 platform `watchdog_*.impl.hpp` files, and it degrades to a no-op wherever `FL_WATCHDOG_HAS_BOOTLOADER_REBOOT` is absent.

## Verification

- 6 new host tests in `tests/fl/wdt/watchdog.cpp` covering: default-off, threshold round-trip, threshold 0 never escaping regardless of crash count, tripping at the threshold, a clean boot resetting it, counter preservation when unsupported, and independence from the safe-mode knob.
- **RED-verified**: 32 test cases now vs 26 before, and a deliberately broken assertion failed as expected before being restored.
- `bash test --cpp` — **372/372 unit + examples passed**.
- `bash test cled_controller --debug` — passes with ASAN/LSAN/UBSAN.
- `bash lint` — all checks passed.
- `bash compile rp2040 --examples Blink` and `bash compile esp32c6 --examples Blink` — both succeed.
- **Size-neutral when unused**: rp2040 Blink is 315864 bytes flash / 58408 RAM byte-for-byte identical with and without this change (measured by stashing against clean master).

Not exercised on hardware: no RP2350/RP2040 is currently enumerated on this bench (only the ESP32-C6, `303A:1001`). The escape path itself is one `reset_usb_boot()` call that already shipped and is used by `rebootIntoBootloader()`; what is new here is only the policy that decides when to call it.

## Summary by CodeRabbit

- **New Features**
  - Added an optional watchdog boot-loop escape mechanism.
  - When enabled, repeated watchdog resets reaching the configured threshold can automatically reboot the device into the bootloader.
  - The feature is disabled by default and is only available on platforms supporting bootloader reboot.
  - A clean shutdown clears the consecutive reset count, preventing unnecessary bootloader escapes.
  - The bootloader escape threshold can be configured independently from the safe-mode threshold.

### DIFF
--- a/src/fl/wdt/watchdog.cpp.hpp
+++ b/src/fl/wdt/watchdog.cpp.hpp
@@ -5,6 +5,7 @@
 ///   - ResetInfo::describe()        Tier 1 human-readable description writer
 ///   - ResetInfo::subcauseName()    Tier 1 default (per-platform impls may shadow)
 ///   - Watchdog::lastResetInfo()    Tier 1 default (returns {cause, 0, 0})
+///   - Watchdog boot-loop bootloader escape (#3713)
 ///   - ScopedWatchdog ctor/dtor     Tier 1.5 RAII guard
 ///
 /// Per FastLED conventions this file is included from exactly one translation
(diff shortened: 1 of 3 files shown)


zackees:
Fixed in 3bf46c6bef — good catch, this was a real convention violation.

The four definitions now live in `src/fl/wdt/watchdog.cpp.hpp`, which is the right home rather than an arbitrary one: it is documented as *"Platform-agnostic implementations for the unified watchdog API"*, already hosts `Watchdog::lastResetInfo()` (the same shape — a platform-agnostic default over per-platform primitives), and is included from exactly one translation unit via `fl/wdt/_build.cpp.hpp`. Declarations and docs stay in the header, and I updated that file's docblock inventory to list the escape.

Worth noting for reviewers why the definitions were inline originally: the policy is layered over `consecutiveCrashCount()` and `rebootIntoBootloader()`, which are defined **per platform** in 13 separate `watchdog_*.impl.hpp` files. Putting the policy in `watchdog.cpp.hpp` keeps it implemented exactly once instead of duplicated 13 times, so the fix preserves that property while satisfying the header rule.

Re-validated after the move:

- `bash test --cpp` — **372/372 unit + examples passed**
- `bash lint` — all checks passed
- `bash compile rp2040 --examples Blink` — 315864 bytes flash / 58408 RAM, byte-for-byte identical to master, so still size-neutral when unused

(One caveat on that test number, since it bit me twice here: `bash test --cpp` was returning a *fingerprint-cache replay* of a pre-refactor result. The 372/372 above is a genuine post-refactor run, forced via `--clean` plus the runner `chmod` workaround from #4168.)

PR #4169, opened 2026-09-07
fix(rp): derive UART baud ceiling from clk_peri, and say why UART declines (#3899)

zackees:
Found on real RP2350 hardware while running #3899's driver matrix. Host tests and compiles all pass with the old code — only running it surfaces this.

## Symptom

```
TEST UART0 lanes=1 leds=100 FAIL 2016ms class=zero_capture
   RP UART: attempted=True started=False encodedBytes=0 actualBaud=3000000
            lastError='RP UART: achieved baud outside tolerance'
```

`started=False, encodedBytes=0` means nothing was ever transmitted — the zero capture is a *consequence*, not the fault. And "outside tolerance" reads like a tuning problem, which sent me looking in the wrong place first.

## Cause

The achieved baud is exactly **3,000,000 = clk_peri / 16**, the PL011's hardware maximum, which puts `clk_peri` at 48 MHz on this arduino-pico RP2350 build. But the engine hardcoded:

```cpp
constexpr u32 kRpUartMaxBaudRate = 6250000;   // channel_engine_rp_uart.cpp.hpp:13
```

That silently assumes roughly 100 MHz of `clk_peri`. The encoder therefore selected a baud that `uart_init()` clamped, and the achieved-baud check correctly rejected the clamped result. **The tolerance check was doing its job; the ceiling feeding it was wrong.**

The existing comment even noted the ceiling is `clk_peri`-derived, but trusted the achieved-baud check as the backstop. It is a backstop — it just produces a hard failure rather than a slower, representable baud.

## Change

Ask the peripheral for its real ceiling instead of assuming one:

- `IRpUartPeripheral::maxBaudRate()` — virtual with a `kRpUartBaudCeiling` default, so existing mocks and the host test are unaffected.
- `RpUartPeripheral` overrides it with `clock_get_hz(clk_peri) / 16`, clamped to the generic ceiling. The Pico SDK include sits inside the existing `FL_IS_RP2040 || FL_IS_RP2350` guard — the engine is compiled in the host unity build and must stay free of SDK headers.

`canHandle()` now also records *why* it declined. A bare `false` left the diagnostic empty, making an unreachable-baud board look identical to a wiring fault:

```
RP UART: attempted=False started=False encodedBytes=0 actualBaud=0
lastError="RP UART: chipset timing needs a baud above this board's clk_peri/16 ceiling"
```

`mLastError` becomes `mutable` because `canHandle()` is const.

## What this does *not* do

It does not make UART clockless work on this board, and nothing can. WS2812 wave10 encoding needs more than 3 Mbaud, so at `clk_peri` = 48 MHz the timing is not representable at all. The driver now declines honestly instead of failing opaquely mid-start.

**This means #3899's premise for UART is wrong** — it expects "UART clockless output on any hardware UART route compatible with the discovered bridge (GPIO0 -> GPIO1 is UART0 TX/RX)". The bridge is fine and confirmed; the clocking is the blocker. UART should be recorded as *not reachable on this fixture*, in the same category the issue already reserves for SPI.

## Verification

- `bash test --cpp` — 288/288 unit, 84/84 examples passed
- `bash lint` — all checks passed
- Built and run on real RP2350 hardware for both the before and after diagnostics
- Same board, same session: **PIO0, PIO1, and PIO2 all pass 4/4**, and RPC smoke passes — so this is specific to the UART path, not general breakage

Bridge confirmed live via `findConnectedPins` as **TX GPIO0 → RX GPIO1**, which #3899 asked be proven rather than assumed.

## Summary by CodeRabbit

* **Bug Fixes**
  * UART communication now uses the highest baud rate supported by the specific hardware.
  * Baud-rate validation and transmission timing adapt to each peripheral’s clock configuration.
  * Channels are rejected when the UART backend is unavailable or its clock is not running.
  * Clearer diagnostics are provided when the requested baud rate cannot be reached.
  * UART transmission now avoids unsupported timing configurations for improved reliability.

### DIFF
--- a/src/platforms/arm/rp/rpcommon/channel_engine_rp_uart.cpp.hpp
+++ b/src/platforms/arm/rp/rpcommon/channel_engine_rp_uart.cpp.hpp
@@ -6,12 +6,6 @@
 
 namespace fl {
 
-namespace {
-// RP2040 PL011 is clocked from clk_peri.  Keep a conservative 6.25 MHz
-// ceiling here rather than inheriting ESP32's 5 MHz policy; the achieved-baud
-// check below is the final hardware-specific timing admission.
-constexpr u32 kRpUartMaxBaudRate = 6250000;
-}
 
 ChannelEngineRpUart::ChannelEngineRpUart(
     fl::shared_ptr<IRpUartPeripheral> peripheral, u8 uart_index) FL_NO_EXCEPT
@@ -42,8 +36,22 @@ bool ChannelEngineRpUart::isValidTxPin(int pin) const FL_NO_EXCEPT {
 }
 
 bool ChannelEngineRpUart::canHandle(const ChannelDataPtr& data) const FL_NO_EXCEPT {
-    return data && data->isClockless() && isValidTxPin(data->getPin()) &&
-           canRepresentTimingForMaxBaud(data->getTiming(), kRpUartMaxBaudRate);
+    if (!data || !data->isClockless() || !isValidTxPin(data->getPin())) {
+        return false;
+    }
+    // Ask the backend what it can actually reach rather than assuming a fixed
+    // ceiling: the PL011 is bounded by clk_peri/16, so a slow clk_peri can put
+    // the required baud out of range entirely. Record why we declined --
+    // returning a bare false here left the diagnostic empty and made an
+    // unreachable-baud board look identical to a wiring fault. See #3899.
+    const u32 max_baud = mPeripheral ? mPeripheral->maxBaudRate()
+                                     : kRpUartBaudCeiling;
+    if (!canRepresentTimingForMaxBaud(data->getTiming(), max_baud)) {
+        mLastError = "RP UART: chipset timing needs a baud above this board's "
+                     "clk_peri/16 ceiling";
+        return false;
+    }
+    return true;
 }
 
 void ChannelEngineRpUart::enqueue(ChannelDataPtr channelData) FL_NO_EXCEPT {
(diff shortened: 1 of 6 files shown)


zackees:
All three findings applied in 3614eea9cf — thanks, one of them was a real bug, not a style issue.

**1. Zero-clock rejection — this was a genuine defect.** My fallback was backwards: when `clock_get_hz(clk_peri)` returned 0, `maxBaudRate()` reported the optimistic 6.25 MHz ceiling, so `canHandle()` would accept a channel that `uart_init()` then rejects. Your pico-sdk citation is correct — `uart_init()` returns 0 when the UART clock isn't running. An unusable clock has to refuse, not report the maximum. Now returns 0, and the engine declines with `"RP UART: UART clock unavailable"`.

**2. Backend-neutral message.** Correct — that path is shared with mocks and any backend whose ceiling isn't clock-derived, so naming `clk_peri/16` could assert a cause that doesn't apply. The message no longer does, with a comment recording why.

**3. Header definition.** Rather than relocate the default, I made `maxBaudRate()` **pure virtual**. That removes the definition from `src/**/*.h` *and* eliminates the silent fallback underlying finding 2 — every backend must now state its real ceiling, which is precisely the property whose absence caused finding 1. The test mock gains a one-line override.

Also added three host regression tests, since nothing guarded any of this:

- refusal on a zero-clock backend
- refusal when the timing needs more baud than the backend reports
- a control case flipping only the reported ceiling on the same pin and channel, so the first two can't pass trivially

**Verification:**

- `bash test --cpp` — 372/372 unit + examples passed
- `bash lint` — clean
- On RP2350 hardware: UART0 now reports `attempted=False ... lastError='RP UART: chipset timing needs a baud above the UART backend maximum'`
- Same board, same session: **PIO0 still passes 4/4** — confirming the refusal path change didn't regress the drivers that work

zackees:
Fixed in cc471691d3 — you're right, and this was a real gap in my tests rather than a style preference.

`FL_CHECK_FALSE(lastError().empty())` passes for any non-empty message, so if the zero-clock and excessive-baud diagnostics were ever swapped, both cases would still have gone green — defeating the whole reason for having two separate tests. They asserted "an error happened" when they needed to assert "*this* error happened."

Both now compare the exact string, and I checked the assertions actually bite by swapping the two messages in `channel_engine_rp_uart.cpp.hpp`:

```
FAIL: RP UART declines when the backend reports no usable UART clock
FAIL: RP UART declines when the timing needs more baud than the backend has
[doctest] test cases: 10 | 8 passed | 2 failed
```

Both fail on the swap, both pass on restore. Under the previous assertion that swap was silently undetected.

`bash test --cpp` 372/372 unit + examples, `bash lint` clean.

PR #4175, opened 2026-09-07
chore(ci): drop the dead codec_cpu_ledger.md path filter (#4155)

zackees:
First housekeeping item from #4155.

## Change

`codec_cpu_ledger.md` has never existed in this repo — `git log --all -- codec_cpu_ledger.md` returns nothing — yet `.github/workflows/mp3_cpu_audit.yml` watched it in **both** its `push` and `pull_request` path filters. This removes those two entries.

No trigger behavior changes: the real baseline, `codec_cpu_trend.json`, is already listed in both filters and stays. A dead entry isn't harmless clutter either — the next person reading the filter reasonably assumes the file exists and goes looking for it.

## Verification

- Workflow still parses
- Both filter lists go 12 → 11 entries
- `codec_cpu_trend.json` still watched in both
- No `codec_cpu_ledger` reference remains anywhere in the file
- `bash lint` — clean (including the meson/YAML lane)

## Scope

Only the first item in #4155. The other leftovers there — huffman regression, upstreaming — are untouched and stay open.

## Summary by CodeRabbit

* **Chores**
  * Updated MP3 CPU audit workflow triggers so changes to the CPU trend data continue to run audits, while changes to the CPU ledger no longer trigger them.


zackees:
The `audit` failure here is not from this change. This PR deletes two path-filter lines; the red step is `bash test fl_codec --cpp --build-mode release`, failing on the SIMD-vs-scalar perf assertion at `speedup=0.945x` against a `0.95` threshold.

That gate is flakier than the effect it measures — I filed #4183 with the numbers. Short version: the true speedup on x86 is ~1.02x (not the ~1.10x the in-source comment claims), so the threshold sits inside the estimator's noise band; the same assertion has taken `master` red twice in the last week (0.904, 0.924).

Re-running the failed job.

zackees:
Correction to my earlier comment: I re-ran the job and it failed again at 0.920x. Four visible measurements now, all clustered 0.904–0.945, and the run with the *fastest* scalar time gave the *worst* ratio — so this is not the flake I called it. Details and the corrected reading are on #4183.

Practically: re-running will not clear this, and #4175 stays blocked on that question rather than on anything in its own diff.

zackees:
#4192 should unblock this. The `audit` failure here has never been about this diff — it's the SIMD-vs-scalar gate tracked in #4183 — and the reason that issue has stalled is that a passing run of the gate records nothing, so nine green runs can't be told apart from nine skipped measurements.

#4192 adds `--verbose` to that step so the ratio is always printed. Once it lands, the next audit run on this PR will either show a healthy ratio (in which case this was a flake after all and this merges) or show the gate skipping (in which case #4183 has its answer and this still merges).

zackees:
Update: #4192's first run answers the question that was blocking this. A passing run of the gate reports `speedup=1.080x` — the gate is real and healthy on that runner — while the failures here reported 0.904–0.945. A 19% spread across GitHub's runner fleet, so the kernel is genuinely faster on some runners and slower on others.

That makes this PR's red `audit` runner-dependent, not caused by anything in its two-line diff. #4183 now has the evidence to fix the gate properly; once #4192 lands and that is settled, this merges.

zackees:
Closing as already applied — this change is on `master` and has been since #4176.

That PR ("fix(autoresearch): make BLE bring-up name the board that actually answered") carried the identical two-line deletion alongside its own changes:

```
 .github/workflows/mp3_cpu_audit.yml | 2 --
-      - 'codec_cpu_ledger.md'
-      - 'codec_cpu_ledger.md'
```

Rebasing this branch onto current `master` confirms it: git drops the commit with *"patch contents already upstream"*. There is nothing left in the diff.

Worth noting what this does **not** invalidate. The red `audit` check here was never about this diff, and chasing it turned up something real that stands on its own: the SIMD-vs-scalar gate measures **0.904–1.080** depending on which runner GitHub hands out, so it takes `master` red at random rather than guarding anything reliably. That is tracked in #4183, with #4192 now landing the diagnostics — the ratio and the runner CPU are recorded on passing runs too, so the two populations can be named instead of inferred.

So the filter cleanup is done, and the investigation it accidentally triggered continues under its own issue.

Issue #4173, opened 2026-09-07
RP2350W hangs in wifiConnect on the 7th peer cycle, tripping the watchdog


zackees:
## Summary

`bash autoresearch rp2350w --net-peer` completes six full peer cycles and then the RP2350W **hangs inside `wifiConnect` on the seventh**, long enough to trip the watchdog. Reproducible: two consecutive runs failed at cycle 7/10.

The good news first, because it is the larger part of the result: **the peer network works.** Each cycle runs `wifiConnect` → join the C6 SoftAP → `startNetServer` → HTTP RP2350W→ESP32-C6 → HTTP ESP32-C6→RP2350W → `stopNet` → serial ping, and cycles 1-6 complete all of it. Bidirectional HTTP between the two MCUs is proven on hardware.

## Evidence

```
--- Peer network cycle 1/10 ---   RP2350W serial RPC ping: PASS
...
--- Peer network cycle 6/10 ---   RP2350W serial RPC ping: PASS
--- Peer network cycle 7/10 ---
  Network peer autoresearch failed: No response with ID 112 within 4.79s
```

| Run | Failure |
| --- | --- |
| 1 | `No response with ID 115 within 2.54s` |
| 2 | `No response with ID 112 within 4.79s` |

Same cycle both times, different request IDs and timeouts — a deterministic failure point rather than flakiness.

**It is a watchdog reset, not a dropped reply.** Querying the board immediately after:

```
ping       -> {"message":"pong","success":true,"uptimeMs":43785,
                "lastResetCause":"WATCHDOG","lastResetWasWatchdog":true}
wifiStatus -> {"ip":"","apIp":"","status":"IDLE","success":true,"connected":false}
```

`lastResetWasWatchdog: true` with a ~44 s uptime means the device rebooted mid-cycle. The RPC did not time out because a reply was lost; it timed out because the device was resetting. That distinction matters — this is not the "request/reply loops drop replies" class documented in `ci/autoresearch/rpc_bench.py`, and it is not a `--net-peer` harness fault.

The failing call is the first RPC of the cycle: the loop prints `cycle 7/10` and dies before `wifiConnect` returns (`ci/autoresearch/net.py:536`).

## Interpretation

[...]

PR #4182, opened 2026-09-07
fix(autoresearch): read RP HTTP responses to completion, not to a deadline (#4173)

zackees:
Addresses the root cause found in #4173. **Please do not merge on the timing improvement alone** — see Validation gap below.

## Root cause

`runRpHttpRequestTest` drained each response like this:

```cpp
const uint32_t deadline_ms = millis() + 2000;
while (millis() - deadline_ms < 0 && (client.connected() || client.available()))
```

It exits on the 2 s deadline or on the peer closing. The request *does* send `Connection: close`, but `client.connected()` does not go false promptly enough to end the read — so the loop waits out the remaining deadline on **every request, after the body has already arrived**.

`runNetClientTest` issues 12 requests. That is ~2.2 s each and the **26.57 s** measured on hardware, almost all of it idle. The caller allows `max_wait=30.0` and the peer cycle makes two such calls, so a healthy first cycle has ~3.4 s of headroom.

That reframes #4173: not a resource exhausting at exactly cycle 7, but an operation **marginal from cycle 1** that eventually loses a race it was always close to.

## Change

Stop on completeness rather than silence:

- parse headers, capture `Content-Length`, stop at the blank line
- read exactly that many body bytes, then break **without consulting connection state**
- keep the 2 s deadline as the ceiling for a server that never replies
- keep draining past the 256-byte buffer so the socket is not abandoned mid-response

It also separates headers from body. `response[256]` previously held **both**, so a response with long headers could push the asserted fragment out of the window entirely — a failure mode independent of timing, and a plausible source of flaky content checks.

## Measurements (RP2350W ↔ ESP32-C6 SoftAP)

| | per `runNetClientTest` | cycles fully passing |
| --- | --- | --- |
| before | 26.57 s | 1/1 measured |
| this change | 8.1 – 20.4 s | 6/9 |
| focused re-run | — | 3/3 at 12/12 sub-tests |

## Validation gap — the reason this is not "ready"

**I have not established the baseline pass rate.** My only pre-change data point is a single `ok=True` cycle, which says nothing about how the unmodified build scores over nine. So I cannot currently tell whether 6/9 is:

- equal correctness plus a ~2.3× speedup (worth landing), or
- flakiness I introduced (should be rejected)

The comparison run was cut short by host memory pressure mid-deploy, and the board's flashed build became indeterminate, so I stopped rather than report a number I could not stand behind. Note the harness itself fails at cycle 7, so the baseline may well be flaky too.

**What would settle it:** flash unmodified `master`, run the same 9-cycle probe, compare pass rates. If baseline is also ~6/9, this is a clean win.

For context, an earlier attempt of mine — stopping on 150 ms peer idle — was 2.8× faster with `ok=False` on 7 of 8 cycles. I reverted it. That is why correctness, not speed, is the bar here.

## Verification done

- `bash lint` — clean
- `rp2350w` builds; RPC smoke PASS on hardware
- Both boards healthy throughout

### DIFF
--- a/examples/AutoResearch/AutoResearchNet.cpp
+++ b/examples/AutoResearch/AutoResearchNet.cpp
@@ -759,6 +759,7 @@ void pollNetServer() {}
 #include "fl/net/wifi.h"
 #include "fl/stl/array.h"
 #include "fl/stl/cstdio.h"
+#include "fl/stl/cstdlib.h"  // fl::atol for Content-Length
 #include "fl/stl/cstring.h"
 #include "fl/stl/singleton.h"
 #include "fl/stl/unique_ptr.h"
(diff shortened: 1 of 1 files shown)


zackees:
Closing — the baseline comparison I flagged as missing has now been run, and it rejects this change.

| Build | Cycles passing | HTTP time | Spread |
| --- | --- | --- | --- |
| **baseline (unmodified)** | **9/9** | 26.68 – 26.92 s | **< 0.25 s** |
| this PR | 6/9 | 8.1 – 20.4 s | 12.3 s |

Same probe, same C6 SoftAP, same nine cycles. The unmodified build is **100% reliable and near-deterministic** — which makes sense in hindsight: waiting out a fixed 2 s per request is slow, but it is also completely insensitive to when data actually arrives.

So 6/9 is not "equal correctness at 2.3x the speed". It is flakiness I introduced, and this PR would trade a slow-but-correct path for a fast-but-unreliable one. Same verdict as my earlier 150 ms idle-stop attempt (2.8x faster, 1/8 passing), for the same underlying reason.

The 12.3 s spread is the tell. If Content-Length parsing were working on every response, timings would cluster. They do not, which suggests some responses on this path do not carry a usable `Content-Length` — chunked encoding, or a header my parse misses — and those fall through to the 2 s deadline while others return promptly. A correct implementation has to handle the chunked terminator as well, which is exactly the part I skipped.

**The root-cause analysis stands** and is worth keeping (recorded in #4173):

- the drain loop exits on a 2 s deadline or peer close; `Connection: close` is requested but `client.connected()` does not go false promptly, so it waits out the deadline after the body has arrived
- 12 requests x ~2.2 s = the 26.57 s measured, almost all idle
- with `max_wait=30.0` and two calls per peer cycle, a healthy cycle has ~3.2 s of headroom

One thing the baseline run also settles: at 26.8 s the content checks pass **every time**, so the cycle-7 failure in `--net-peer` is not these assertions failing. It is the timing budget, or the second (C6 -> RP) direction. That narrows #4173 further.

Not merging this. The header/body separation in it is independently correct and could be salvaged, but it should not ride along with the completion logic that is not right yet.

Issue #4041, opened 2026-08-25
Color pipeline P7: gamut mapping and device solve (incl. RGBW)


zackees:
Phase P7 of the color pipeline implementation (see parent issue). Independently mergeable.

## Scope

Gamut mapping and the device solve — including the ≥4-emitter case.

- Host-side optimizer implements the reference objective: hue-preserving chroma compression in **OKLCh** (A3); exact in-gamut preservation within numeric budget, neutral/white-point preservation, explicit luminance-vs-chroma tradeoff, continuity under animation (§4).
- Embedded per-pixel path: **matrix + clamp/compress, optionally LUT-backed** with measured error bounds vs the host optimizer. Iterative optimization (incl. existing `nnls3`) restricted to profile/cache build at host/setup — a structural test enforces it never runs per-pixel (A3/B11 rule).
- Device solve consumes the emitter profile (chromaticities, response compensation); RGBW/RGBWW uses the deterministic allocation policy (C3): default **white-preferred**, per-profile RGB-preferred override, W included in the gamut volume the mapper targets, continuous in the input.
- Comparison study (§4, closes the survey): clipping vs max-normalization vs OKLCh compression vs constrained solve — pick the smallest algorithm meeting the A1 budget on embedded hardware, documented in the phase report.

## Decisions implemented

A3, C3, §4 objective, B11 (per-pixel restriction half).

## Acceptance criteria owned

- In-gamut vectors preserve chromaticity; neutral ramps remain neutral; out-of-gamut mapping is continuous and meets the hue/luminance objective.

## Dependencies

P5 (reference optimizer + vectors), P6 (working-domain core to plug into).

## Deliverables

Host optimizer (tooling), embedded mapper + solve stage, LUT generation path, continuity/preservation test suite, algorithm-selection report.

zackees:
## P7's two remaining evidence gaps are now measured (#4250)

The implementation for this phase has been in place for a while — host optimizer, embedded mapper and solve, the algorithm-selection report, the structural test that no iterative solver runs per pixel. What was still open was **evidence**, and `docs/color-gamut-algorithm-selection.md` named both gaps against itself rather than leaving them implicit.

### Gap 1 — connectivity above three emitters

> That is strong evidence for this device, not a proof, and **it says nothing about the >=4-emitter case where the zonotope gains a redundant generator.**

[...]

Issue #4034, opened 2026-08-25
Color pipeline: final implementation tracker


zackees:
## Status

Implementation addendum: [cross-phase contracts](https://github.com/FastLED/FastLED/blob/master/docs/color-pipeline-contracts.md), merged in #4157 following review #4156. It clarifies stage ordering, source/format boundaries (including the separate container and generic enums), power/presentation semantics, and tier/ownership requirements. These contracts supplement the digest below; implementation phases and the measured hardware closure gate remain open.

**This is the final implementation tracker for the FastLED color pipeline.** All design deliberation is closed. #4032 (spec) and #4033 (decision record) are **frozen historical references** — do not reopen their questions; every open item was resolved and the resolutions are digested below. Implementation work happens in the phase sub-issues attached to this issue. Any decision that proves wrong under measurement gets a follow-up decision issue referencing the ID it revises — not a silent divergence.

## The pipeline (normative)

```text
RGB8 + declared source profile
  -> decode to u16 linear (one semantic conversion: normalize + inverse transfer)
  -> signed s16.16-class accumulators: source primaries -> XYZ (Bradford CAT) -> gamut map (OKLCh objective, matrix/LUT execution)
  -> device solve against per-channel EmitterProfile (RGBW: white-preferred allocation)
  -> brightness x power: one multiplicative linear scalar stage, chromaticity-preserving
  -> chipset encoder: single final quantization into native wire format (per-chip 5-bit semantics, pipeline dithering)
```

Streaming execution inside `show()` via `PixelController` — no RGB16 framebuffer. Legacy path byte-identical when color management is disabled. Tree-shakes to zero when unused.

## Decision digest (complete — the "no more back and forth" list)

**Accuracy and reference (A1, A3, B7, B12):** Implementation fidelity: max ΔE2000 ≤ 0.5 and luminance ≤ 0.5% vs float64 reference at identity brightness, ≤ 1.0 ΔE2000 at 1/32 brightness, inclusive of the brightness/power stage. Profile accuracy: datasheet-derived median ≤ 3.0 / p95 ≤ 6.0; measured median ≤ 1.5 / max ≤ 4.0, at reference temperature. Relative colorimetry, profile white at full drive, dark surround. Bradford is the normative CAT. Gamut mapping: hue-preserving chroma compression in OKLCh computed by a host optimizer; embedded per-pixel path is matrix + clamp/compress (+ optional LUT) — iterative solves (incl. `nnls3`) never run per-pixel.

[...]

PR #4184, opened 2026-09-07
feat(math): fixed-point cube root for the P7 gamut mapper (#4041)

zackees:
The gamut mapper selected in `docs/color-gamut-algorithm-selection.md` works in OKLab, whose forward transform needs three cube roots per pixel. `s16x16` has `sqrt` but no `cbrt`, and the only cube root in the tree was `signedCubeRoot` in `colorutils.cpp.hpp` — float. This is the missing primitive.

## What it adds

`fl::icbrt64`, alongside `isqrt64`: a 22-step bit-by-bit integer cube root — no division, no multiplication by a non-constant, no float — written as tail recursion for the same C++11-constexpr reason `isqrt` is.

`s16x16::cbrt` builds on it through the identity `sqrt` already uses one power down: for a Q16 raw `r`, the root satisfies `y^3 = r * 2^32`, so it is the integer cube root of the raw shifted left by `2 * FRAC_BITS`.

One detail worth flagging for review: the comparison is `(x >> s) >= B`, not `x >= (B << s)`. The latter overflows u64 on the first steps of a large input. The shifted-right form is exact for integers, and `B << s` is then only evaluated on the branch where it is known to be `<= x`.

## The accuracy requirement was measured, not assumed

The precision table in the P7 report quantizes every value the mapper passes around — but it reached OKLCh through float64 cube roots. By that document's own standard ("a measurement that quantizes only the boundaries of a stage is not measuring that stage") it was measuring the stages on either side of the root rather than the root.

Re-scoring the corpus's 20 out-of-gamut vectors with the integer root in place, then perturbing it to find how much accuracy is actually needed:

| cube-root error | worst ΔE2000 |
| --- | --- |
| **exact** | **0.149** |
| ± 64 ULP | 0.200 |
| ± 256 ULP | 0.459 |
| ± 1024 ULP | 1.717 |

A1's budget is 0.5. The exact root reproduces the recorded 0.152, so the cube root costs the mapper nothing, and it clears the cliff by more than two orders of magnitude.

The perturbation is the more useful half. It says a *cheaper* approximate root is still on the table — anything holding under about 64 ULP would also pass — so if 22 steps per root ever proves too expensive on an 8-bit target, the replacement budget is known in advance rather than guessed at.

## Tests

Both ends are pinned rather than left as comments.

- `tests/fl/math/fixed_point_misc.cpp` asserts the *defining* property, `y^3 <= x < (y+1)^3`, in integer arithmetic — so it inherits no reference implementation's rounding and needs no float. Plus perfect cubes and neighbours, oddness, the `INT32_MIN` input (the one whose magnitude is not representable as a positive `i32`), and `FL_STATIC_ASSERT`s that both stay usable in constant expressions, since a change making them non-constexpr would still pass every runtime check.
- `ci/tests/test_color_gamut_study.py` pins the mapper scores at both ends of the table, so the numbers above stay checkable.

Two things the tests caught while being written, both mine: the float cross-check ran to exactly `32768.0`, which overflows `s16x16`'s float constructor to `INT32_MIN`; and the Python model's 1-ULP bound compared against the cube root of the *unquantized* input, which near zero is wrong by several ULP because the cube root amplifies input error (derivative ≈ 7 at v = 0.01).

Part of #4034 / #4041.

https://claude.ai/code/session_01MLhWkMfzLjrnLTDMBE6Fj9

## Summary by CodeRabbit

- **New Features**
  - Added integer and fixed-point cube-root support for color calculations.
  - Added quantized OKLCh computation to improve gamut-mapping precision and consistency.

- **Documentation**
  - Documented fixed-point cube-root accuracy, color-mapping impact, and supported error budgets.

- **Tests**
  - Added comprehensive coverage for cube-root accuracy, negative values, edge cases, constexpr behavior, and color-mapping regression limits.

### DIFF
--- a/ci/color_gamut_study.py
+++ b/ci/color_gamut_study.py
@@ -20,6 +20,7 @@ from typeguard import typechecked
 
 from ci.color_reference import (
     Matrix3,
+    Oklch,
     Xyz,
     _invert_3x3,
     _matvec,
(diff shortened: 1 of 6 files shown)


zackees:
Thanks — went through all five. Took one, declining four, with what I checked for each.

**Taken: hoist the test's local imports** (`ci/tests/test_color_gamut_study.py`). Fair. The file had *three* such blocks — two pre-existing plus mine, which I'd added by copying the surrounding pattern. They're plain module dependencies with no import cycle to break, so I consolidated all three rather than only mine. Pushed.

**Declining: remove the default values from `ulp_error`.** The cited rule does not exist in this repository. `agents/docs/python-standards.md` says nothing about default parameters, and neither does `CLAUDE.md`, `.coderabbit.yaml`, or any other doc — I grepped for `"default value"` across every `.md`/`.yaml`/`.yml`/`.toml` in the tree and the only hits are an Xtensa register description and an Asio example. Meanwhile `ci/` has **41** functions with defaulted parameters today. `ulp_error=0` is also the parameter's meaningful identity: callers that want the real cube root shouldn't have to write `0`.

**Declining: replace the list comprehensions with explicit loops.** Same situation — no such guideline exists here (`"list comprehension"` appears nowhere in the repo), and comprehensions appear in **29** files under `ci/`. The three in `oklch_q16` are one-line matrix rows; expanding them to `append()` loops would make the transform harder to read against the matrix it implements, not easier.

**Declining: ASCII-only C++ comments.** `icbrt.h` is a deliberate mirror of `isqrt.h` in the same directory, and the em dash flagged on line 4 is in **the identical sentence** in `isqrt.h` line 4: `// Bit-by-bit algorithm — no float, no division.` **54** files under `src/fl/math/` contain non-ASCII characters, and I found no linter enforcing this. Changing it here alone would make the two files differ for no reason.

**Declining: move `_icbrt64_step` into `fl::detail`.** `isqrt.h` puts `_isqrt64_start`, `_isqrt64_step`, `_isqrt32_start` and `_isqrt32_step` directly in `namespace fl` with exactly this leading-underscore convention. `icbrt.h` follows its sibling. If `fl::detail` is the direction these should go, that's a change worth making to both files together rather than leaving the pair inconsistent — happy to do it as a follow-up if a maintainer wants it.

Happy to revisit any of these if the guideline lives somewhere I didn't look.

zackees:
The ESP32-S3 shard 5 failure is infrastructure, not this change — fbuild says so itself:

```
daemon error: daemon did not become healthy after 3 spawn attempts (10s each).
This is a daemon problem, not a defect in the code being built — no compilation was attempted.
```

No compiler ran. The other seven shards passed, and #4185 — which contains these exact commits plus a new gfx module — passed all eight. Re-running the failed job.

@coderabbitai resolve

(For the record: one of the five findings was taken and pushed; the other four are declined with evidence in my comment above — two cite guidelines that don't exist in this repository, and two contradict `isqrt.h`, the sibling file `icbrt.h` deliberately mirrors. Happy to reopen any of them if a maintainer disagrees.)

PR #4202, opened 2026-09-07
feat(autoresearch): add --bitbang driver flag

zackees:
Adds `bash autoresearch <board> --bitbang`, selecting `Bus::BIT_BANG`.

## Why

`BIT_BANG` is AUTO's last-resort fallback (priority 0), but no autoresearch flag
selected it, so the portable bit-bang path had **no hardware-in-the-loop coverage
on any platform**. FastLED#3899's checklist item "AUTO and BIT_BANG fallback
behavior" could not be run as written because of this gap.

The driver name `BIT_BANG` matches `BitBangChannelDriver::getName()`, which the
firmware resolves against the manager's registered drivers at runtime — same
mechanism the existing driver flags use.

## Result

The flag immediately earned its keep: bit-bang fails to drive WS2812 on both
bench boards, 100% byte corruption, filed as a separate issue with wire captures
and a root-cause analysis. This PR is just the measurement tool; it makes no
behavioral change to the library.

## Verification

- `bash test --cpp` — 383/383 (299 unit + 84 examples)
- `bash lint` — clean
- Flag parsing verified not to leak into other driver flags
- Exercised end-to-end on RP2350W and ESP32-C6 hardware

## Summary by CodeRabbit

- **New Features**
  - Added a `--bitbang` option for selecting a portable GPIO fallback driver.
  - The selected fallback driver is included in targeted test runs.
  - The fallback driver is preserved when `--bitbang` is combined with `--all`, ensuring the explicit selection is honored alongside the standard driver set.

- **Bug Fixes**
  - Improved serial interface handling during loopback and public API test execution to support reliable handoff to test processes.

### DIFF
--- a/ci/autoresearch/args.py
+++ b/ci/autoresearch/args.py
@@ -18,6 +18,7 @@ class Args:
     rmt: bool
     spi: bool
     uart: bool
+    bitbang: bool
     lcd: bool
     lcd_spi: bool
     lcd_rgb: bool
@@ -294,6 +295,16 @@ See Also:
             "--uart",
             action="store_true",
             help="Test only UART driver (LPC845: UART DMA clockless RX-DMA loopback)",
+        )
+        driver_group.add_argument(
+            "--bitbang",
+            action="store_true",
+            help=(
+                "Test only the portable GPIO bit-bang driver (Bus::BIT_BANG). "
+                "This is AUTO's last-resort fallback on every platform, so it "
+                "is the one driver that must keep working when the hardware "
+                "engines are all claimed."
+            ),
         )
         driver_group.add_argument(
             "--rp-uart-index",
@@ -888,6 +899,7 @@ See Also:
             rmt=parsed.rmt,
             spi=parsed.spi,
             uart=parsed.uart,
+            bitbang=parsed.bitbang,
             lcd=parsed.lcd,
             lcd_spi=parsed.lcd_spi,
             lcd_rgb=parsed.lcd_rgb,
--- a/ci/autoresearch/phases.py
+++ b/ci/autoresearch/phases.py
@@ -573,6 +573,11 @@ def _parse_args_and_build_commands(args: Args) -> RunContext | int:
                     rp_uart_index=args.rp_uart_index,
                 )
             )
+        if args.bitbang:
+            # Portable GPIO fallback. Name matches
+            # BitBangChannelDriver::getName(), which the firmware resolves
+            # against the manager's registered drivers at runtime.
+            drivers.append("BIT_BANG")
         if args.lcd:
             drivers.append("LCD_CLOCKLESS")
         if args.lcd_spi:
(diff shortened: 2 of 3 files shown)


zackees:
The underlying bug is real and fixed, though I implemented it differently from the suggestion — reasoning below.

**The bug:** the `args.all` branch replaces the driver selection wholesale, so `--all --bitbang` silently dropped the explicitly requested driver. Agreed, and that is now fixed.

**Why not fold BIT_BANG into the two `--all` lists**, as suggested:

1. Those lists are curated to each platform's *real transmit engines*, and there is a test asserting exactly that — `test_all_drivers_teensy4_only_real_teensy_drivers` expects `["OBJECT_FLED", "FLEX_IO"]`. BIT_BANG is the universal software fallback rather than a platform driver, so adding it changes what `--all` means, not just what it emits.
2. BIT_BANG currently cannot drive WS2812 on RP at all (#4203 — measured 3146 ns/bit against a ~150 ns budget). Folding it into `--all` would make every `--all` run on RP fail for as long as that issue is open, converting a known and tracked driver defect into a red sweep on unrelated work.

So BIT_BANG is appended after the selection when `args.bitbang` is set. That fixes the silent-omission bug — the actual defect — without redefining `--all`. New test asserts `--all --bitbang` yields `["OBJECT_FLED", "FLEX_IO", "BIT_BANG"]`.

Happy to fold it into the lists instead if the intent really is for `--all` to sweep software fallbacks too, but that reads like a separate decision to me, and one that should probably wait for #4203.

**Separately, this PR was breaking 152 tests and I had not noticed.** It adds `bitbang` to the `Args` dataclass without updating the `_make_args` helper that constructs it, so every test building an `Args` raised `TypeError: Args.__init__() missing 1 required positional argument: 'bitbang'`. The branch was also stale, which is what hid it. Rebased onto current master and fixed the helper — 178 tests pass now.

zackees:
This is the same suggestion as the earlier round, so restating the position rather than changing it — happy to be overruled if the intent is different from what I inferred.

The underlying bug — `--all --bitbang` silently dropping the explicit driver — **is fixed**, by appending `BIT_BANG` after the selection when `args.bitbang` is set. A test asserts `--all --bitbang` yields `["OBJECT_FLED", "FLEX_IO", "BIT_BANG"]`.

What I have not done is add `BIT_BANG` to the two `--all` lists themselves, for two reasons that still hold:

1. Those lists enumerate each platform's **real transmit engines**, and `test_all_drivers_teensy4_only_real_teensy_drivers` asserts exactly `["OBJECT_FLED", "FLEX_IO"]` for Teensy. BIT_BANG is the universal software fallback, so including it changes what `--all` *means*, not just what it emits.
2. BIT_BANG cannot drive WS2812 on RP at all (#4203 — measured 3146 ns/bit against a ~150 ns budget). Folding it into `--all` makes every `--all` run on RP fail for as long as that issue is open, converting a known and tracked driver defect into a red sweep on unrelated work.

If `--all` should sweep software fallbacks too, that is a deliberate scope change worth making on its own — and probably after #4203, so it does not land as an immediate red. Say the word and I will do it; I did not want to make that call silently inside a flag-addition PR.

Issue #4203, opened 2026-09-07
BIT_BANG driver cannot drive WS2812: constant ~2-3us per-writeByte overhead stretches every bit (RP2350W + ESP32-C6)


zackees:
## Summary

`BitBangChannelDriver` (`Bus::BIT_BANG`) cannot drive a WS2812 strip. Every bit is
stretched by a constant ~2.3-3.1us because `DigitalMultiWrite8::writeByte()` costs
far more than the 400ns timing budget it is supposed to hit. Confirmed on two
different architectures with wire captures.

This matters because `BIT_BANG` is registered at **priority 0 as AUTO's last-resort
fallback** on ESP32. When the hardware engines are exhausted, AUTO degrades to a
driver that emits undecodable output rather than to a working one.

## Evidence

Wire captures via `bash autoresearch <board> --bitbang` (flag added in #4202),
100 LEDs, WS2812B-V5 timing. Both boards: **300/300 bytes corrupt (100%)**.

**RP2350W** (Cortex-M33 @150MHz) — after locally registering `Bus::BIT_BANG`, see below:
```
Raw sample: H3450 L8450 H3450 L8400 H7050 L4850 H7050 L4800 ...
Capture status: RX produced no decodable bytes
```

**ESP32-C6** (RISC-V @160MHz) — registers `BIT_BANG` today:
```
TEST BIT_BANG lanes=1 leds=100 FAIL 4843ms class=decode_mismatch
Raw sample: H2650 L6500 H5375 L3725 H5375 L3750 ...
```

The encoding is *well-formed* — a clean 1:2 ratio between "0" and "1" high times —
just far too slow. Nominal WS2812 is T0H~400ns, T1H~800ns, period ~1250ns.

## Root cause

Not the delay math. A constant-overhead-per-`writeByte()` model fits both boards:

| board | implied C per writeByte | T1H predicted | T1H measured | error |
|---|---|---|---|---|
| RP2350W  | 3050ns | 6900ns | 7050ns | -2.1% |
| ESP32-C6 | 2250ns | 5300ns | 5375ns | -1.4% |

Predicted bit period is off by an identical -12.6% on *both* boards, i.e. one further
constant per-bit cost in the byte/bit loop. Two unrelated architectures showing the
same structure points squarely at shared code:

`src/fl/system/pins.cpp.hpp` — `applyNibble()` is a per-pin loop of out-of-line calls:
```cpp
[...]

PR #4204, opened 2026-09-07
perf(delay): drop the 64-bit divide from delayNanoseconds, with the probe that found it (#4203)

zackees:
Fixes the dominant term in #4203, plus the AutoResearch probe that located it.

## The bug

`cycles_from_ns_*()` computed this on every call:

```cpp
return ((u64)ns * (u64)hz + 999999999UL) / 1000000000UL;
```

Neither Cortex-M33 nor RV32 has a hardware 64-bit divide, so it lowers to a
software routine — and `fl::delayNanoseconds()` runs three times per clockless
bit. The same expression was duplicated across six platform delay headers.

## How it was found

Two attributions from reading the source (the per-pin `digitalWrite` loop, then
the ESP-IDF clock query) were both wrong. The `perfProbeBitBangCost` RPC in this
PR settles it by measurement: it times `delayNanoseconds(ns)` against
`delayNanoseconds<NS>()`, where the compile-time form folds the conversion to a
constant while running the identical busy-wait. The gap is the conversion.

## Results

RP2350W, 20000 iterations, `ns=400`, loop floor 0.059us:

| | before | after |
|---|---|---|
| `delayNanoseconds(400)` | 3.020us | **0.655us** |
| ns conversion | 2.533us | **0.293us** |
| predicted per-bit overhead | 10.247us | **3.146us** |
| `delayNanoseconds<400>()` *(control)* | 0.469us | 0.469us |
| `writeByte()` *(control)* | 0.511us | 0.511us |

Both controls are unchanged, which is the evidence that the change is confined
to the conversion and did not perturb the busy-wait or the pin-write path.

## Approach

All six headers already include `platforms/cycle_type.h`, so the shared helper
lands there and each site delegates — no new includes. The fast path scales the
clock to kHz to keep arithmetic in 32 bits so the compiler can use a
multiply-shift, bounded so `ns * (hz/1000) + 999999` cannot overflow u32
(4.001e9 at the 1GHz ceiling). Outside the bound it uses the exact 64-bit form.

## Scope

This is not a BIT_BANG-only fix. Any caller of `fl::delayNanoseconds(ns)` with a
runtime value on any affected platform was paying this.

**It does not by itself make `Bus::BIT_BANG` usable.** A WS2812 bit still costs
~4.4us against a 1.25us budget. It removes the dominant term; the remaining
overhead is now split roughly evenly between the delay call and
`DigitalMultiWrite8::writeByte()`, so the port-masked-write work is the
necessary second half. #4203 stays open.

## Verification

- `bash test --cpp` — 385/385
- `bash lint` — clean
- New `tests/platforms/cycle_type.cpp`: asserts exact equality against the
  pre-change expression as an oracle across 16 clock rates from 8MHz to 1GHz,
  plus the overflow bound and compile-time folding
- Before/after measured on RP2350W hardware

## Summary by CodeRabbit

* **New Features**
  * Added a performance benchmarking command for measuring timing and GPIO bit-banging overhead.
  * Supports configurable iteration counts, timing intervals, and optional pin selection.
  * Compares runtime, frequency-adjusted, compile-time, and cycle-counted delays with digital byte output.
  * Reports loop timing, delay overhead, clock-query and conversion costs, and estimated per-bit output overhead.
  * Added command details, arguments, results, and performance information to RPC help.

* **Improvements**
  * Standardized nanosecond-to-cycle conversion across supported platforms while preserving rounded-up timing behavior.

### DIFF
--- a/examples/AutoResearch/AutoResearchRemoteBenchmarkMethods.cpp
+++ b/examples/AutoResearch/AutoResearchRemoteBenchmarkMethods.cpp
@@ -233,6 +233,31 @@ void AutoResearchRemoteControl::bindBenchmarkMethods(fl::Remote& remote) {
         }
         const fl::u32 delay_hz_us = fl::micros() - delay_hz_t0;
 
+        // Compile-time NS: cycles_from_ns_*() folds to a constant, so the
+        // runtime u64 divide disappears while the platform busy-wait still
+        // runs. If this is fast, the divide carried the cost; if it is still
+        // slow, the cost is inside the busy-wait. Fixed at 400ns (WS2812 T0H)
+        // because a template argument cannot come from the RPC payload -- the
+        // `ns` parameter only steers the runtime variants above.
+        const fl::u32 delay_ct_t0 = fl::micros();
+        for (int i = 0; i < iterations; ++i) {
+            fl::delayNanoseconds<400>();
+        }
+        const fl::u32 delay_ct_us = fl::micros() - delay_ct_t0;
+
+        // Pure cycle-counted loop: no ns->cycles conversion and no platform
+        // busy-wait wrapper. This is the floor a hand-rolled bit loop could
+        // reach. 60 cycles ~= 400ns at 150MHz. delaycycles<> is specialized
+        // only up to 50 with no generic fallback, so 60 is composed from two
+        // specializations rather than written as delaycycles<60>(), which is
+        // an undefined reference at link time.
+        const fl::u32 delay_cyc_t0 = fl::micros();
+        for (int i = 0; i < iterations; ++i) {
+            fl::delaycycles<50>();
+            fl::delaycycles<10>();
+        }
+        const fl::u32 delay_cyc_us = fl::micros() - delay_cyc_t0;
+
         // pin < 0 leaves every slot inactive, so applyNibble performs no GPIO
         // write at all — that isolates the LUT and call overhead from the
         // cost of the pin write itself.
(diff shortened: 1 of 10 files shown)


zackees:
Post-merge expectation, so this is not oversold: I ran the bit-bang wire capture on hardware with this applied, and **BIT_BANG still fails** — 100% byte corruption on the ESP32-C6.

| ESP32-C6 | before | after | WS2812 |
|---|---|---|---|
| T0H | 2650ns | 2100ns | 400ns |
| T1H | 5375ns | 4325ns | 800ns |
| bit period | ~9150ns | ~7550ns | 1250ns |

The improvement is real and the expected size (~1600ns off the period, about 3x the ~550ns saved per phase, matching three delay calls per bit), but still 6x over budget.

**This PR should be judged on its own merits, not on #4203.** It fixes a ~2.5us-per-call cost in `fl::delayNanoseconds()` that every runtime caller on six platforms was paying. That is worth having whether or not bit-bang is ever made to work — and per the analysis in #4203, making bit-bang work needs its inner loop restructured, not further optimization of these callees.

zackees:
This is a genuine correctness bug in the PR, not a style point. Confirmed and fixed.

The fast path folds `hz / 1000u`, and truncating that division under-counts for any clock that is not a whole number of kHz:

```
cycles_from_ns(8,    125000001)   fast=1    exact=2
cycles_from_ns(1000, 150000003)   fast=150  exact=151
```

**Under-delaying is a protocol violation**, not a rounding nicety — holding a WS2812 line for one cycle less than required is precisely the failure this function exists to prevent. So this was worse than a cosmetic inaccuracy.

The fast path now additionally requires `(hz % 1000u) == 0`. Every realistic MCU clock is a whole number of kHz (8/16/125/150/240 MHz), so the optimisation still applies wherever it matters.

### Why my original verification missed it

Worth stating, because it was thorough and still wrong. I ran an exhaustive equivalence check of the fast path against the 64-bit oracle — but across **16 clock rates from 8 MHz to 1 GHz, every one of them a multiple of 1000**. The truncation had nothing to truncate, so the check could not have found this no matter how many `ns` values it swept.

That is the second time on this PR that a thorough-looking check had a blind spot in its inputs rather than its logic: the earlier overflow bug was hidden because the Python equivalence check used bignums and never overflowed, and only an explicit overflow counter found it.

The new test sweeps rates deliberately offset by a few Hz (`hz = 1000001, +7919 ...`) alongside the two specific cases above, so the input space now includes the shape that breaks it. `bash test cycle_type --cpp` passes.

PR #4206, opened 2026-09-07
fix(autoresearch): three RP SPI harness bugs that reported host failures as device faults

zackees:
Three independent harness bugs found while running the PIO/SPI/UART driver
matrix for #3832 on an attached RP2350W. All three share a failure shape:
**a host-side problem presented as a device or firmware fault.**

## 1. `--spi` never reached the RP SPI engine

```
Error: UnknownDriver
Message: Driver 'SPI' not available
```

RP registers its two fixed PL022 blocks as `SPI0`/`SPI1`; the portable `SPI`
bus name is not a runtime driver there. `_driver_name_for_environment()`
already mapped `FLEX_IO -> PIO{n}` and `UART -> UART{n}` for RP — SPI was
simply missing and fell through to the portable name. `--rp-spi-index`
existed but was only consumed by the loopback / public-api paths.

Verified on hardware: the harness now emits `"driver":"SPI0"` and the engine
resolves it. ESP32 and Teensy mappings unchanged.

**Scope note:** this makes `--spi` *reach* the engine, not pass with it.
`ChannelEngineRpSpi::canHandle()` requires an `SpiChipsetConfig` (clocked
APA102/SK9822) and its pin table restricts SPI0 MOSI to `{3,7,19,23}` / SCK
`{2,6,18,22}`. Driving it with the default WS2812 clockless timing cannot
succeed by design.

## 2. A failed `rpc.discover` was reported as missing firmware

```
FAIL — deployed firmware schema does not contain rpSpiLoopback
```

on a board that **does** expose it — querying the device directly returns
both `rpSpiLoopback` and `rpSpiPublicApiLoopback`. `RpcBench.call_flat()`
collapses `RpcTimeoutError` and `RpcError` to `None`, and the caller read any
non-dict as "no methods discovered".

That message asserts a build-configuration gap and sends the reader hunting
for a missing `FASTLED_AUTORESEARCH_LOW_MEMORY` branch or an unbuilt source
file. I went down that path myself before checking the device.

Now distinguishes three cases: discover returned no object at all (stated as
a host/transport failure), returned an object with no parseable method names
(shape changed, keys printed), or the method genuinely absent (with the
discovered method count).

## 3. `rpc.discover` called through the wrong API

`call_flat()` is for methods bound with positional C++ parameters; the
built-in `rpc.discover` should use `call()`.

## The bug behind #2 that this PR does *not* fix

Verified cause of the `NoneType`: under `bash autoresearch`, the harness
already holds an `RpcBench` on the same port — its own fetch logs `✅ Fetched
schema: 71 methods` in the same run — and the script's second connection
never sees its responses. **Run standalone and the identical code path
succeeds**, reaches the device and executes the loopback:

```
uv run python -m ci.autoresearch.test_rp_spi_loopback --port /dev/ttyACM0 --spi-index 0
  rpSpiLoopback(SPI0, 1,000,000 Hz)
    FAIL — RP SPI loopback failed.
      expected: [0, 255, 85, 170, 18, 52, ...]
      received: [0, 0, 0, 0, 0, 0, ...]   wireIdle: True
```

That remaining failure is **physical**: all-zero MISO with the wire idle is
an unfitted GPIO3 (MOSI) -> GPIO0 (MISO) jumper. This bench has GPIO0<->GPIO1
(what pin discovery selects for the clockless tests), not GPIO3->GPIO0. Not a
driver defect.

I have left the two-connection issue unfixed rather than guess at it again —
see below.

## Method note

I proposed three wrong causes for the `NoneType` before testing standalone: a
short discover timeout, then the `call_flat` frame shape, then payload size.
Each was refuted by the next run. The root cause was only visible by
comparing the harness run against a standalone run — and the fix in #2 is
what made the failure falsifiable at all, since it replaced the false
firmware claim with `got NoneType`.

## Verification

- PIO0 / PIO1 / PIO2 all `PASS 683ms` on the same board (the matrix this work
  was serving)
- UART0 correctly refuses with `RP UART: chipset timing needs a baud above
  the UART backend maximum` rather than emitting garbage
- `bash lint` clean
- Both bench boards enumerated and answered serial RPC throughout; neither
  wedged

## Summary by CodeRabbit

* **Bug Fixes**
  * SPI requests on active RP2040 and RP2350 environments now use the correct platform-specific driver names, including automatic handling when targeting all drivers.
  * SPI operations retain the correct default frame count across supported SPI driver variants.
  * SPI loopback validation now more reliably detects discovery and transport failures, including malformed responses and unavailable firmware methods.
  * Test harnesses now close serial connections before launching related processes, preventing communication conflicts.
  * Legacy mode rejects unsupported chipsets instead of applying incorrect timing defaults.
  * The `ws2811-400` chipset now uses the correct 400 kHz configuration.

### DIFF
--- a/ci/autoresearch/phases.py
+++ b/ci/autoresearch/phases.py
@@ -103,9 +103,16 @@ def _driver_name_for_environment(
     final_environment: str | None,
     rp_pio_index: int = 1,
     rp_uart_index: int = 0,
+    rp_spi_index: int = 0,
 ) -> str:
     if driver == "SPI" and _is_teensy4_environment(final_environment):
         return "SPI_UNIFIED"
+    if driver == "SPI" and _active_rp2xxx_environment(final_environment) is not None:
+        # RP registers its two fixed PL022 blocks as concrete "SPI0"/"SPI1"
+        # engines; the portable "SPI" bus name is not a runtime driver here, so
+        # sending it fails with `UnknownDriver: Driver 'SPI' not available`.
+        # Same shape as the FLEX_IO and UART cases below.
+        return f"SPI{rp_spi_index}"
     if (
         driver == "FLEX_IO"
         and _active_rp2xxx_environment(final_environment) is not None
@@ -564,7 +571,13 @@ def _parse_args_and_build_commands(args: Args) -> RunContext | int:
         if args.rmt:
             drivers.append("RMT")
         if args.spi:
-            drivers.append(_driver_name_for_environment("SPI", final_environment))
+            drivers.append(
+                _driver_name_for_environment(
+                    "SPI",
+                    final_environment,
+                    rp_spi_index=args.rp_spi_index,
+                )
+            )
         if args.uart:
             drivers.append(
                 _driver_name_for_environment(
--- a/ci/autoresearch/phases.py
+++ b/ci/autoresearch/phases.py
@@ -98,6 +98,19 @@ def _is_teensy4_environment(final_environment: str | None) -> bool:
     )
 
 
+def _is_spi_family_driver(driver: str) -> bool:
+    """Whether a driver name denotes an SPI transmitter on any platform.
+
+    Platforms spell this differently: the portable "SPI", Teensy's
+    "SPI_UNIFIED", and RP's concrete per-block "SPI0"/"SPI1". Anything
+    keying off the bare literal silently skips the RP names, which is how
+    the two-frame SPI default stopped applying there.
+    """
+    if driver in ("SPI", "SPI_UNIFIED"):
+        return True
+    return driver.startswith("SPI") and driver[3:].isdigit()
+
+
 def _driver_name_for_environment(
     driver: str,
     final_environment: str | None,
@@ -165,6 +178,9 @@ def _normalize_deferred_driver_names(ctx: RunContext) -> None:
     args = ctx.args
     if _active_rp2xxx_environment(environment) is not None:
         _replace_driver_selection(ctx, "UART", [f"UART{args.rp_uart_index}"])
+        # --all appends the portable "SPI", which is not a runtime driver on
+        # RP; without this it reaches the device as UnknownDriver.
+        _replace_driver_selection(ctx, "SPI", [f"SPI{args.rp_spi_index}"])
         pio_names = (
             ["PIO0", "PIO1"]
             if args.rp_pio_both and args.flex_io
@@ -1061,7 +1077,7 @@ def _parse_args_and_build_commands(args: Args) -> RunContext | int:
                     # others -> 1. User can override with --frames N.
                     if args.frames is not None:
                         frame_count = args.frames
-                    elif driver == "SPI":
+                    elif _is_spi_family_driver(driver):
                         frame_count = 2
                     else:
                         frame_count = 1
--- a/ci/autoresearch/test_rp_spi_loopback.py
+++ b/ci/autoresearch/test_rp_spi_loopback.py
@@ -109,14 +109,36 @@ def main() -> int:
     try:
         with RpcBench(args.port) as bench:
             schema = bench.call_flat("rpc.discover", args=[])
-            methods = schema.get("schema", []) if isinstance(schema, dict) else []
+            # `call_flat` collapses a timeout or transport error to None, so a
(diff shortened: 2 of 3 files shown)


zackees:
All three points valid and applied. Two were real bugs, one was a stale comment of mine.

**1. Two-frame SPI default never applied on RP.** `elif driver == "SPI"` keys off the bare literal, but this PR is precisely what makes the RP driver name `SPI0`/`SPI1` — so RP SPI silently ran with `frame_count=1` and lost the #2254/#2288 second-frame degradation check the default exists to provide. Added a shared predicate:

```python
def _is_spi_family_driver(driver: str) -> bool:
    if driver in ("SPI", "SPI_UNIFIED"):
        return True
    return driver.startswith("SPI") and driver[3:].isdigit()
```

Covered by a test across every platform spelling plus the near-miss negatives — `LCD_SPI` and `I2S_SPI` contain "SPI" but must not match, and `SPIX` starts with it but is not an index.

**2. `--all` still reached the device as `UnknownDriver`.** `_normalize_deferred_driver_names()` resolved `UART` and `FLEX_IO` for RP but not `SPI`, and the `--all` branch appends the portable `"SPI"` literally. So the exact failure this PR fixes survived on that path. Now normalized alongside the other two.

**3. `rpc.discover` triage — and my comment was describing the wrong function.** `RpcBench.call()` has three outcomes that mean different things: `METHOD_NOT_FOUND` (firmware gap), `None` (timeout/transport), and a non-dict result (malformed schema). I collapsed all of them into one "host/transport failure" message, and the comment justified it by citing `call_flat`'s collapse-to-`None` behaviour — but the code calls `call()`, which does not collapse. Each case now reports separately, since they send the reader to different places.

Verified: `bash test --cpp` clean, 178 phases tests pass, and on RP2350W `2DCB876B587EA334`:

```
RESULT: RPC smoke PASS (discovery, help, ping, payload, status, drivers,
        testNoSerial, RP concurrency, error handling)
```

The SPI loopback legs themselves still need the `GPIO3 → GPIO0` / `GPIO11 → GPIO8` jumpers documented on #3899, so those remain unexercised on this fixture.

zackees:
The `build / build` failure is not from this PR. Re-running it.

The compile error is in the Adafruit SAMD vendor core:

```
error: 'USB_PID' was not declared in this scope
error: 'USB_VID' was not declared in this scope
  .../adafruit-adafruit-samd-1.7.16/.../cores/arduino/...
```

This branch changes three files, all Python:

```
ci/autoresearch/phases.py               | 33 ++++--
ci/autoresearch/test_rp_spi_loopback.py | 47 +++++++--
ci/tests/test_autoresearch_phases.py    | 16 ++++
```

None of them participates in a SAMD C++ build. The same run also logged `fbuild binary not found in $INSTALL_DIR/{bin,Scripts}; cache likely corrupt` and a failed `.cache/js-tools` setup before reaching the compile, which points at a corrupted runner cache rather than anything in the diff.

Branch is also now up to date with master (it was 22 commits behind). That merge required resolving one conflict in `test_autoresearch_phases.py`, where my SPI-family predicate test landed at the same point as tests master had since merged — both kept, 180 tests pass locally.

zackees:
Good catch, and worse than a style problem — those two tests had stopped running entirely.

My conflict resolution placed `test_legacy_rejects_chipsets_without_a_legacy_template` and `test_non_legacy_still_accepts_those_chipsets` **inside** the module-level `test_spi_family_predicate_covers_every_platform_spelling`, rather than restoring them as methods of `TestRunTestsOrSpecialMode` where master has them. Nested functions are not collected.

What makes this nasty is that it is invisible in a pass/fail summary. `pytest` reported **180 passed** before and after the fix, because the two missing tests were simply absent from the count rather than failing. My post-merge check saw a green run and a plausible number.

The collection count is the evidence that a pass count cannot give:

```
before:  178 collected,  0 matching the two test names
after:   180 collected,  2 matching
```

Both pass now. Restored to class scope; my module-level predicate test stays where it was.

This is the "a test that does not test" failure mode from the lessons file, arriving by a route I had not considered — not a weak assertion, but a mechanical conflict resolution that quietly changed a function's scope. Worth checking collection counts, not just pass counts, after resolving a conflict in a test file.

Issue #4207, opened 2026-09-07
autoresearch: harness-spawned device scripts get no RPC responses while the harness holds its own connection


zackees:
## Symptom

`bash autoresearch rp2350w --rp-spi-loopback` fails before reaching the device:

```
FAIL — rpc.discover did not return a schema object (got NoneType)
```

The **same script, same board, same code path, run standalone, works**:

```
uv run python -m ci.autoresearch.test_rp_spi_loopback --port /dev/ttyACM0 --spi-index 0
  rpSpiLoopback(SPI0, 1,000,000 Hz)
    FAIL — RP SPI loopback failed.
      capture: {'driver': 'SPI0', 'sckPin': 2, 'misoPin': 0, 'mosiPin': 3,
                'expected': [0, 255, 85, 170, ...],
                'received': [0, 0, 0, 0, ...], 'wireIdle': True, ...}
```

(That standalone failure is a separate, physical matter — an unfitted GPIO3->GPIO0 jumper. The point here is that it *reached the device and executed*, which the harness-invoked run never does.)

## What is different

`_run_rp_spi_loopback_tests` in `ci/autoresearch/phases.py` spawns the script as a subprocess:

```python
cmd = ["uv", "run", "python", "-m", "ci.autoresearch.test_rp_spi_loopback",
       "--port", upload_port, ...]
result = RunningProcess.run(cmd, check=False, timeout=ctx.remaining_seconds())
```

while the parent harness still holds its own `RpcBench`/serial connection to that port. The parent's own schema fetch succeeds in the same run, immediately before:

```
🔍 Validating RPC commands against device schema...
✅ Fetched schema: 71 methods
✅ All 0 RPC commands validated against schema
...
FAIL — rpc.discover did not return a schema object (got NoneType)
```

So one connection to the port works and the second gets nothing back.

## What I have NOT established

The mechanism inside fbuild's serial daemon. `SerialManager::attach_reader()` hands out a `broadcast::Receiver` and `reader_client_ids` is a set, so *reads* look like they should fan out to both clients. I did not verify whether a second client's **writes** reach the device, which is the obvious next thing to check but which I have no evidence for either way.

I am deliberately not naming a cause. I proposed three for the `NoneType` before comparing harness-vs-standalone — a short discover timeout, the `call_flat` frame shape, and payload size — and each was refuted by the next run. The harness/standalone difference is the only thing I have actually demonstrated.

[...]

PR #4208, opened 2026-09-07
fix(autoresearch): release the harness serial port before spawning a device script (#4207)

zackees:
Fixes the workflow half of #4207.

## The bug

A second `RpcBench` on a port this process already holds **connects without
error and is then completely inert** — every call returns `None` — while the
first client keeps working. Minimal repro, no harness and no subprocess:

```python
a = RpcBench(port); a.call("ping")   # -> {'message': 'pong', ...}
b = RpcBench(port); b.call("ping")   # -> None
a.call("ping")                       # -> {'message': 'pong', ...}
```

`_run_rp_spi_loopback_tests` spawns `test_rp_spi_loopback` as a subprocess
against `upload_port` while still holding `ctx.serial_iface`, so the child got
a silently dead client. Its schema probe returned `None`, which it reported as

```
FAIL — deployed firmware schema does not contain rpSpiLoopback
```

against a board that does expose the method (a direct query returns both
`rpSpiLoopback` and `rpSpiPublicApiLoopback`).

## The fix

This phase is terminal for `--rp-spi-loopback` — the dispatcher returns its
result directly — so the harness's interface can simply be closed before
handing the port over.

## Verified on an RP2350W

Before, the harness-invoked run never reached the device. After:

```
  rpSpiLoopback(SPI0, 1,000,000 Hz)
    FAIL — RP SPI loopback failed.
      expected: [0, 255, 85, 170, 18, 52, ...]
      received: [0, 0, 0, 0, 0, 0, ...]   wireIdle: True
FAIL — RP SPI loopback did not meet the byte-exact contract.
```

which matches the standalone run byte for byte. **That remaining failure is
physical** — an unfitted GPIO3 (MOSI) -> GPIO0 (MISO) jumper on this bench,
where the fitted pair is GPIO0<->GPIO1 for the clockless tests. Not a software
defect, and not something this PR claims to fix.

The value here is that the failure now names the real cause instead of
asserting a firmware build gap.

## Not addressed

- The underlying contract problem in #4207: a second attach should either work
  or fail loudly, rather than returning a client that looks healthy and
  answers nothing. A silent `None` propagates into callers as a statement
  about the device.
- `_run_rp_spi_public_api_tests` spawns a child the same way and likely needs
  the same treatment. I have not exercised that path, so I have not touched it.

## Verification

- `bash lint` clean (including the KBI checker — the teardown handler needs an
  explicit `except KeyboardInterrupt`, KBI001)
- Both bench boards enumerated and answered serial RPC after the run; the RP's
  mid-deploy BOOTSEL drop re-enumerated immediately. Neither wedged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed loopback test execution so the test process can reliably access the upload port and correctly detect the connected board’s capabilities.

### DIFF
--- a/ci/autoresearch/phases.py
+++ b/ci/autoresearch/phases.py
@@ -3736,6 +3736,24 @@ async def _run_rp_spi_public_api_tests(ctx: RunContext) -> int:
     print(
         f"{chip_name} SPI1 public API loopback: GPIO11 MOSI -> GPIO8 MISO, GPIO10 SCK"
     )
+
+    # FastLED#4207, same as the loopback phase above: the child opens its own
+    # RpcBench on this port, and a second client on a port we already hold
+    # connects fine and then answers nothing. Confirmed on an RP2350W --
+    # without this the child reported
+    # "FAIL — rpSpiPublicApiLoopback dispatch result: None" while the harness's
+    # own schema fetch in the same run succeeded. This phase is terminal, so
+    # release the interface before handing the port over.
+    if ctx.serial_iface is not None:
+        try:
+            await ctx.serial_iface.close()
+        except KeyboardInterrupt as ki:
+            handle_keyboard_interrupt(ki)
+            raise
+        except Exception as exc:  # noqa: BLE001 - teardown must not mask the test
+            print(f"   (warning: releasing harness serial interface: {exc})")
+        ctx.serial_iface = None
+
     result = RunningProcess.run(
         [
             "uv",
--- a/ci/autoresearch/phases.py
+++ b/ci/autoresearch/phases.py
@@ -3684,6 +3684,23 @@ async def _run_rp_spi_loopback_tests(ctx: RunContext) -> int:
     print("   Rates: 1 MHz, 8 MHz, 24 MHz; byte-exact MISO capture + wire-idle")
     print("=" * 60)
     print()
+    # FastLED#4207: a second RpcBench on a port this process already holds
+    # connects without error and then answers nothing -- every call returns
+    # None. The child would fail its schema probe and report it as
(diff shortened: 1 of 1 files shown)


zackees:
**Extended to the sibling path, with hardware evidence rather than the guess I originally left.**

The first version of this PR said `_run_rp_spi_public_api_tests` "spawns a child the same way and likely needs the same treatment; I have not exercised that path so I have not changed it." I have now exercised it, and it did have the same fault:

```
✅ Fetched schema: 71 methods                          <- harness's own connection, same run
FAIL — rpSpiPublicApiLoopback dispatch result: None    <- child's RPC
```

Same `None`, different surface — this path has no schema pre-check, so it fails at dispatch instead of producing the false "firmware schema does not contain..." message.

With the interface released before the spawn, the same command reaches the device and returns real data:

```
FAIL — black: {'chipset': 'APA102', 'pattern': 0, 'wireIdle': True,
               'frameBytes': 1068, 'actualClockHz': 8000000}
       'RP SPI public API loopback failed.'
```

1068 frame bytes at 8 MHz on APA102 — the RPC executed on the device. The remaining failure is physical: **GPIO11 (MOSI) -> GPIO8 (MISO)** for SPI1 is unfitted on this bench, the same gap as SPI0's GPIO3 -> GPIO0 pair. So both SPI legs of #3832's driver criterion are now blocked only on wiring, not on software.

Both phases are terminal in the dispatcher, so releasing the interface is safe in each.

Also worth recording: the first commit on this branch briefly carried a KBI001 violation because I chained `bash lint ; git commit` instead of gating on it. Amended before push; this commit was lint-gated with `&&`.

Issue #4205, opened 2026-09-07
build: link wrapper strips the exec bit from test executables on every relink, causing os error 13 with no diagnostic


zackees:
## Symptom

`bash test --cpp` fails with all tests unable to start:

```
[1/383] ✗ cled_controller FAILED
  [MESON] Test execution error: failed to spawn process: Permission denied (os error 13)
...
[MESON] ❌ Some tests failed (0/383 unit)
```

Sometimes it presents instead as a bare, reasonless failure:

```
ERROR: Unhandled python OSError. This is probably not a Meson bug, but an issue with your build environment.
[MESON] ⚠️  Test failed but could not extract detailed error information
[MESON] ❌ Tests failed with return code 13
Ok:  0    Fail:  0
```

This is the recurring symptom behind #4168. It is not stale state left by any particular merge — it reproduces on demand.

## Cause

Freshly linked **plain executables** come out without the exec bit. Shared objects are unaffected. Local umask is a normal `0022`.

After one build:
```
-rw-r--r--  1 niteris users    53920 tests/runner                      <- no +x
-rwxr-xr-x  1 niteris users    95320 tests/cycle_type.so               <- fine
-rw-r--r--  1 niteris users     6384 tests/color_profile_tiny_layout   <- no +x
```

`tests/runner` is the process the driver spawns, so when it loses the bit every single test fails to start. `color_profile_tiny_layout` is regenerated on **every** build, so it re-breaks continuously.

## Why it is hard to diagnose

Two things hide it:

1. `ci/meson/test_execution.py` streams the child's output but only echoes lines containing `FAILED` or `ERROR`, so the underlying traceback never reaches the user. You get `Ok: 0 Fail: 0` and "could not extract detailed error information". I had to bypass the wrapper to see the real error:
   ```
   PermissionError: [Errno 13] Permission denied:
     '.build/meson-quick/tests/color_profile_tiny_runtime'
   ```
2. `chmod +x` then re-running does not reliably help, because the build step inside `bash test` regenerates the affected file and strips the bit again before tests run. It only sticks if the build happens to be fully up to date.

The failure also triggers the "RE-RUN FAILED TESTS IN DEBUG MODE" banner, which is misleading here — sanitizers cannot instrument a process that will not spawn.

## Suggested fixes

[...]

Issue #3011, opened 2026-06-14
test runner produces false-pass when link fails (zccache daemon crash under all-with-examples)


zackees:
## Summary

When the `zccache` daemon dies mid-build during the `all-with-examples` alias target (245 test DLLs + 96 example DLLs linking in parallel), individual `tests/<name>.dll` link steps fail with `lost connection to daemon`. The build tool reports `FAILED` for those targets — but the downstream test runner does NOT treat the missing/stale DLL as a build error. It silently runs whatever `.dll` is on disk from a previous build, which can produce **false test failures** (and worse, **false test passes** in the opposite case).

This was discovered while debugging what looked like 9 audio test failures (`fft`, `audio_reactive`, `deficiencies`, `vibe`, `vocal`, `beat`, `audio_processor`, `detectors`, `audio`). Re-running each test in isolation after the daemon recovered showed all 9 pass. The original "failures" were assertions from stale binaries built before an unrelated `math.cpp.hpp` edit that prompted the run.

## What the log looks like

```
Detected stale lock 'build:libfastled_build' (dead PIDs: [66680]). Removing...
...
[BUILD] Linking tests...
[BUILD] [6/316] Linking target .build\meson-quick\tests\fl_audio_audio_reactive.dll
...
FAILED: [code=1] tests/fl_audio_audio_reactive.dll
"...\zccache.exe" "...\ctc-clang++.exe" -o tests/fl_audio_audio_reactive.dll ...
zccache[err][R]: lost connection to daemon (no response). The daemon may have crashed
                  or been killed mid-request — run `zccache status` to check, or
                  `zccache stop` then retry to restart it.
[7/316] Linking target tests/fl_audio_url.dll
FAILED: [code=1] tests/fl_audio_url.dll
... (same pattern for 9+ DLLs)

[MESON] Some tests failed (0/0 tests in 92.31s)
[MESON] Test failure details (9 tests): ...
   FAIL: audio::Sample - rms for known signal at ...:405
     ...:420: 500042.64 < 9000.00
```

Note `(0/0 tests in 92.31s)` — runner reports zero tests built, but then dumps 9 tests worth of stale assertion output. The disconnect is the bug.

## Why it is worse than a noisy log

1. **False-failure direction:** wastes investigation time. Stale binary fails new assertions, looks like the diff broke something it did not.
[...]

PR #4213, opened 2026-09-08
fix(build): restore mtime and exec bit on zccache-served link results (#4205, #4212)

### LINKED ISSUE #4212: bash test --cpp fails with 'Stale test artifact' on a healthy tree; --clean and zccache stop do not clear it
## Symptom

`bash test --cpp` fails all tests without running any:

```
[1/383] x cled_controller FAILED
[MESON] Stale test artifact: .build/meson-quick/tests/cled_controller.so
  DLL mtime (1788816732.21) predates this build's start (1788837138.54).
  The current link likely failed and left a previous build's DLL behind --
  running it would produce misleading results (see FastLED #3011).
  Resolve the underlying link failure (`zccache stop` + retry usually clears
  the daemon-crash variant) and re-run.
```

Ten such failures, then the runner halts early. This is #3011's guard firing. The guard is working as designed; the question is why it fires on a tree with no link failure.

## What I verified

**There is no link failure.** The build log contains no `FAILED:`, `link failed`, `lost connection`, or `ninja: build stopped`. The only match for those patterns is the runner's own `Failed: 10` summary. Linking progressed normally (`[586/824] Linking target ...`).

**Not caused by any local change.** Reproduced identically with my working changes stashed (clean tree, `git status` empty) -- same tests, same message.

**The suggested remedies do not clear it.**
- `zccache stop` -> `daemon stopped`, then the identical failure on retry.
- `bash test --clean --cpp` -> the failure *moved* from unit tests to examples (`example-AdafruitBridge`, `example-ParallelSPI`, `example-MoodRing`, ...), still 10 x "Stale test artifact".
- Across all three runs the reported DLL mtime was **unchanged** (`1788816732.21`), i.e. the artifacts are not being relinked.

**The artifacts really are old.** `cled_controller.so` and `fastled_core.so` both carry mtime `1788813865` (13:44), while the failing build started at `1788837394` (20:16) -- about 6.5 hours earlier.

**But an earlier run passed with those same artifacts.** A `bash test --cpp` at 14:33 reported `385/385 unit + examples passed` with `Compilation | 0.34s`, i.e. mostly up-to-date, against artifacts already carrying the 13:44 mtime. So "artifact older than build start" is not by itself sufficient to trigger this.

**One possibly relevant log line**, offered as an observation and not a diagnosis:

```
[BUILD] Stabilized 2 output mtimes (zccache mtime fix)
```

zccache deliberately normalises some output mtimes, which is the kind of thing that could interact with an mtime-vs-build-start check. I did not establish that it is the cause and am not claiming it.

## What I am not claiming

I could not identify the trigger. The difference between the 14:33 pass and the 20:16 failure is not accounted for by anything I measured, and I would rather leave that open than guess at it.

## Why it matters

`bash test` is unusable on this bench, and the failure presents as "your change broke 10 tests". It cost a stash/restore cycle and a full `--clean` rebuild to establish that a `src/fl/channels/bus_info.h` change was unrelated -- the first plausible reading of the output is that the change is at fault. A message distinguishing "artifact older than build start" from "link demonstrably failed during this run" would make that call much cheaper; only the second warrants suspecting the current build.

## Environment

NixOS, zccache 1.13.22, quick mode, `.build/meson-quick`. Same bench as #4205 (link wrapper stripping exec bits from test executables), which is a separate fault that also presents as broken tests.

zackees:
zccache serves a cached link by writing the bytes back out, and the restored file loses two filesystem attributes. Both were failing builds on this bench, with symptoms that look unrelated.

### mtime — #4212

`ci/meson/mtime_stabilizer.py` exists specifically to repair this. It globbed only `tests/*.dll` and `examples/*.dll`. Linux emits `.so`, macOS `.dylib` — so on those hosts it matched **zero of 383 outputs** and was a silent no-op.

`_KEY_INPUTS` had the same flaw: it looked for `fastled.dll.p/fastled.dll.symbols`, but the file is `fastled.so.p/fastled.so.symbols`, so the only input mtime it ever found was `libcrash_handler.a`.

Both are now derived from `("dll", "so", "dylib")`. Verified against the live build dir: **0 → 383** outputs matched, symbols file resolves.

### exec bit — #4205

Cached executables come back mode `0644`. The test run then dies on the first binary it spawns:

```
PermissionError: [Errno 13] Permission denied:
  .build/meson-quick/tests/color_profile_tiny_runtime
```

which reaches the user only as `ERROR: Unhandled python OSError` / `return code 13`, because the top-level handler swallows the traceback. Getting the real exception required forcing a backtrace.

This is systematic, not a fluke — a survey of the build dir found **4 ELF executables, 4 without `+x`, 0 with it**. Restored in the same post-build pass, gated on ELF/Mach-O magic so data files keep their mode.

### Verification

- `bash test --cpp` → **385/385 unit + examples passed**
- 5 new tests in `ci/tests/test_mtime_stabilizer.py`, covering all three host suffixes, the exec-bit restore, idempotence, and the negative cases (fresh links and `.so`/data files left alone)
- Confirmed the mtime tests discriminate: pre-fix config touches 0, post-fix touches 1

### What this does *not* do

It does not by itself close #4212. The stabilizer's criterion is "output older than its **input files**"; the #3011 guard's is "output older than this **build's start**". Those are different comparisons, and a cache hit can legitimately predate the build while satisfying the first. The suffix bug is real and worth fixing on its own merits; the guard's criterion is tracked on the issue.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved build artifact handling across Windows, Linux, and macOS.
  - Corrected cached executable permissions so restored binaries can run as expected.
  - Improved timestamp stabilization for shared libraries and related symbol files.
  - Preserved non-executable permissions for libraries and data files.
  - Permission restoration and artifact inspection failures now provide clearer error information.

- **Tests**
  - Added coverage for cross-platform library outputs, symbols-only updates, fresh artifacts, permission restoration, and repeatable behavior.


zackees:
Three of the four review points applied; declining one with reasoning.

**Applied**

- **chmod fail-fast** — `restore_executable_bits()` now raises with the offending path instead of continuing. Agreed: an executable we cannot chmod is one the runner dies on later with an opaque `PermissionError`, which is exactly what this function exists to prevent, so swallowing the error defeats its purpose.
- **`_build_tree` returns a `@typechecked @dataclass`** rather than a positional tuple. This is repo policy, not just style — `agents/docs/python-standards.md:11` says to prefer `@dataclass` over tuples for return types.
- **Module-level import** of `restore_executable_bits` alongside `stabilize_dll_mtimes`.

**Declined**

- **Replacing the `_DLL_PATTERNS` / `_KEY_INPUT_GLOBS` comprehensions with explicit loops.** I could not find a repo rule requiring this, and comprehensions are used throughout `ci/meson/` already (`discover_examples.py`, `discover_examples_all.py`, `native_launchers.py`, `test_discovery.py`, `example_metadata_cache.py`). Rewriting these two would make them inconsistent with their neighbours for no stated benefit. Happy to change it if there is a convention I have missed.

**Separately — running this found a real gap in my own fix.**

The hook lived only in `stream_compile_only()`. A fully-cached build never reaches that path; the log goes straight from `Core libraries: up-to-date` to running tests with no link step. So artifacts zccache had restored without the executable bit on an *earlier* run survived into the runner and failed it — the exact failure this PR is about, still reproducible on the branch that fixes it.

`restore_executable_bits()` is now also called immediately before spawning the test command. It is idempotent and costs ~1 ms, so the duplicate call on the compile path is harmless.

Reproduced and verified:

```
$ chmod -x .build/meson-quick/tests/color_profile_tiny_{layout,runtime}
$ bash test --cpp
[BUILD] ⚡ Restored 2 executable bit(s) (zccache link-cache fix)
✅ All tests passed (385/385 unit + examples in 18.53s)
```

zackees:
Applied, and it is the same principle as the chmod change already in this PR — I had only carried it half way.

`stat()` and the magic-number read both swallowed `OSError` and continued, so a file the module could not inspect was treated as one that needed nothing. That silently drops exactly the artifact this code exists to repair: the binary stays without `+x`, and the run dies later on a `PermissionError` with no connection to the cause. Skipping on an inspection failure is indistinguishable from "correctly determined there was nothing to do", which is the worst property for a repair pass to have.

Listing, `is_file()`, `stat()` and the header read now raise with the path and the underlying error. New test covers an unreadable binary:

```python
binary = _elf(tmp_path / "tests" / "runner", mode=0o000)
with pytest.raises(OSError, match="cannot inspect"):
    restore_executable_bits(tmp_path)
```

6 tests pass, and `bash test --cpp` is clean.

zackees:
Correction to my previous comment: **`bash test --cpp` is not clean on this branch, and I should not have said it was.**

The run exits 1 with 10 failures — all `Stale test artifact`, i.e. the #4212 mtime symptom that this PR explicitly does *not* fix. That limitation is stated in the PR description, so the result is expected; my summary line was simply careless.

What the run does establish:

- No `cannot inspect` errors, so the new fail-fast paths do not trip on a normal build.
- The failure count and class are unchanged from before this commit (10, all stale-artifact), so nothing regressed.
- The unit tests for this module pass: 6/6.

The accurate statement is: this branch fixes the executable-bit half (#4205) and the `.so`/`.dylib` suffix bug, and leaves the mtime-versus-build-start half of #4212 open — which is exactly what the description says and why those 10 failures are still there.

zackees:
Applied to all seven handlers in the file.

The fail-fast paths added in this PR catch `OSError` and wrap it with context, but a Ctrl+C during a `stat`, directory listing or header read would bypass the repo's interrupt registration. Every other module here pairs an `OSError` handler with an explicit `KeyboardInterrupt` clause that calls `handle_keyboard_interrupt()` and re-raises, so this file was the odd one out — and more so after I added handlers to it.

Each clause is ordered before the `OSError` one; chaining and the wrapping behaviour are unchanged. 6 unit tests pass.

For completeness on the suite: `bash test --cpp` still exits 1 with 10 `Stale test artifact` failures, which is the #4212 mtime symptom this PR does not fix — stated in the description and unchanged by this commit. No `cannot inspect` / `cannot stat` / `cannot list` errors, so the new fail-fast paths do not trip on a healthy build.

zackees:
Correction to the last paragraph of my previous comment: **this run of `bash test --cpp` passed cleanly** — exit 0, zero `Stale test artifact` failures.

I wrote "still exits 1 with 10 failures" from the previous run's behaviour instead of reading the one I had just executed. That is the second time in two days I have asserted a suite result without checking the output first, so it is a habit rather than a slip.

The accurate picture, which is more informative than what I claimed:

```
cpp exit=0
Stale test artifact count = 0
cannot inspect / cannot stat / cannot list = 0
```

The stale-artifact failure is **intermittent**, depending on whether a given build served its links from the zccache or relinked them. That is consistent with what this PR argues: cache-restored outputs come back missing filesystem attributes, so whether the #3011 guard fires depends on cache state rather than on anything in the source tree. A clean run does not mean #4212 is fixed — it means this particular build relinked enough to avoid it.

The parts that stand unchanged: the new fail-fast paths do not trip on a healthy build, and the module's 6 unit tests pass.

PR #4217, opened 2026-09-08
test(net): surface the 4 KiB payload leg in --net-peer output (#3899)

zackees:
#3899 requires "≥4 KiB payloads in both directions" with decisive evidence.

The exchange was already happening — I checked before writing anything, and deliberately did **not** add a duplicate test. `runNetClientTest` sets `success` to `passed == 12` over a fixed battery that includes a 4096-byte FNV-1a `POST /echo` round trip, and the peer loop invokes it in both directions on every cycle. Both sides implement it independently: `runHttpPayloadEchoTest` in the ESP32 block and `runRpHttpPayloadEchoTest` in the RP block, each 4096 bytes with hash verification, against the other side's `POST /echo` handler.

What was missing was *visibility*. The harness printed only the boolean, so the payload leg had to be deduced by reading firmware rather than read off the run — which is not what "decisive `REMOTE:`/`RESULT:` lines" means.

This prints the sub-test tally and the payload row per direction. Reporting only, no behaviour change.

### Result on RP2350W `2DCB876B587EA334` ↔ ESP32-C6 `8C:BF:EA:CF:87:B4`

```
--- Peer network cycle 10/10 ---
  RP2350W -> ESP32-C6: tests_passed=12 tests_failed=0
    payload: test='POST /echo 4096-byte FNV-1a' bytes=4096 passed=True
  ESP32-C6 -> RP2350W: tests_passed=12 tests_failed=0
    payload: test='POST /echo 4096-byte FNV-1a' bytes=4096 passed=True
NET PEER AUTORESEARCH PASSED (10 reconnect cycles)
```

Both directions, all ten cycles, `tests_failed=0` throughout.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation of network client-test reports.
  * Clear errors are now raised when required result data is missing, malformed, or incomplete instead of silently producing invalid output.
  * Network test summaries now consistently include matching payload results and pass/fail information.

* **Tests**
  * Expanded coverage for detailed payload results, result metadata, reconnect cycles, and invalid report scenarios.

### DIFF
--- a/ci/autoresearch/net.py
+++ b/ci/autoresearch/net.py
@@ -445,22 +445,47 @@ def _summarize_client_tests(label: str, data: dict[str, Any]) -> None:
     POST /echo round trip. #3899 asks for decisive evidence of >=4 KiB payloads
     in both directions, so print the tally and the payload row rather than
     leaving it to be deduced from the exit status.
+
+    A missing field is treated as a failure rather than printed as None: the
+    point of this function is to make the payload leg auditable, and a report
+    that cannot substantiate it is not evidence of a pass.
     """
-    passed = data.get("tests_passed")
-    failed = data.get("tests_failed")
-    print(f"  {label}: tests_passed={passed} tests_failed={failed}")
+    for field in ("tests_passed", "tests_failed"):
+        value = data.get(field)
+        if not isinstance(value, int) or isinstance(value, bool):
+            raise RpcError(
+                f"{label}: runNetClientTest response is missing an integer "
+                f"{field!r}; cannot substantiate the payload leg: {data!r}"
+            )
     results = data.get("results")
     if not isinstance(results, list):
-        return
-    for entry in results:
-        if not isinstance(entry, dict):
-            continue
-        name = str(entry.get("test", ""))
-        if "echo" in name.lower() or "byte" in name.lower():
-            print(
-                f"    {label} payload: test={name!r} "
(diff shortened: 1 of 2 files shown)


zackees:
Both applied.

**Assertions were too weak.** Correct — the test checked call counts and a zero exit, which would pass even if `_summarize_client_tests()` printed nothing. That is precisely the gap this PR exists to close, so the test was not testing the change.

It now captures stdout and asserts what a reader would actually have to read off the log:

```python
for label in ("RP2350W -> ESP32-C6", "ESP32-C6 -> RP2350W"):
    assert f"{label}: tests_passed=12 tests_failed=0" in out
    assert (f"{label} payload: test='POST /echo 4096-byte FNV-1a' "
            "bytes=4096 passed=True") in out
assert out.count("payload: test='POST /echo 4096-byte FNV-1a'") == 20
```

The count catches a summariser that fires once instead of per cycle per direction — ten cycles across two directions is twenty rows.

**Verified the assertions discriminate** rather than assuming: with both `_summarize_client_tests()` calls stubbed out the test fails; restoring them makes it pass. A test that passes whether or not the code works is worse than no test.

I added the assertions to the existing peer test rather than writing a second one — the mock setup is inline and substantial, and extracting a helper would have meant refactoring a working test to add coverage to it.

**Imports hoisted** to the module block alongside the existing ones.

6 tests pass.

PR #4220, opened 2026-09-08
fix(autoresearch): say why the RP failed to join, not just that it did

zackees:
`RP2350W did not join the ESP32-C6 AP` discards the `wifiStatus` the loop just spent ten seconds polling. That makes three very different problems indistinguishable: still associating, association rejected, or associated with no DHCP lease. Each has a different fix.

Now reports elapsed time, poll count and the last status:

```
RP2350W did not join the ESP32-C6 AP after 33.3s (60 polls);
last wifiStatus={'ip': '', 'apIp': '', 'status': 'FAILED', 'connected': False}
```

`status=FAILED` means the radio actively gave up. So this was never a timeout, and no amount of extra waiting would have fixed it — which is exactly what the old message let you assume.

Also prints join time on success. That is what showed the previous budget was marginal rather than generous: healthy joins measured **5.8, 6.2, 6.3 and 6.4 s** against a 20-poll budget worth roughly 10 s. Raised to 60 polls (~30 s); a healthy join still returns on the first or second poll, so it costs nothing when the link works.

### What I did not include, and why

I tried a 3 s settle delay after `stopNet`, on the theory that CYW43 teardown needs time before re-associating. First run looked convincing — **4 successful joins, up from 1**. Second run with the identical change: **0 joins**, failing on the very first association.

So the improvement was noise, and the settle delay is not in this PR. Mentioning it because the 1→4 result was tempting enough that I would have shipped it on a single run.

### Open, and not claimed as fixed

RP2350W ↔ ESP32-C6 association intermittently fails outright with `status=FAILED`. Observed at cycle 1, cycle 2, and not at all, across otherwise identical runs. There is also a separate intermittent HTTP failure with the RP acting as client (`body_read: 0`, empty status line). Neither is addressed here — this PR only makes them legible. Worth their own investigation with the better diagnostics now in place.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Wi-Fi connection reliability by allowing up to 60 polling attempts while preserving time for diagnostics and cleanup.
  * Enhanced failure diagnostics with elapsed join time, polling count, early-stop status, and final Wi-Fi status.
  * Improved successful connection reporting with the join duration and assigned IP address.
  * Improved peer network test reporting by validating test results and displaying pass/fail totals and payload evidence.
  * Added clearer errors when network test responses are incomplete or lack payload results.

### DIFF
--- a/ci/autoresearch/net.py
+++ b/ci/autoresearch/net.py
@@ -437,6 +437,13 @@ async def run_net_loopback_autoresearch(
             await client.close()
 
 
+# CYW43 association is not instant, and a re-join after stopNet has been
+# observed to take longer than the original 10 s budget allowed. Poll for up
+# to ~30 s before calling it a failure; a healthy join still returns on the
+# first or second poll, so this costs nothing when things are working.
+kJoinPollAttempts = 60
+
+
 async def run_net_peer_autoresearch(
     upload_port: str,
     peer_upload_port: str,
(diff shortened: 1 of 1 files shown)


zackees:
Valid, and it undercuts the point of this PR specifically.

Raising the join budget to 60 polls (~30 s) means a run whose joins are slow can spend the entire `--timeout` inside the polling loop. The first casualty is the failure path — so the `wifiStatus` report this PR exists to add would be lost precisely when it matters, and the run would end with the same opaque timeout it started with.

The loop now stops polling once the whole-run deadline is within `kJoinReserveSeconds` (20 s), leaving room for the diagnostic RPC and the teardown pings:

```python
for _ in range(kJoinPollAttempts):
    if deadline - time.monotonic() < kJoinReserveSeconds:
        break
    wifi_status = await rpc_data(primary, "wifiStatus")
    ...
```

I chose reserving a slice over deriving the timeout from the full cycle budget because the derived figure would be large and misleading — 10 cycles x 30 s of join budget alone is 300 s, which is not a realistic run time, it is a worst case that never occurs when the link is healthy (observed joins take 5.8-6.4 s). Reserving keeps the common case unchanged and only bites when joins are genuinely pathological, which is the case that needs the diagnostics.

zackees:
Valid, and it matters more here than it would elsewhere — this PR exists specifically to make a failure legible, so a number in the failure message overstating the evidence undercuts the whole change.

The message quoted `kJoinPollAttempts` (60), but the deadline reserve I added in the previous commit can end the loop early. A run that managed twelve polls would have claimed sixty.

Now counts the requests actually issued, and says when the reserve cut the sweep short:

```
RP2350W did not join the ESP32-C6 AP after 33.2s across 60 wifiStatus poll(s);
last wifiStatus={'status': 'FAILED', ...}
```

versus

```
... after 8.1s across 14 wifiStatus poll(s) (stopped early to reserve deadline
for this report); last wifiStatus={...}
```

Those are different situations — "swept the full budget and it never associated" versus "ran out of run time before finishing the sweep" — and a reader should not have to guess which one they are looking at.

PR #4221, opened 2026-09-08
fix(autoresearch): stall-based timeouts for the RP HTTP client path

zackees:
Three deadline corrections in the RP networking path, all the same shape as the server-side fix merged in #4219.

### 1. `runRpHttpRequestTest` — one flat deadline for the whole exchange

```cpp
const uint32_t deadline_ms = millis() + 2000;
```

Set once, then used for the first-byte wait, the status line, every header *and* the body. #4196 fixed the body drain to read to completion, but this overarching budget survived — so a peer merely slow to start failed the entire request. It surfaces as:

```
'error': 'Unexpected HTTP status', 'body_read': 0,
'status_line': '', 'content_length': -1
```

which reads as a transport fault rather than a timeout. Now measures a stall refreshed by every byte read, with a 4 s ceiling.

### 2. `runRpHttpPayloadEchoTest` — same shape

A flat 4 s for an entire 4 KiB round trip, reporting its timeout as a hash mismatch against `2166136261` — the FNV-1a offset basis, i.e. the hash of zero bytes. Same stall treatment.

### 3. The #4219 server timer charged the peer for unserviced time

My own fix, incomplete. The stall timer measures wall time, but `pollNetServer()` is called once per `loop()` **after** `fl::task::run()` drains the RPC queue. A blocking handler — `runNetClientTest` issues 12 HTTP requests — starves it for seconds, and on the next poll `millis() - last_progress_ms` is large, so a connection that never went quiet is answered 408.

Service gaps over 100 ms are now credited back to both the stall watermark and the request budget: the peer cannot be judged silent over a window nobody was reading.

### Honest limit on verification

**None of these is demonstrated to fix the intermittent peer-link failures**, and I am not claiming they do.

Across runs today the failure moved between association (`status=FAILED`), RP-as-client reads (`body_read: 0`), the 4 KiB echo, and — after fix 3 — a 408 on a small `POST /rpc unknown`. No configuration reaches a clean 10/10, including unmodified master. Each change here is justified by inspection and by the precedent already accepted in #4219, so they are landing as correctness fixes rather than as a claimed repair.

### New open finding

After fix 3 the server still answered 408 to a *small* POST. That is worth noting separately: #4219 made the server *report* timeouts instead of closing silently, so requests that previously failed mutely now surface as 408 — meaning there may be a pre-existing "server does not complete this request" problem that the timeout change merely exposed. Not chased here; flagging it with the observation rather than guessing.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved RP2350W HTTP reliability by distinguishing periods of inactivity from overall request duration.
  - Slow or fragmented exchanges now continue while data is being received, with inactivity timing refreshed for each received byte.
  - HTTP operations now stop after a maximum exchange duration, even when intermittent data is received.
  - Improved timeout handling for large payload echo requests and outbound connections.

### DIFF
--- a/examples/AutoResearch/AutoResearchNet.cpp
+++ b/examples/AutoResearch/AutoResearchNet.cpp
@@ -769,9 +769,6 @@ namespace {
 // A stalled peer is one that has sent nothing for this long. Kept at the
 // old flat budget so a truly dead connection is reaped just as quickly.
 constexpr uint32_t kRpPeerStallMs = 2000;
-// A gap longer than this between service calls means loop() was busy
-// elsewhere, not that the peer stopped sending.
-constexpr int32_t kRpPeerServiceGapMs = 100;
 // Absolute ceiling for one request, so a peer that dribbles a byte at a time
 // cannot occupy the single-client server forever.
 constexpr uint32_t kRpPeerMaxRequestMs = 10000;
@@ -791,11 +788,9 @@ struct RpPeerState {
     // total elapsed time: a 4096-byte POST body arrives across several TCP
     // segments and legitimately outruns a fixed budget on CYW43 SoftAP.
     uint32_t last_progress_ms = 0;
-    // When pollNetServer() was last given a chance to run. loop() only calls
     // it after draining the RPC queue, so a blocking handler can starve it
     // for seconds. Without this the stall timer charges that gap to the peer
     // and answers 408 to a connection that never went quiet.
-    uint32_t last_service_ms = 0;
 };
 
 RpPeerState& rpPeerState() {
@@ -810,7 +805,6 @@ void resetRpPeerRequest(RpPeerState& state) {
     state.response_complete = false;
     state.request_started_ms = 0;
     state.last_progress_ms = 0;
-    state.last_service_ms = 0;
     state.request[0] = '\0';
     state.body[0] = '\0';
 }
@@ -1285,7 +1279,6 @@ void pollNetServer() {
         resetRpPeerRequest(state);
         state.request_started_ms = millis();
         state.last_progress_ms = state.request_started_ms;
-        state.last_service_ms = state.request_started_ms;
     }
     if (!state.client) {
         return;
@@ -1303,20 +1296,6 @@ void pollNetServer() {
     // Answering 408 instead of closing mutely means a real timeout is
     // reported as one rather than as a transport error.
     const uint32_t now_ms = millis();
-    // Credit an unserviced interval to the *stall* watermark only: the peer
-    // cannot be judged silent over a window nobody was reading. The absolute
-    // budget is deliberately not credited -- advancing it on every poll gap
-    // would push the ceiling forward indefinitely, so a peer that never sends
-    // anything could hold this single-client server for good, which is the
-    // exact case that ceiling exists to bound.
-    if (state.last_service_ms != 0) {
-        const int32_t service_gap =
-            static_cast<int32_t>(now_ms - state.last_service_ms);
-        if (service_gap > kRpPeerServiceGapMs) {
-            state.last_progress_ms += static_cast<uint32_t>(service_gap);
-        }
-    }
-    state.last_service_ms = now_ms;
     const bool stalled =
         static_cast<int32_t>(now_ms - state.last_progress_ms) >= kRpPeerStallMs;
     const bool over_budget =
--- a/examples/AutoResearch/AutoResearchNet.cpp
+++ b/examples/AutoResearch/AutoResearchNet.cpp
@@ -769,6 +769,9 @@ namespace {
 // A stalled peer is one that has sent nothing for this long. Kept at the
 // old flat budget so a truly dead connection is reaped just as quickly.
 constexpr uint32_t kRpPeerStallMs = 2000;
+// A gap longer than this between service calls means loop() was busy
+// elsewhere, not that the peer stopped sending.
+constexpr int32_t kRpPeerServiceGapMs = 100;
 // Absolute ceiling for one request, so a peer that dribbles a byte at a time
 // cannot occupy the single-client server forever.
 constexpr uint32_t kRpPeerMaxRequestMs = 10000;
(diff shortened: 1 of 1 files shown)


zackees:
Both points valid and applied.

**Deadline armed after `connect()`.** Correct, and it defeats the purpose of the change: `WiFiClient::connect()` is precisely where an unreachable or slow peer stalls, and arming the clock afterwards excluded that from the budget. Both request paths now take their start timestamp before `connect()`:

```cpp
const uint32_t exchange_started_ms = millis();   // before connect()
uint32_t last_progress_ms = exchange_started_ms;
if (!client.connect(host_ip, port)) { ... }
...
const uint32_t hard_deadline_ms = exchange_started_ms + kRpHttpClientMaxMs;
```

Same shape in `runRpHttpPayloadEchoTest`.

**Inner drain loop.** Also correct. `while (client.available())` never re-checked expiry, so a peer streaming without pause would keep it running past the hard deadline — the outer check is only reached when `available()` goes false. It now breaks on `echo_expired()`.

While applying this I briefly duplicated the echo declarations into `runRpHttpRequestTest` via a careless scripted edit; caught by the compiler and cleaned up before pushing.

**Verification is partial, and I am not claiming otherwise.** The peer link on this bench is intermittently failing for unrelated reasons. The run after this change failed at cycle 1 in the **C6 → RP** direction:

```
POST /echo 4096-byte FNV-1a  passed=False  status_code=408
```

That is the RP acting as *server* answering 408 — the server-side path, which these client-side changes do not touch. The RP → C6 client requests ran without issue. So the changes are exercised but not proven by a clean 10/10, which this bench is not currently producing in any configuration.

zackees:
Correct, and this was a bug I introduced in the previous commit rather than a pre-existing weakness.

Crediting an unserviced poll gap to `request_started_ms` as well as to the stall watermark pushes the absolute ceiling forward on **every** gap. A peer that sends nothing at all therefore keeps extending its own deadline and can hold the single-client server indefinitely — exactly the slow-loris case `kRpPeerMaxRequestMs` exists to bound. I defeated my own guard while trying to make the stall timer fair.

Only the stall watermark is credited now:

```cpp
if (service_gap > kRpPeerServiceGapMs) {
    state.last_progress_ms += static_cast<uint32_t>(service_gap);
    // request_started_ms deliberately NOT credited
}
```

That keeps the property I wanted — a peer is not judged silent over a window nobody was reading — while the wall-clock budget stays a real bound.

Verification is still partial and I am not claiming otherwise: the peer link on this bench is intermittently failing for unrelated reasons, and the run after this change failed again in the **C6 → RP** direction, which is the server path these client-side deadlines do not touch.

zackees:
**I have removed fix 3 (the server-side service-gap credit) from this PR — it was a regression.**

The A/B leaves no room for interpretation:

| configuration | result |
|---|---|
| **master** (change absent) | 5 cycles, **no 408 responses at all** |
| credit on stall + budget | cycle 1, server 408 on `POST /echo` |
| credit on stall only (absolute ceiling) | cycle 1, server 408 on `POST /rpc discover` |
| serviced-time ceiling | cycle 1, server 408 on `POST /rpc unknown` |
| **change removed** | **10/10 cycles, zero 408s, exit 0** |

Three formulations, each moving the failing request to a different one without fixing it, against a master that does not exhibit the failure. And the final run — with the change gone — is the first clean 10/10 this bench has produced in many attempts.

I should have run that master comparison after the first 408 rather than after the third. I had been attributing those failures to "the known intermittency" because the peer link genuinely is flaky, which made a real regression easy to wave through. The failing sub-test moving to a *smaller* POST after my "fix" was the signal I under-weighted.

**What remains in this PR** — the two client-side deadline fixes, which are what it was for:

- the exchange clock starts before `connect()`, so a stalled connect is inside the budget it is meant to bound
- the echo drain loop honours the hard deadline instead of running while `available()` stays true

**The starvation concern is real and now unaddressed.** `pollNetServer()` is still called once per `loop()` after the RPC queue drains, so a blocking handler can still starve it. But fixing that needs to start from evidence about which budget actually fires — the `netServerStats` diagnostic in #4222 reports `lastTimeoutWasStall`, `lastTimeoutBodyLength` and `lastTimeoutElapsedMs` precisely for this. That should land first.

zackees:
### Attempted to demonstrate these fixes work. The evidence does not support it.

Recording this because the cross-build numbers looked convincing and I nearly reported them as proof.

**The apparent result.** In a window where the peer link was healthy (association succeeding throughout, so it was no longer the confound it had been earlier), five runs:

| build | runs | outcome |
|---|---|---|
| master | 2 | **both failed at cycle 9** — `GET /status` and `GET /leds`, each `body_read: 0` |
| this branch | 3 | **all 10/10 PASS** |

Both master failures are exactly the symptom this PR addresses: the RP acting as client, getting nothing back, with an empty status line. Mechanism matched, conditions comparable, 0/2 against 3/3.

**The isolating test.** Rather than trust a cross-build count, I reverted *only* the client deadline logic back to master's flat 2 s total-elapsed form, on this branch, changing nothing else — the same technique that settled the join-budget question on #3899:

```
flat-deadline run1: exit=0  cycles=10
flat-deadline run2: exit=0  cycles=10
```

**Both still pass.** If the fix were what separated master from this branch, restoring master's behaviour should have restored master's failure. It did not.

**Conclusion.** The 0/2 versus 3/3 split is not attributable to these changes. The peer link's intermittency varies on a scale of hours — established on #3899, where five consecutive clean runs happened at two different poll budgets — and two master runs is simply too small a sample against that background.

So this PR remains **exercised, not proven**, which is what its description says. The two client-side changes are still correct by inspection: a deadline armed after `connect()` cannot bound time spent inside `connect()`, and a drain loop that never re-checks expiry can outrun its own budget. Neither claim needs the runs.

This is the third hypothesis I have tested and refuted on this bench with a within-build A/B — after the `stopNet` settle delay and the join budget. In each case a small cross-build sample looked like causation and an in-build revert showed it was not. Worth stating plainly: on this fixture, comparisons across builds are close to worthless without one.

zackees:
CI failure was `QEMU ESP32C3 test-emu failed: timed out` — an emulator timeout unrelated to this branch.

Verified rather than assumed: both changes here sit at lines 862 and 1054 of `AutoResearchNet.cpp`, inside the `#elif defined(FL_IS_RP2350)` block that begins at line 748. The ESP32 branch spans lines 50-748, so an ESP32C3 build never compiles this code.

Branch refreshed against master (merged cleanly, no conflicts) and re-verified on hardware: **10/10 cycles, `NET PEER AUTORESEARCH PASSED`**.

Issue #4216, opened 2026-09-08
s16x16x4 packed type is ~15x slower than scalar on the simd_noop fallback (RP2350/Cortex-M33)


zackees:
## Measurement

`testSimdBenchmark` on RP2350W `2DCB876B587EA334`, 10000 iterations, times in µs (lower is better):

| op | float | s16x16 | u16x16 | s8x8 | **s16x16x4 (packed)** |
|---|---|---|---|---|---|
| mul | 738 | 984 | 980 | 891 | **11568** |
| sub | 653 | 570 | 568 | 821 | **10434** |

Reproduce with `bash autoresearch rp2350w --simd`.

## The comparison is like-for-like

Every helper in `examples/AutoResearch/AutoResearchSimd.h` processes **four lanes per iteration**, so this is not a 1-lane-vs-4-lane artefact:

- `benchFloat4` — four scalar `float` ops
- `benchS16x16_4` — four scalar `fl::s16x16` ops
- `benchSimd4` — one packed `fl::s16x16x4` op

Same work, ~15× the time.

## Why this is a bug rather than "no SIMD on this chip"

RP2350 is Cortex-M33 without `__ARM_FEATURE_DSP`, so `src/platforms/simd.h` falls through to `platforms/shared/simd_noop.hpp`:

```
#if   x86/x64                -> simd_x86.hpp
#elif FL_IS_ESP32            -> simd_esp32.hpp
#elif __ARM_FEATURE_DSP == 1 -> simd_arm_dsp.hpp
#else                        -> simd_noop.hpp   // "No SIMD support - use scalar fallback"
```

A scalar *fallback* should cost approximately what the equivalent scalar code costs. Failing to accelerate is expected and fine. Being **15× slower than the code it falls back to** is not — it means reaching for the portable vector type silently costs an order of magnitude on every platform that lands in this branch.

That branch is not RP-specific: it is the default for every ARM part without the DSP extension, and for any platform not otherwise matched.

## Hypothesis — not verified

The fallback vector is:

```cpp
struct FL_ALIGNAS(16) simd_u32x4 { u32 data[4]; };
```

16 bytes. Under AAPCS a 16-byte aggregate is not returned in registers, so each operator likely returns through memory, and a dependent chain like the benchmark's (`a = op(a, b); b = a + bump;`) would round-trip the stack every iteration.

[...]

PR #4223, opened 2026-09-08
diag(net): discriminate transient PCB exhaustion from a leak on listen failure

zackees:
`startNetServer` reported `TCP server failed to listen` at cycle 8 of a peer run, with no indication why. Rather than extend the retry window and hope, I read the implementation.

Arduino-Pico's `WiFiServer::begin()` (`rp2040-5.7.0/libraries/WiFi/src/WiFiServer.cpp:52`) rules out the two obvious guesses:

```cpp
void WiFiServer::begin(uint16_t port, uint8_t backlog) {
    close();                                   // fresh start each call
    tcp_pcb* pcb = tcp_new();
    if (!pcb) return;                          // <-- pool exhausted
    pcb->so_options |= SOF_REUSEADDR;          // TIME_WAIT will not block bind
    if (tcp_bind(pcb, _addr, _port) != ERR_OK) { tcp_close(pcb); return; }
    ...
}
```

- Repeated `begin()` on the same object is **safe** — it closes and rebuilds the PCB, so a stale object is not the problem.
- `SOF_REUSEADDR` is set before `tcp_bind`, so a bind blocked by `TIME_WAIT` is not it either.

That leaves `tcp_new()` returning NULL with the lwIP PCB pool exhausted — plausible at roughly 24 connections per cycle, every request carrying `Connection: close`.

**That hypothesis is testable rather than arguable.** Pool exhaustion drains as `TIME_WAIT` entries expire; a leak does not. On failure the harness now waits 30 s, retries once, and reports which it saw:

```
startNetServer recovered after the drain -> transient resource exhaustion, not a leak
```

or, if it persists, an error saying so explicitly.

### Honest status

The retry **did not fire across three runs**, so this path is unexercised. Landing it because the failure is infrequent enough that the next occurrence should yield an answer rather than another opaque message.

### A pattern I nearly reported, and shouldn't have

Two consecutive runs failed at exactly cycle 8 — one on this listen failure, one on WiFi association. That looked like cumulative resource exhaustion with whichever operation came next taking the hit.

The third run reached **10/10 and passed** with identical code. So there is no cycle-8 limit; per-cycle failures are probabilistic and two-at-eight was coincidence. Recording it because the two-point "pattern" was convincing enough that I was about to write it up as one.

That run is also the first clean `NET PEER AUTORESEARCH PASSED (10 reconnect cycles)` in a while, which is worth knowing on its own — the intermittency is not a permanent regression.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network server startup resilience for RP2350W peer connections by waiting briefly and retrying once after a startup failure.
  * Startup now respects the caller’s deadline and reports an error when there is insufficient time to retry or when the retry also fails.
  * Improved HTTP test result validation by reporting sub-test counts and payload-test results, with clear errors when required result details or payload evidence are missing.


zackees:
Both points applied. The second is a correction to my reasoning, not just the code, so worth spelling out.

**Drain ignored the caller's deadline.** Fixed — the wait is capped against the remaining budget with a reserve for reporting the outcome, and refuses up front when too little remains rather than silently overrunning `--timeout`:

```python
remaining = deadline - time.monotonic()
drain_s = min(kListenRetryDelayS, remaining - kListenReserveSeconds)
if drain_s <= 0:
    raise RpcError(... "Not enough of the run deadline remains to test whether it recovers.")
```

Otherwise a run near its deadline would trade the diagnostic for a timeout and report nothing — the same failure mode as the join budget in #4220.

**Inferring a PCB leak from the retry outcome.** You are right, and this was an overclaim on my part. The code printed *"transient resource exhaustion, not a leak"* on recovery and *"points at a PCB leak"* on persistence. Recovery establishes only that the condition was **transient** — it says nothing about which resource ran out.

`WiFiServer::begin()` returns a bare bool, so nothing host-side can distinguish `tcp_new()` exhaustion from a bind failure or an allocation failure. My reasoning was: TIME_WAIT is ruled out by `SOF_REUSEADDR`, therefore PCB exhaustion — but "the obvious candidate is eliminated" does not make the next candidate proven.

The messages now report transient versus persistent, which is what the experiment actually establishes, and the comment records PCB exhaustion as the leading candidate that would need a firmware-side error code to confirm.

This is the same mistake I made on #4216, where I named a cause from arithmetic and the disassembly later showed I was wrong. Naming a mechanism is cheap; establishing one is not.

zackees:
CI failure was `QEMU ESP32C3 test-emu failed: timed out` — an emulator timeout, not a defect in this branch. This PR touches only `ci/autoresearch/net.py`, which the ESP32C3 QEMU job does not execute.

Branch refreshed against master (which has since merged #4217 and #4225). That required resolving one conflict in `net.py`: my listen-retry constants landed at the same place as #4217's `_summarize_client_tests`. Both are additive and unrelated, so both are kept.

Merged rather than rebased, deliberately — rebasing cost me a dropped merge commit on #4214 earlier in this work.

Post-merge verification on RP2350W: the run reached cycle 4 before the known intermittent peer failure, with 6 payload rows printed, confirming #4217's summariser survives the merge intact. Given this link's variance (documented on #3899 — five consecutive clean runs earlier today, failures at cycles 1-9 at other times), a single run is not evidence either way.

zackees:

PR #4225, opened 2026-09-08
fix(autoresearch): stop under-sizing the RP PIO RX capture request

zackees:
Stacked on #4224 (which adds the 400 kHz chipset that made this observable).

### Root cause

The RP capture request assumes a clean frame:

```cpp
constexpr size_t kPioPhasesPerDataByte = 16;   // 8 bits x (high + low)
rx_config.edge_capacity = expected_data_bytes * kPioPhasesPerDataByte + 1u;
```

Exact for 800 kHz, too small for slower chipsets — their captures produce more entries than one per phase. The capture then hits `RxWaitResult::BUFFER_OVERFLOW` and the buffer is discarded, surfacing as `decode_mismatch` with zero decoded bytes.

That overflow is what the cryptic `RX wait/raw: 2/256` was reporting all along: `RxWaitResult` is `SUCCESS = 0, TIMEOUT = 1, BUFFER_OVERFLOW = 2`. The `256` is unrelated — it is the diagnostic dump window.

### Fix

Request 2× the phases, then **clamp to `kRpPioRxEdgeCapacity`**.

The clamp is load-bearing, not defensive tidiness. That constant sizes `RpPioRxEdgeStorage = FixedVector<EdgeTime, kRpPioRxEdgeCapacity>`, the static container the device appends into — so requesting more than it holds authorises `appendDuration()` to write past the end. I confirmed that by doing it: an unclamped request fails in 17 ms with nothing captured, versus 684 ms for a healthy run.

### Measured on RP2350W `2DCB876B587EA334`

| case | before | after |
|---|---|---|
| 800 kHz, 100 LEDs | PASS 684 ms | **PASS 683 ms** — no regression |
| 400 kHz, 40 LEDs | FAIL `decode_mismatch` | **PASS 518 ms** — new capability |
| 400 kHz, 100 LEDs | FAIL | FAIL — clamped at the pool |

### What this deliberately does not do

The 100-LED 400 kHz case stays blocked, because `kRpPioRxEdgeCapacity` is `100 * 3 * 16 + 1` — exactly one clean 800 kHz frame, with a single spare slot that `finishTailIfIdle()` consumes with its own tail append. There is no headroom by construction.

Doubling that pool **does** fix it — verified, 400 kHz/100 LEDs passes at 1222 ms with 800 kHz unaffected. But it costs:

```
pool entries   4801 -> 9601
words array    18.8 KB -> 37.5 KB
edges array    18.8 KB -> 37.5 KB
delta          +37.5 KB static RAM, unconditional
               = 14.2% of an RP2040's 264 KB
```

`gPioRxCaptureBuffers` is a static global, allocated whether or not any RX capture is used, so that cost lands on every RP build. That is a real trade-off and I am not taking it unilaterally in a fix aimed at a test harness — flagging it with the measurement instead.

### Note on the include

`AutoResearchTest.cpp` now includes `rx_pio_channel.h`, which is marked `IWYU pragma: private`, inside the existing RP-only guard. It is needed for the clamp bound. If that is unwelcome, the alternative is exposing the capacity through the RX channel interface rather than duplicating the constant — I did not want to duplicate a magic number that must track the pool.

## Summary by CodeRabbit

- **Bug Fixes**
  - Legacy mode now reports an error when used with chipsets that do not support legacy timing templates, instead of silently falling back to WS2812B timing.
  - RP-based signal capture now validates sizing and allocates buffer capacity more reliably, reducing the risk of insufficient capture space or invalid sizing.

- **Tests**
  - Added coverage for supported and unsupported chipset combinations in legacy and non-legacy modes.
  - Added validation for RP-based capture capacity handling.

### DIFF
--- a/ci/autoresearch/phases.py
+++ b/ci/autoresearch/phases.py
@@ -824,6 +824,21 @@ def _parse_args_and_build_commands(args: Args) -> RunContext | int:
         print(f"\u274c Error: {flag} requires --legacy")
         return 1
 
+    # The legacy path resolves timing from a LegacyClocklessChipset template
+    # and ignores timing_name entirely (AutoResearchRemoteRunSingleTest.cpp),
+    # so a chipset with no legacy template silently runs as WS2812B rather
+    # than failing. Reject the combination instead of reporting a pass for
+    # timing that was never applied.
+    legacy_capable_chipsets = {"ws2812", "ws2814", "ws2818"}
+    if args.legacy and args.chipset not in legacy_capable_chipsets:
+        print(
+            f"\u274c Error: --chipset {args.chipset} has no legacy template, so "
+            "--legacy would silently fall back to WS2812B timing. Use "
+            f"--chipset {args.chipset} without --legacy, or pick one of: "
+            + ", ".join(sorted(legacy_capable_chipsets))
+        )
+        return 1
+
     explicit_legacy_chipsets = {
         "ws2814": "WS2814",
         "ws2818": "WS2818",
--- a/ci/tests/test_autoresearch_phases.py
+++ b/ci/tests/test_autoresearch_phases.py
@@ -9,6 +9,8 @@ Tests the refactored phase decomposition:
 """
 
 import asyncio
+import contextlib
+import io
 import os
 from pathlib import Path
 from unittest.mock import AsyncMock, MagicMock, patch
@@ -3026,9 +3028,22 @@ class TestRunTestsOrSpecialMode:
                 chipset=chipset,
                 flex_io=True,
                 parlio=False,
+                environment_positional="teensy40",
                 project_dir=fake_project_dir,
+                use_root_platformio_ini=False,
             )
-            assert _parse_args_and_build_commands(args) == 1, chipset
+            with patch(
+                "ci.autoresearch.staging.synthesise_autoresearch_project",
+                return_value=fake_project_dir,
+            ):
+                buffer = io.StringIO()
+                with contextlib.redirect_stdout(buffer):
+                    result = _parse_args_and_build_commands(args)
+            assert result == 1, chipset
+            # Assert on the reason, not just the code. Without this the test
+            # passed on the unrelated --use-root-platformio-ini rejection and
+            # never reached the guard it was written for.
+            assert "has no legacy template" in buffer.getvalue(), chipset
 
     def test_non_legacy_still_accepts_those_chipsets(
         self, fake_project_dir: Path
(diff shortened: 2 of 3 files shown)


zackees:
Both points valid and applied.

**Overflow guard.** Correct — I added the `* 2` after the existing check, so the guard no longer covered the full expression. An oversized `expected_data_bytes` could pass it, wrap during the widened multiply, and produce a small `requested` that gets past the clamp into `RpPioRxDevice::begin()`. The guard now divides by the combined factor:

```cpp
constexpr size_t kPioPhasesPerDataByteWithHeadroom =
    kPioPhasesPerDataByte * kPioPhaseHeadroomNumerator;
if (expected_data_bytes > (size_t(-1) - 1u) / kPioPhasesPerDataByteWithHeadroom) { ... }
...
size_t requested = expected_data_bytes * kPioPhasesPerDataByteWithHeadroom + 1u;
```

Both sites use the same constant now, so the guard and the request cannot drift apart.

**`--legacy` with an unsupported chipset.** Also correct, and it is a trap I introduced by adding `ws2811-400`. The legacy path resolves timing from a `LegacyClocklessChipset` template and ignores `timing_name` entirely, so `--legacy --chipset ws2811-400` would run WS2812B timing and report a **pass for timing that was never applied** — the worst kind of failure, since it looks like success.

`ucs7604` had the same hole already. Both are now rejected with a message naming the chipsets that do have legacy templates. Note the converse check (ws2814/ws2818 *requiring* `--legacy`) already existed; this is the missing direction.

Tests: 179 pass, including the two rejection cases and a positive case proving the guard does not block those chipsets outside legacy mode. My first attempt at that positive test failed for an unrelated reason — I had not set `use_root_platformio_ini=False`, which the sibling tests do — so it was the test, not the guard.

Re-verified on RP2350W `2DCB876B587EA334`: 800 kHz/100 LEDs `PASS 684ms`, 400 kHz/40 LEDs `PASS 518ms`.

zackees:
Good catch, and worse than it looks — the test was passing for an entirely unrelated reason.

It left `use_root_platformio_ini` at its default `True`, so the run returned 1 from the Teensy/root-platformio check and never reached the legacy-chipset validation at all. Asserting only on the exit code hid that completely. Confirmed by capturing stdout:

```
rc = 1
reason: ❌ Error: --use-root-platformio-ini is not allowed for Teensy AutoResearch
        acceptance. Use the synthesized fbuild project from ci/boards.py instead.
```

Not my guard's message. The test asserted the right number for the wrong reason, which is the same failure mode as the too-weak assertions on #4217 — a test that would pass with the feature deleted.

Now configured to reach the validation, and asserting on the **reason** rather than the code:

```python
assert result == 1, chipset
assert "has no legacy template" in buffer.getvalue(), chipset
```

Verified it discriminates: disabling the guard makes the test fail; restoring it passes. 179 tests green.

PR #4227, opened 2026-09-09
feat(channels): expose the wire timing the UART encoder actually produces

zackees:
Groundwork for the last unfixed UART defect on #3899. This PR adds only the encoder-side helper; it does not yet change any decoder.

### The problem it addresses

The UART encoder quantises T0H/T1H onto a grid of `period / pulses_per_bit`, so what reaches the wire is not the chipset's nominal timing. A receiver decoding a UART-encoded frame must classify against the **quantised** values.

`AutoResearchTest.cpp` instead hardcodes one chipset's worth of them:

```cpp
fl::ChipsetTiming uart_timing{250, 500, 500, 50, "UART_4Mbps"};
```

Measured consequence on RP2350W at 400 kHz (the only rate the RP PL011 can clock — see #4224):

| decode timing | LEDs wrong |
|---|---|
| hardcoded `{250, 500, 500}` | **40/40** |
| derived from the chipset | **1/40** |

### What this adds

```cpp
ChipsetTiming uartWireTiming(const ChipsetTimingConfig& timing,
                             u32 max_baud_rate) FL_NO_EXCEPT;
```

It runs the same `fitUartWave()` the LUT was built from, so decode windows cannot drift from the encoding. Recomputing the rounding at the decode site would duplicate that logic — which the encoder's own comment warns against: *"Shared by buildWave10Lut() and canRepresentTiming() so the LUT that gets built is always judged by the same rules that admitted it."*

### Correcting a claim I made on #3899

I asserted that deriving reproduces the hardcoded constants **exactly** for WS2812, and used that as the argument the change is risk-free on platforms I cannot test. **That was wrong**, and the test caught it:

```
WS2812B-V5 is T1=225 T2=355 T3=645  ->  period 1225 ns, not 1250
  derived   {245, 490, 490}
  hardcoded {250, 500, 500}
  deltas     5 ns, 10 ns, 10 ns   against a +/-250 ns tolerance
```

The constants correspond to a nominal 1250 ns part; V5 is 1225 ns. I had computed the "proof" from an assumed period instead of reading the actual timing macros.

The safety argument survives in weaker form — 5-10 ns against a ±250 ns classification tolerance cannot move any symbol across a boundary — but "well inside tolerance" is not "identical", and I had asserted the stronger claim. The test now pins the real derived values so a future timing change cannot drift silently.

### Scope

Encoder-side only. Switching the AutoResearch decoder over is a separate change, deliberately, so the helper and its ESP32-visible consequences can be reviewed on their own.

Tests cover the derived values exactly, the tolerance property that makes substitution safe, and infeasible-ceiling handling.

## Summary by CodeRabbit

* **New Features**
  * Added an API to report the actual quantized UART wire timing for supported chipset configurations.
  * Reports when timing is infeasible under the selected baud-rate limit instead of providing an estimated result.
  * Timing results reflect pulse-grid alignment and chipset-specific encoding constraints.

* **Tests**
  * Added coverage for chipset-specific timing, pulse-grid alignment, decoder tolerance, and infeasible baud-rate constraints.


zackees:
Two CI failures, one real and now fixed, one infrastructure.

**Bloat regression — real, and mine.**

```
esp32s3-bloat-regression: FAIL — total_flash=342,407 B exceeds baseline=342,393 B by 14 B
```

14 bytes, from this PR's helper. The gate's message asks for the baseline to move in the same PR when the growth is intended, so `tests/data/esp32s3_bloat_baseline.txt` is updated to 342407.

Worth being explicit about what is being paid for: **this PR alone adds a function nothing calls yet.** The caller is #4228, which turns UART loopback on RP from 40/40 LEDs wrong into a pass. If #4227 landed without #4228 following, those 14 bytes would buy nothing.

I looked at avoiding them rather than just moving the number. The helper cannot be header-inlined so the linker drops it when unused — it calls `fitUartWave()`, which is internal to the encoder translation unit, and sharing exactly those rules is the entire point of the change. Reimplementing the rounding in the header to save 14 B would reintroduce the drift this exists to prevent.

**`build / build` — infrastructure.**

```
error: package error: failed to download
       https://www.teknic.com/files/downloads/ClearCore-1.7.4.zip:
       error sending request
```

An external package fetch, nothing to do with this diff. Re-running it.

PR #4228, opened 2026-09-09
fix(autoresearch): derive the UART decode timing — UART loopback now passes on RP

zackees:
Stacked on #4227, which adds the `uartWireTiming()` helper this uses.

### Result

```
before   TEST UART0 lanes=1 leds=40 FAIL 533ms   40/40 LEDs wrong
after    TEST UART0 lanes=1 leds=40 PASS 527ms
```

Three consecutive passes on RP2350W `2DCB876B587EA334`. PIO loopback unaffected (`PASS 685ms`).

This closes the UART clockless leg of #3899, which had been recorded as unreachable.

### The change

The decoder hardcoded one chipset's quantised wire timing:

```cpp
fl::ChipsetTiming uart_timing{250, 500, 500, 50, "UART_4Mbps"};
```

Those describe WS2812 at a nominal 1250 ns. At 400 kHz — the only rate an RP PL011 can clock (#4224) — they misclassify every symbol. It now derives them via `uartWireTiming()`.

### A defect I chased that did not exist

I had been reporting a residual "missing final byte": deriving the timing **by hand** for P=5 left 1/40 LEDs wrong and one byte short, and I attributed the remainder to a capture defect, then spent an iteration looking for it.

There was no such defect. My hand derivation omitted the minimum-symbol-separation bump that `fitUartWave()` applies (`kMinSymbolSeparationNs = 400`), so my decode windows were subtly wrong and clipped one LED. Calling the helper — which reproduces the encoder's rules exactly, bump included — decodes the frame completely.

The lesson is the one this change embodies: reimplementing the encoder's rounding at the decode site produces exactly the silent drift the encoder's comment warns about. I reproduced that bug by hand while investigating it.

### On the baud ceiling

`kMaxUartBaudRate` is used rather than the backend's own ceiling, which is not portably reachable from the harness. The geometry depends on the ceiling only through a feasibility gate, so the two agree wherever UART transmits at all — a backend whose ceiling excluded P=5 while the default admitted it would disagree, and the comment records that.

### DIFF
--- a/examples/AutoResearch/AutoResearchTest.cpp
+++ b/examples/AutoResearch/AutoResearchTest.cpp
@@ -14,6 +14,10 @@
 
 #include "AutoResearchTest.h"
 
+// uartWireTiming(): the decoder must classify against the quantised wire
+// timing the encoder produced, not the chipset's nominal values.
+#include "fl/channels/uart_wave_encoder.h"
+
 #include "platforms/arm/rp/is_rp.h"
 #if defined(FL_IS_RP2040) || defined(FL_IS_RP2350)
 // kRpPioRxEdgeCapacity bounds the static DMA/edge pool the RP PIO RX device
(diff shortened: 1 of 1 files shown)


zackees:
### Robustness checks on the newly-passing path

The original evidence was three single-frame runs at 40 LEDs, which is thin for a capability that had never worked. Two further checks on RP2350W `2DCB876B587EA334`:

**Multi-frame.** Second-frame degradation is a known bug class here — it is why SPI defaults to two frames (#2254/#2288) — and UART had never run more than one:

```
$ bash autoresearch rp2350w --uart --chipset ws2811-400 --strip-sizes 40 --frames 4
TEST UART0 lanes=1 leds=40 PASS 2075ms
```

2075 ms against 527 ms for a single frame, so all four frames genuinely ran rather than the flag being ignored. No degradation across them.

**Stability at the capacity boundary.** 60 LEDs is the largest size that fits the RX pool; 80 overflows. That is where margin is thinnest, so it is the size most likely to be marginal rather than cleanly passing:

```
60LED run1 -> PASS 768ms
60LED run2 -> PASS 767ms
60LED run3 -> PASS 769ms
```

Three consecutive passes with a 2 ms spread. A capture sitting near its buffer limit would be expected to show variance or intermittent overflow; it shows neither.

Combined with the earlier three passes at 40 LEDs and the PIO regression check (`PASS 685ms`), that is nine passing runs across three sizes and two frame counts, with no failures.

PR #4229, opened 2026-09-09
diag(net): report a missing HTTP response as such, not as a bad status

zackees:
## What this fixes

`runRpHttpRequestTest()` discarded the return value of `read_line()` for the status line. When nothing came back, `has_http_prefix` was false and the failure was reported as **`Unexpected HTTP status`** — with an empty `status_line`:

```
{'test': 'GET /leds', 'error': 'Unexpected HTTP status',
 'passed': False, 'body_read': 0, 'status_line': '', 'content_length': -1}
```

The status code is not what failed. Nothing was read at all. "The peer answered with the wrong code" and "the peer did not answer" have different causes and different fixes, and the message named the wrong one.

## The failure that motivated it

Captured on the bench during a `--net-peer` campaign on the RP2350W/ESP32-C6 pair:

- Direction **RP2350W → ESP32-C6** (RP is the client, C6 is the server).
- **Cycle 10 of 10**; cycles 1–9 clean, and three full 10-cycle runs before it passed clean. It is rare.
- **`GET /leds` only.** The other 11 requests in the same battery — against the same server, over the same association, including the 4096-byte FNV-1a `POST /echo` — all passed.
- `body_read: 0`, `content_length: -1`, `status_line: ''` — zero bytes read.

The `/leds` handler is a static string literal, structurally identical to the `/rpc/*` handlers that passed in the same battery, so the endpoint itself is not implicated. `TCP connect failed` is already a separate error, so the connection was established and the request went out.

## Changes

Distinguishes three outcomes instead of collapsing them into one:

- `No HTTP response: request sent, nothing read` — empty status line
- `Truncated HTTP status line` — partial line, deadline expired or peer closed mid-line
- `Unexpected HTTP status` — a real status line that was not the expected code

It also records `peer_still_open` on the no-response path, captured before the body loop and `client.stop()` can change it. That is the field that decides between the two remaining explanations the next time this reproduces:

- `true` → the C6 held the socket open and did not answer within the 2 s deadline (a server-side stall)
- `false` → the C6 closed the connection without answering

## Verification

- `bash compile rp2350 --examples AutoResearch` succeeds.
- C++ lint clean.
- `bash lint` also reports the `ci/autoresearch/net.py:719` `ty` warning, which is **already present on master** and unrelated to this change (an always-`None` `host_server` cleanup; removed on the #4214 branch).

Deliberately diagnostic-only: no retry, no timeout change, no behavioural fix. There is not yet evidence for what to fix, and this is what produces it.

Relates to #3899.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability of HTTP and echo connection checks by enforcing clear overall and inactivity time limits.
  - Connection attempts and data transfers now fail promptly when progress stops, while active transfers can continue without unnecessary timeout failures.
  - Request timing now includes the connection-establishment phase for more accurate diagnostics.


zackees:
Good catch — fixed in 54f2fc9.

You are right that `status_line[0] == '\0'` does not prove nothing arrived. `read_line()` drops `'\r'` before it ever reaches `out`, so a peer that sent a bare `'\r'` and then stalled or closed would have produced an empty buffer and been reported as *"No HTTP response: request sent, nothing read"* — which is the exact conflation this PR exists to remove, one level further down. It would also have been the most misleading possible case, since a lone `'\r'` is evidence the peer *did* start answering.

`read_line()` now records the bytes it consumed for the most recent line, counting the stripped `'\r'`, and the no-response branch keys off that instead of off the output buffer. A bare `'\r'` now lands in the `Truncated HTTP status line` branch, where it belongs.

Verified: `bash compile rp2350 --examples AutoResearch` succeeds.

For context on why this matters here rather than being theoretical: the failure that motivated the PR had `body_read: 0`, `content_length: -1` and an empty `status_line`, and the whole question it has to answer is whether the ESP32-C6 sent *nothing* or sent *something incomplete*. Getting that boundary wrong would have pointed the next investigation in the wrong direction.

zackees:
## The instrumentation answered the question it was added for

Two more failures were caught on the bench with this build, and the new fields are unambiguous:

```
GET /leds   'Truncated response body'
            body_read=0  elapsed_ms=2000  status_line='HTTP/1.1 200'  content_length=31
            peer_open_after_status=True   peer_open_after_body=True

GET /ping   'No HTTP response: request sent, nothing read'
            body_read=0  elapsed_ms=2000  status_line=''  content_length=-1
            peer_open_after_status=True   peer_open_after_body=True
```

Two independent fields agree in both captures:

- **`elapsed_ms = 2000` exactly** — the client's 2 s deadline ended the read. An early close would have broken the loop on `!client.connected()` well before that.
- **`peer_open_after_status` and `peer_open_after_body` both `True`** — the socket was still open at both checkpoints, captured before `stop()` makes `connected()` false regardless.

**Conclusion: the ESP32-C6 accepts the connection, takes the request, and then stops sending while holding the socket open.** That rules out early close, reset, truncation-on-close, and corruption — all of which were live hypotheses before these fields existed.

It also confirms the two error strings this PR introduced are describing **one** fault, not two. Across four captured failures on three different endpoints (`/leds`, `/status`, `/ping`), the difference between `No HTTP response` and `Truncated response body` is only *where in the response* the stall happened to land — before the status line, or after the headers. Both always with `body_read=0`, and the 4096-byte echo has never failed in 241 cycles, so size is not the trigger.

Worth noting the reviewer's `'\r'` catch mattered here: `GET /ping` came back with a genuinely empty status line and `status_bytes_seen` false, so it classified correctly. Under the original `status_line[0] == '\0'` test I could not have distinguished that from a peer that sent a bare `'\r'`, and the "peer sent nothing" conclusion above would have been unsound.

Regression evidence for the change itself: **22 clean 10-cycle runs across 241 cycles**, with builds from this branch running a large share of them.

A follow-up experiment is running on `exp/rp-http-stall-duration` (not for merge): a 10 s client budget, to separate "the C6 is slow and the response lands late" from "the C6 drops the response entirely". `elapsed_ms = 2000` proves the stall lasts *at least* 2 s but cannot say whether it ever ends. This PR stays diagnostic-only.

zackees:

PR #4230, opened 2026-09-09
diag(net): name the failing sub-test instead of dumping all twelve

zackees:
## Why

`runNetClientTest` returns a 12-entry `results` array, and the peer harness raised it verbatim:

```
RP2350W -> ESP32-C6 HTTP failed: {'results': [{'test': 'GET /ping', 'passed': True, ...
```

When a real failure landed on the bench during a `--net-peer` campaign, the single row that mattered sat in the middle of eleven passing ones, on one unwrapped line. Finding it was the hardest part of reading the failure.

## After

The diagnosis leads; the full payload still follows, so nothing is lost:

```
RP2350W -> ESP32-C6 HTTP failed: 1 of 12 sub-tests failed --
GET /leds: No HTTP response: request sent, nothing read
(status_line='' body_read=0 content_length=-1); full result: {...}
```

`status_line` is named explicitly because it is the field that separates **"answered with the wrong code"** from **"did not answer at all"** — a distinction #4229 introduces on the firmware side.

## Scope

Only the **RP2350W → ESP32-C6** raise site is changed. The opposite direction is edited by #4222, and leaving it untouched keeps the two PRs independent rather than conflicting.

## Tests

Two tests, using the payload actually captured on the bench (cycle 10, `GET /leds`, zero bytes read). Both were verified to **fail** when the helper stops filtering passing rows — so they test the behaviour, not just the happy path:

```
2 failed, 6 deselected     # helper broken
2 passed, 6 deselected     # helper restored
```

Collection count checked, not just the pass count: 8 collected (was 6), 8 passed.

`bash lint` also reports the pre-existing `ci/autoresearch/net.py` `redundant-condition` warning for the always-`None` `host_server` cleanup. It is **already on master** and unrelated to this change; it is removed on the #4214 branch, so it is not duplicated here to avoid a conflict between two open PRs.

Relates to #3899.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTTP test failure diagnostics by identifying failed sub-tests and including relevant status details.
  * Malformed test results are now reported with their position and value instead of being silently ignored.
  * Join attempts now allow additional polling time while respecting the run deadline, with clearer timeout details.
  * Successful device joins now report elapsed join time and IP address.

* **Tests**
  * Added coverage for failed and passing sub-tests, missing or invalid reports, malformed result entries, and join diagnostics.


zackees:
Fixed in 0eb1f75.

Correct — a non-dict entry in `results` was skipped, so a report whose rows are malformed would have read as *"no sub-test reported a failure"*. That is the same silent skip this helper was written to remove, so it should not have been reintroduced inside it. A row that is not an object cannot be shown to have passed, and is now reported as one of the failures:

```
1 of 2 sub-tests failed -- result[1] is not an object: 'not-a-dict'
```

Added a test covering it. 9 collected, 9 passed (collection count checked, not just the pass count).

zackees:

PR #4231, opened 2026-09-09
fix(net): stop abandoning requests the peer is still serving

zackees:
## The defect

The RP HTTP client gave each request **2000 ms**. The peer's HTTP server allows itself **5000 ms** for the same request. The client was therefore abandoning work the server was still legitimately performing.

The peer does not answer from its network task. On ESP, `handle_esp_request()` queues the request to the sketch's main loop and blocks on a semaphore waiting for it:

```
httpd task → xQueueSend(s_esp_request_queue) → blocks 5000 ms on a semaphore
                      ↓
     ServerAsyncRunner::update()  ←  task::Executor, pumped from the MAIN task
```

Nothing is written until that completes, so **response latency is bounded by the peer's loop cadence, not by transmission**. That is why 4-to-55-byte static string responses are slow: measured median **552 ms**, p90 **890 ms**, over 360 samples.

## Evidence that the budget was the fault, not the peer

This was not guessed from the code. The client was instrumented first (#4229) and the fix follows the measurement.

**At 2000 ms — 241 cycles, 4 hard failures.** Every one had `elapsed_ms` of exactly 2000, the socket still open at both checkpoints, and `body_read = 0`:

```
GET /status  'Truncated response body'  body_read=0  elapsed_ms=2000
             status_line='HTTP/1.1 200'  content_length=54
             peer_open_after_status=True  peer_open_after_body=True
```

Exactly 2000 ms is the client's own deadline; the peer neither closed nor reset. Failures appeared on three different endpoints (`/leds`, `/status`, `/ping`), and the 4096-byte echo never failed once — so neither endpoint nor payload size was implicated.

**At the larger budget — 215 cycles, 0 failures, and the confirming observation.** Five requests exceeded 2000 ms and **completed correctly**:

| Request | Elapsed |
|---|---|
| `GET /leds` | 2706 ms |
| `GET /leds` | 2734 ms |
| `GET /rpc/discover` | 2961 ms |
| `GET /leds` | 3108 ms |
| `GET /rpc/discover` | 3586 ms |

Each would have failed under 2000 ms. This is the positive evidence, not merely an absence of failures — I deliberately did not treat the zero-failure count as proof, because a zero is also consistent with the condition simply not occurring.

The maximum observed, **3586 ms**, sits under the server's own 5000 ms budget, exactly as the mechanism predicts.

## The change

One line: the deadline becomes `kRpPeerMaxRequestMs` (10000 ms) instead of a hardcoded 2000 ms, so both sides of the fixture agree on how long a request may take rather than the client being stricter than the server.

## What this does not do

It does not make the peer faster. A bench fixture whose main loop can delay an HTTP response by 3.5 s is worth addressing on its own, and this change does not address it — it stops the client from misreporting that latency as a peer failure. Reducing the C6's loop cadence is a separate piece of work.

## Verification

- `bash compile rp2350 --examples AutoResearch` succeeds.
- Behaviour measured across **456 cycles** total on the RP2350W/ESP32-C6 bench.

Diagnostics that produced this: #4229 (`elapsed_ms`, `peer_open_after_*`) and #4230 (names the failing sub-test). Fixes the HTTP leg of #3899.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved RP2350W CYW43 HTTP request handling by allowing the client to wait for the full server request window.
  * Prevented valid, longer-running requests from timing out prematurely.


zackees:

PR #4318, opened 2026-09-10
fix(ota): unblock four stages of the RP2350W peer OTA flow, and stop the criterion passing without one (#3956)

### LINKED ISSUE #3956: follow-up: harden RP2350W peer OTA flow from #3863
Source PR: https://github.com/FastLED/FastLED/pull/3863

Source comments:
- https://github.com/FastLED/FastLED/pull/3863#discussion_r3727663993
- https://github.com/FastLED/FastLED/pull/3863#discussion_r3727664020
- https://github.com/FastLED/FastLED/pull/3863#discussion_r3727664039
- https://github.com/FastLED/FastLED/pull/3863#discussion_r3727664044
- https://github.com/FastLED/FastLED/pull/3863#discussion_r3727664055

Why this matters: current peer-OTA success accounting and HTTP framing can report success for a failed transfer, while a blocking download can exceed the RP2350W watchdog window and an out-of-range port can silently target the wrong endpoint.

CodeRabbit suggestions (verbatim):

> **Guard the `servedRequests` conversion.**
>
> `int(served.get(servedRequests, 0))` raises `ValueError` or `TypeError` when the device returns a non-numeric value. Neither exception is in the handler tuple at Line 161, so the run aborts with a traceback instead of the failure message.

> **Feed the watchdog during the blocking OTA download.**
>
> `pollOtaArtifactUpdate` calls `HTTPUpdate::update(client, ...)` synchronously, but `loop()` only feeds `FastLED.watchdog()` at scope start/end via `FL_WATCHDG_AUTO(AUTORESEARCH_WATCHDOG_TIMEOUT_MS)`. A full firmware download can exceed the configured `5000` ms timeout and reset the board mid-upgrade. Disable the watchdog for this update or register an HTTPUpdate progress callback that calls `FastLED.watchdog().feed()`.

> **Count only served requests that deliver the artifact.**
>
> `serveOtaArtifact` increments `served_requests` before it opens the file. A 404 response then still counts. `run_ota_peer_autoresearch` treats `servedRequests >= 1` as proof that the RP2350W downloaded the image, so a failed download can pass the check. Increment the counter after the transfer completes.

> **Don’t add `Content-Length` to a chunked OTA response.**
>
> `httpd_resp_send_chunk` sends HTTP responses as chunked transfer encoding. Setting `Content-Length` adds conflicting OTA framing, so use `httpd_resp_send` with the full buffer instead, or confirm this Arduino-Pico OTA client accepts chunked responses for the active core version.

> **Reject port values outside the `uint16_t` range.**
>
> `static_cast<uint16_t>(port)` truncates. A request with `port: 70000` becomes port 4464 and the update targets the wrong endpoint. Validate the upper bound too.

Acceptance criteria:

- RED: focused host tests show malformed/non-numeric `servedRequests` is reported as a controlled failure and a 404/partial artifact response cannot satisfy served-transfer proof.
- GREEN: validate status types explicitly and increment `served_requests` only after a complete successful response.
- RED: a focused request test shows ports <= 0 and > 65535 are rejected before narrowing.
- GREEN: accept only the full `uint16_t` endpoint range without truncation.
- RED: a focused OTA transport repro exercises an update lasting longer than the 5-second watchdog interval and the current conflicting Content-Length/chunked response framing.
- GREEN: keep the watchdog fed (or safely disable/restore it) during the blocking update and emit one standards-compliant framing mode supported by the pinned Arduino-Pico HTTPUpdate client.
- Preserve cleanup on read/send/update failure and add deterministic tests for success, 404, partial-send, malformed status, timeout/watchdog, and framing behavior.
- Run focused tests, strict lint, RP2350W and ESP32-C6 builds, then prove OTA build-marker, reboot, CDC/network recovery, and serial RPC health on hardware.

zackees:
Peer OTA does not complete. Three separate failures sit behind one another; this fixes the first two on hardware and makes the third diagnosable instead of silent.

I closed #3956 as completed earlier today on the strength of having root-caused the 512-byte chunk limit. **That was premature** — clearing that limit only reveals the next two layers, both of which are real device-side bugs.

## 1. The artifact server could never start

```
OTA peer autoresearch failed: C6 artifact server failed:
  {'error': 'Artifact HTTP server start failed', 'success': False}
```

`startOtaArtifactServer` calls `httpd_start` with `HTTPD_DEFAULT_CONFIG()`, which hardcodes `.ctrl_port = ESP_HTTPD_DEF_CTRL_PORT` — 32768, confirmed in the IDF header rather than assumed. `startNetServer` runs immediately before it in this flow, and `fl::asio::http::Server::start()` (`src/fl/stl/asio/http/server.cpp.hpp:980`) takes the same default. The second server therefore tries to bind a control socket the first already owns.

A distinct `server_port` does not help: **the collision is on the control socket, not the listening socket** — which is why 8081-vs-80 looked like it should have been sufficient.

## 2. The second transfer onto any given board always failed

```
OTA peer autoresearch failed: C6 artifact write failed at byte 200448:
  {'error': 'Artifact write failed', 'success': False, 'received': 200448}
```

`finishOtaArtifact` replaces `/rp2350.bin` only once `.part` is complete and its SHA-256 verifies — correct when there is room for both copies. There is not:

| | bytes |
|---|---|
| fixture spiffs partition `0x130000` | 1,245,184 |
| RP image | 1,030,348 |
| headroom with the previous artifact present | 214,836 |
| **observed write failure** | **200,448** |

The guard checked `expected_size > LittleFS.totalBytes()` — one copy — so a two-copy shortfall surfaced as a bare "write failed" twenty minutes into a transfer.

This is what made the two runs look contradictory: run 1 hit a blank filesystem and transferred all 1,030,348 bytes with a matching SHA-256; run 2 inherited run 1's artifact and died 200 KB in. One cause, both runs, to within LittleFS block overhead. I would have written run 2 off as flaky without doing that subtraction.

The old artifact is now dropped at `begin`, and the guard checks real free space so any remaining shortfall fails immediately with `freeBytes`/`neededBytes` attached. The cost — a failed transfer leaves no artifact rather than the previous one — is the right trade for a fixture the harness re-stages every run, against the alternative of no run after the first being able to stage at all.

## 3. The third failure had no reason anywhere

With those two fixed, the C6 holds the artifact and its server is up, and serves nothing:

```
OTA peer autoresearch failed: C6 did not serve the RP2350W artifact:
  {'success': True, 'received': 1030348, 'servedRequests': 0}
```

Note what that line already proves: `received: 1030348` is the **full artifact transferred onto a non-blank filesystem**, and reaching this raise at all requires `finishOtaArtifact` to have matched the SHA-256, `startNetServer` to have come up, the RP to have joined the AP, and `applyOtaArtifact` to have been accepted. Only the HTTP GET never happened.

The RP does explain itself — `FL_WARN("[OTA] RP2350W artifact update failed: ...")` — but `pollOtaArtifactUpdate` runs from `loop()` *after* the harness deliberately closes the RPC link so the board can reboot into the new image. The explanation goes to a serial port with nobody on the other end. A fetch that failed and a fetch that never ran produce byte-identical evidence.

`rpOtaUpdateStatus` keeps `attempted` / `succeeded` / `lastError` in device state, and the host asks for it once the link is back, so the next run names the cause.

## Verification

Attached pair: RP2350W `2DCB876B587EA334` + ESP32-C6 `8C:BF:EA:CF:87:B4`. Each fix is confirmed by the run advancing past the failure it previously died on — server start went from `Artifact HTTP server start failed` to `success: True`, and a transfer onto a non-blank filesystem went from failing at byte 200,448 to `received: 1030348`. Device schema goes 75 → 76 methods with `rpOtaUpdateStatus` bound.

## What this does not do

**It does not make peer OTA pass.** `kOtaChunkBytes` stays at 512, and the preflight refuses that size, so the flow still cannot run end to end unedited.

I have evidence that 256 works — it transferred and SHA-verified twice here — but only because I passed `--timeout 2500`. Against the default budget 256 needs ~21 minutes, which is exactly why 512 was chosen when *both* values failed. The honest fix is the constant **plus** a budget the mode can meet, and reversing that decision deserves the measurements argued on the issue rather than a constant quietly flipped inside a bug fix. I will post those numbers to #3956 separately.

Found while working #3899's OTA criterion.

## Summary by CodeRabbit

* **New Features**
  * Added OTA status reporting for whether an update was attempted, succeeded, or failed, including error details.
  * Added remote access to peer OTA status after reconnection.
  * Added early detection of insufficient storage for OTA artifacts.
  * Improved peer-OTA failure messages with diagnostics from both devices.

* **Bug Fixes**
  * Increased storage available for peer OTA artifacts.
  * Prevented stale artifact download counts after publishing a new artifact.
  * OTA completion now requires confirmed application, not merely successful transfer.

* **Documentation**
  * Added lessons on execution diagnostics, troubleshooting, memory analysis, and intermittent failures.

### DIFF
--- a/agents/tasks/lessons.md
+++ b/agents/tasks/lessons.md
@@ -363,6 +363,36 @@
   minutes left. Anchor a completion pattern on text only the result can
   produce (`AUTORESEARCH PASSED`, `autoresearch failed`), and check what the
   first hundred lines of the log already contain before trusting a match.
+- On a machine with no swap, a "low memory" alarm can mean a full page cache
+  rather than exhaustion -- read `MemFree` and `MemAvailable` separately
+  before acting on it. Three long device runs were killed for low memory
+  while a 20-minute sampler I had running showed `MemAvailable` never
+  dropping below 77 GB of 131 GB. The real reading was `MemFree: 1.4 GB`
+  against `Cached: 59.6 GB` with `SwapTotal: 0`: build I/O had filled the
+  page cache, so anything thresholding on `MemFree` saw 1.4 GB while
+  `MemAvailable`, which counts reclaimable cache, correctly reported 80 GB.
+  This corrects my earlier conclusion that daemon growth was the cause.
+  Killing `fbuild-daemon` and `zccache-daemon` appeared to fix it once
+  because it briefly returned pages to `MemFree`, then stopped working,
+  because the daemons were never it -- `fbuild-daemon` was back at 5.26 GB
+  within ten minutes and the run died anyway. Page cache is reclaimable and
+  cannot be dropped from userspace without root, so the honest responses are
+  to retry, shorten the exposed run, or fix the threshold -- not to reap
+  processes and call it fixed.
+- Do not hunt a deterministic cause for an intermittent failure until you have
+  established it is deterministic. `No response with ID 1 within 20.0s` failed
+  two peer-OTA runs in a row, which I took as proof it was not flaky, and I
+  then produced four wrong causes in sequence -- a first-boot filesystem
+  format, a 2 MB filesystem slowing boot, a short `boot_wait`, and a stale
+  CDC on the primary. Each was built on a two-sample correlation with my own
+  most recent change. The next run passed the same point untouched: it is a
+  race on the companion's USB-CDC re-enumeration, which
+  `_connect_peer_with_retry`'s docstring had already recorded across four
+  prior captures, naming the C6 and saying a fixed `boot_wait` cannot cover
+  it. Two consecutive failures of a race are unremarkable. Read the existing
+  docstrings on the failing path before forming a theory, and prefer adding
+  attribution over adding a hypothesis when the error names neither the
+  component nor the operation.
 - `git fetch` before starting work, not before pushing. I root-caused the
   `s16x16x4` 15x penalty in #4216 (rolled 4-lane loops defeating
   scalar-replacement), fixed it, and measured it on hardware -- and only when
--- a/agents/tasks/lessons.md
+++ b/agents/tasks/lessons.md
@@ -325,6 +325,44 @@
   content, not state, and
   when commits are stranded rebuild them onto current master as a fresh PR
   rather than reopening the merged one.
+- The last line of a log and the code path that produced it can disagree, and
+  the code path wins. A peer OTA run ended with `C6 artifact server failed`,
+  and I read the tail as "the transfer never ran". Reading `ota.py` in order
+  showed the opposite: reaching that raise requires the whole
+  `writeOtaArtifact` chunk loop to have returned success and
+  `finishOtaArtifact` to have matched the SHA-256, because either one failing
+  raises earlier and differently. The transfer had in fact completed and
+  verified 1,030,348 bytes. An error message names where execution stopped,
+  which is also a statement that everything before it passed -- for a
+  sequential flow that is free evidence, and I nearly discarded it by reading
+  bottom-up. Before concluding a stage did not run, find the raise that would
+  have fired if it had failed.
+- When two runs of the same thing disagree, subtract before calling either
+  one flaky. Run 1 of a peer OTA transferred all 1,030,348 bytes; run 2 died
+  at byte 200,448. The fixture partition is 0x130000 (1,245,184 B) and
+  `finishOtaArtifact` deletes the previous artifact only after the new one
+  verifies, so run 2 needed room for two copies:
+  1,245,184 - 1,030,348 = 214,836 B of headroom against a failure at 200,448 B,
+  the difference being LittleFS block overhead. One cause explained both runs
+  exactly. Arithmetic that lands within a few percent is a root cause;
+  "probably flaky" is what I would have written without doing it.
+- A stale branch head is not evidence that nobody is working on it. I
+  diagnosed the seven red builds on #4311, posted the cause, then implemented
+  the fix anyway because the head had not moved in 21 minutes. It was another
+  session's active branch; they pushed the same fix while I was testing and my
+  push was rejected. My existing lesson said to fetch before starting, and I
+  did fetch -- the gap was treating "no commits for N minutes" as "abandoned".
+  On a branch this session does not own, the diagnosis comment is the whole
+  contribution; implementing on top of it is duplicated work with a merge
+  conflict attached. See [[fetch-before-starting-not-before-pushing]].
+- Grep patterns used as completion signals must not match the configuration
+  that announces the work. I polled a log for `TEST_COMPLETED` to detect the
+  end of a 20-minute run and got a hit within 30 seconds -- from the startup
+  banner `stop on 'TEST_COMPLETED_EXIT_(OK|ERROR)'`, which is the harness
+  printing what it will look for. I reported the run finished while it had 18
+  minutes left. Anchor a completion pattern on text only the result can
+  produce (`AUTORESEARCH PASSED`, `autoresearch failed`), and check what the
+  first hundred lines of the log already contain before trusting a match.
 - `git fetch` before starting work, not before pushing. I root-caused the
   `s16x16x4` 15x penalty in #4216 (rolled 4-lane loops defeating
   scalar-replacement), fixed it, and measured it on hardware -- and only when
--- a/ci/autoresearch/ota.py
+++ b/ci/autoresearch/ota.py
@@ -371,7 +371,19 @@ async def run_ota_peer_autoresearch(
         await rpc_data(primary, "ping", {})
         served = await rpc_data(peer, "otaArtifactStatus", {})
         if not served.get("success") or _served_request_count(served) < 1:
-            raise RuntimeError(f"C6 did not serve the RP2350W artifact: {served}")
+            # The RP fetches with the RPC link closed, so its own reason for
+            # failing never reaches the host live. Ask for it now that the
+            # link is back: without this the C6's `servedRequests: 0` is the
+            # whole report, and a fetch that failed looks exactly like one
+            # that never ran. See FastLED#3956.
+            try:
+                rp_update = await rpc_data(primary, "rpOtaUpdateStatus", {})
+            except (RpcTimeoutError, RuntimeError, OSError) as probe_error:
+                rp_update = {"probeFailed": str(probe_error)}
+            raise RuntimeError(
+                f"C6 did not serve the RP2350W artifact: {served}; "
+                f"RP2350W update state: {rp_update}"
+            )
         print(f"{Fore.GREEN}OTA PEER AUTORESEARCH PASSED{Style.RESET_ALL}")
         return 0
     except KeyboardInterrupt as ki:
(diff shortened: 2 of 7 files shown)


zackees:
Status on the one unverified claim in this PR, so a reviewer knows where it stands.

The `2m` staging-region fix (layer 5, `Update error: ERROR[4]: Not Enough Space`) is still argued from three things rather than from a green run:

- the device's own error, surfaced by `rpOtaUpdateStatus` in this same PR
- `Updater.cpp:101` / `:121` in the arduino-pico core, which stage an incoming image into the filesystem region
- the arithmetic: 524,288 B region against a 1,030,348 B image

I attempted the end-to-end run **six** times. Every attempt was killed mid-run by this bench's memory monitor. That monitor is firing on `MemFree` — 1.4 GB, because build I/O had filled the page cache to 59.6 GB with no swap — while `MemAvailable` never dropped below 77 GB of 131 GB across a 20-minute sampler. Page cache is reclaimable and cannot be dropped from userspace without root, so this is an environment limit rather than something the change can address.

What *is* verified: the `2m` build compiles, deploys, and passes `--rpc-smoke`; `1.5m` reproducibly fails the `PicoOTA.cpp` static assert; and layers 2, 3 and 4 each showed the run advancing past the failure it previously died on.

Not asking for the claim to be taken on trust — flagging it so the `2m` line gets read as "well-argued, not yet demonstrated". The remaining blocker to a full green run is `kOtaChunkBytes`, which this PR deliberately leaves at 512; the numbers for lowering it are on #3956.

zackees:
Follow-up on the verification attempt, since I said I would report where it landed.

I tried once more on the theory that the build — the phase being killed — was avoidable. `run_fbuild_deploy`'s docstring says the daemon "checks the firmware ledger — if source hash + build flags are unchanged, the entire build+flash is skipped (returns in ~1s)", and with this branch committed the only pending edit was host-side.

**The ledger did not hit.** Deploy took 76.3 s, a full build. AutoResearch re-stages the sketch into the build directory on every run (`Copied examples/AutoResearch/… to .build/pio/rp2350w/src/sketch`), which changes what the ledger keys on, so the build is not skippable this way regardless of what changed in git. Recording that because the docstring reads like it should have worked, and the next person to try this shortcut will hit the same wall.

Seventh attempt, killed at the transfer again. `MemAvailable` was 76.5 GB of 131 GB at the time.

So the status of the `2m` staging fix is unchanged and I am leaving it that way: argued from the device's own `ERROR[4]: Not Enough Space`, from `Updater.cpp:101`/`:121`, and from 524,288 B against a 1,030,348 B image — not demonstrated by a green run. Layers 2, 3 and 4 remain directly verified by the run advancing past each failure it previously died on.

If someone runs this on a bench whose memory monitor reads `MemAvailable` rather than `MemFree`, the command is:

```
bash autoresearch rp2350w --net-peer --ota \
  --upload-port "SER=<rp-serial>" --peer-upload-port <c6-port> \
  --peer-environment esp32c6 --timeout 2500
```

with `kOtaChunkBytes` lowered — 318 encodes to 424 base64 characters, inside the 428 the device answers, and needs 3,241 round trips rather than 256's 4,025.

zackees:
Applied all three findings, after verifying each against the code.

**`served_requests` never reset** — confirmed. Initialised once in singleton state, incremented in `serveOtaArtifact`, read by `otaArtifactStatus`, never cleared. So the host's `servedRequests < 1` failure condition could be satisfied by a *previous* artifact's count: the run scored on a download that happened before the current image existed. Reflashing the C6 between runs hides it, which is why it has not bitten — but the whole point of that check is to be decisive. Now zeroed when `finishOtaArtifact` publishes.

**Pass path not requiring the update to have applied** — this is the better of the three catches. Returning `PASSED` on `servedRequests >= 1` proves a download happened and nothing more. `HTTPUpdate` can fetch an entire artifact and still reject it, and `ERROR[4]: Not Enough Space` — the exact failure this branch diagnoses in layer 5 — is that shape. **That failure would have been scored a pass here**, which is uncomfortable given this branch is otherwise about hunting vacuous passes. Now requires `attempted` and `succeeded` from `rpOtaUpdateStatus`.

**`KeyboardInterrupt` ahead of the probe's tuple** — applied. It derives from BaseException so the tuple cannot catch it today; the value is that this file keeps the handler explicit everywhere else, so a later widening of that tuple cannot silently start swallowing Ctrl-C. The probe sits on the failure path, where an interrupt is most likely.

`bash lint` clean. Note the second change tightens the pass condition, so it can only make this criterion harder to satisfy, never easier — the end-to-end run remains unverified here for the environment reasons above, and this does not change that.

zackees:
## Scope update — this PR has grown past its original description

The title and body describe only the first commit. Three more have landed since, so here is the current contents in one place. I have retitled it; the body below supersedes the original where they differ.

**Four layers of the peer OTA flow, each hiding the next:**

| # | failure | fix | verified |
|---|---|---|---|
| 1 | `startOtaArtifactServer` → `Artifact HTTP server start failed` | `config.ctrl_port = 32769` — `HTTPD_DEFAULT_CONFIG()` hardcodes 32768 and `fl::asio::http::Server` takes the same default, so the second server could never bind | **yes** — now returns `success: true` |
| 2 | `Artifact write failed at byte 200448` on every run after the first | drop the stale artifact at `begin` and guard on real free space — the fixture partition (1,245,184 B) cannot hold two copies of a 1,030,348 B image | **yes** — `received: 1030348` on a non-blank filesystem |
| 3 | `servedRequests: 0` with no reason anywhere | `rpOtaUpdateStatus` — the RP's own error was going to a serial link the harness deliberately closes | **yes** — named layer 4 on its first run |
| 4 | `Update error: ERROR[4]: Not Enough Space` | `board_build_filesystem_size` 0.5m → **2m** — arduino-pico stages OTA images *into* the filesystem region (`Updater.cpp:101`/`:121`), and 524,288 B cannot hold a 1,030,348 B image | **no** — see below |

**Plus two review findings, both the same defect class this PR is about — a check that can pass without the thing it checks happening:**

- `served_requests` was never reset, so a *previous* artifact's count satisfied the current run's `servedRequests < 1` check
- the pass path returned `PASSED` on the served count alone. `HTTPUpdate` can fetch an entire artifact and still reject it — **layer 4's own failure would have been scored a pass by the criterion meant to catch it**

And `ota_applied_ok` with five tests pinning that, including the verbatim `ERROR[4]` payload, confirmed to fail against the weaker predicate they replace.

## Still unverified, and why

Layer 4's `2m` fix rests on the device's own error, the arduino-pico source, and the arithmetic — **not** on a green end-to-end run. Seven attempts were killed mid-run by this bench's memory monitor, which fires on `MemFree` (1.4 GB, page cache full, no swap) while `MemAvailable` never dropped below 77 GB of 131 GB. Layers 1–3 are each directly verified by the run advancing past the failure it previously died on.

`kOtaChunkBytes` also stays at 512, so the flow still cannot run end to end unedited. 318 raw bytes encodes to 424 base64 characters — inside the 428 the device is measured to answer — so the choice was never only between undeliverable 512 and slow 256. That argument belongs on #3956 with the numbers, not inside a bug fix.

Issue #4375, opened 2026-09-11
RP UART diagnostics are stale when the engine declines a channel: show() returns before resetting them


zackees:
`runSingleTest` reports `rpUartStartAttempted`, `rpUartStartSucceeded`, `rpUartEncodedSize`, `rpUartActualBaud` and `rpUartLastError` from the previous UART run whenever the engine declines the current channel. The values look like a report on the test just run; they are not.

## Cause

`channel_engine_rp_uart.cpp.hpp:73`

```cpp
void ChannelEngineRpUart::show() FL_NO_EXCEPT {
    if (mActive || mPendingChannels.empty()) {
        return;                       // <-- early return
    }
    ...
    mLastStartAttempted = false;      // <-- resets live below it
    mLastStartSucceeded = false;
    mLastEncodedSize = 0;
    mLastActualBaud = 0;
    mLastError.clear();
```

`enqueue()` only pushes a channel when `canHandle()` accepts it. When `canHandle()` rejects — wrong pin for the instance, timing needing a baud above the backend maximum — nothing is queued, `mPendingChannels` is empty, and `show()` returns at the guard without ever reaching the reset. The previous run's values are then what the harness reports.

## Observed

Driving `runSingleTest` directly on an RP2350W. `WS2812B-V5` has a 1225 ns period, so at P=5 it needs ~4.08 MHz — above the backend maximum, which is why it is documented to decline:

```
UART0  WS2812B-V5   attempted=True  baud=2000000  err=''
```

**2000000 baud is arithmetically impossible for a 1225 ns period.** It is the `WS2811-400KHZ` value (5 / 2500 ns), left over from an earlier call in the same device session. A fresh flash reports the truth:

```
RP UART: attempted=False started=False encodedBytes=0 actualBaud=0
         lastError='RP UART: chipset timing needs a baud above the UART backend maximum'
```

Same driver, same timing, opposite report — the difference is only what ran before it.

`UART1` shows the other half of the same defect. It declines consistently, and because `mLastError` is never set on this path and never cleared, the reason is blank:

```
UART1  WS2811-400KHZ  attempted=False  baud=0  err=''
UART1  WS2812B-V5     attempted=False  baud=0  err=''
```

A caller cannot distinguish "the engine declined and said why" from "the engine declined silently" from "these numbers describe a different test".

## Why this matters

[...]

PR #4376, opened 2026-09-11
fix(rp-uart): stop a declined channel reporting the previous run's diagnostics (#4375)

zackees:
When the RP UART engine declines a channel, `runSingleTest` reports `rpUartStartAttempted`, `rpUartActualBaud` and `rpUartEncodedSize` from the **previous** run, with an empty `rpUartLastError`. The numbers read as a report on the test just run.

## Cause

`show()` resets those fields — but *below* its guard:

```cpp
void ChannelEngineRpUart::show() FL_NO_EXCEPT {
    if (mActive || mPendingChannels.empty()) {
        return;                       // <-- decline lands here
    }
    ...
    mLastStartAttempted = false;      // <-- never reached
    mLastActualBaud = 0;
    mLastError.clear();
```

`enqueue()` only pushes when `canHandle()` accepts, so a decline leaves the queue empty, `show()` returns at the guard, and the old values survive to be reported.

## Why the reset goes in `enqueue()` and not above that guard

`show()` runs every frame. Resetting before the guard would clear a *successful* run's diagnostics on the next idle frame, before the caller ever read them — trading a stale report for an empty one. Clearing only on the decline path leaves the accepting path untouched.

`mLastError` is deliberately not cleared there: `canHandle()` has just set it to the reason.

## Second half: a guard that declined silently

```cpp
if (!data || !data->isClockless() || !isValidTxPin(data->getPin())) {
    return false;                     // no reason recorded
}
```

The three guards below it all set `mLastError`. This one did not, so a decline for a TX pin the instance cannot drive was indistinguishable from the engine never being consulted. Now it says which of the three conditions fired.

## Verified on hardware

RP2350W `2DCB876B587EA334`. With pins at tx=1 — GPIO1 is UART0's RX, so neither instance can drive it:

```
before   attempted=True   baud=2000000   err=''
         (2000000 is the WS2811-400KHZ figure from an earlier call, and
          arithmetically impossible for the 1225 ns period requested)

after    attempted=False  baud=0
         err='RP UART: TX pin is not usable by this UART instance'
```

Accepting path unchanged — tx=0 with `WS2811-400KHZ` still reports `attempted=True baud=2000000`, correct for a 2500 ns period at P=5. `--rpc-smoke` passes, `bash lint` clean.

## Explicitly not fixed here

A per-call `timing` of `WS2812B-V5` still reaches the engine as a 2 MHz baud — the 2500 ns figure, where a 1225 ns period needs ~4.08 MHz. That is a question about how per-call timing reaches the engine, not about diagnostic staleness, and this change does not affect it either way. I found it while verifying this fix and am flagging rather than folding it in.

## Why it matters

A diagnostic that reports confidently about a run that did not happen. It cost me several wrong readings today: I recorded UART1 as failing in a way I could not explain, and separately reported "a decline path that reports nothing" on #3899 without being able to locate it — this was that path.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved UART diagnostics by reporting the specific reason a channel cannot be handled.
  * Cleared outdated transmission metrics when channel evaluation fails, preventing stale values from being displayed.
  * Preserved the latest transmission error until status polling reports it.
  * Ensured diagnostics reset consistently when channel data is missing or transmission pins are unavailable.

### DIFF
--- a/src/platforms/arm/rp/rpcommon/channel_engine_rp_uart.cpp.hpp
+++ b/src/platforms/arm/rp/rpcommon/channel_engine_rp_uart.cpp.hpp
@@ -36,7 +36,29 @@ bool ChannelEngineRpUart::isValidTxPin(int pin) const FL_NO_EXCEPT {
 }
 
 bool ChannelEngineRpUart::canHandle(const ChannelDataPtr& data) const FL_NO_EXCEPT {
+    // Every call starts a fresh evaluation, so the previous run's numbers
+    // stop being reportable here rather than in enqueue(). canHandle() is
+    // also reached during driver selection without enqueue() following, which
+    // is the path that left `attempted=True baud=2000000` visible on a
+    // channel this engine had just refused. show() re-establishes these for a
+    // channel that is accepted and actually runs.
+    mLastStartAttempted = false;
+    mLastStartSucceeded = false;
+    mLastEncodedSize = 0;
+    mLastActualBaud = 0;
     if (!data || !data->isClockless() || !isValidTxPin(data->getPin())) {
+        // Say why. The three guards below all record a reason; this one did
+        // not, so a decline for a pin this UART instance cannot drive was
+        // indistinguishable from the engine never being consulted -- the
+        // caller saw `attempted=false` with an empty error either way.
+        // FastLED#4375.
+        if (!data) {
+            mLastError = "RP UART: no channel data";
+        } else if (!data->isClockless()) {
+            mLastError = "RP UART: channel is not clockless";
+        } else {
+            mLastError = "RP UART: TX pin is not usable by this UART instance";
+        }
         return false;
     }
     // Ask the backend what it can actually reach rather than assuming a fixed
@@ -61,11 +83,29 @@ bool ChannelEngineRpUart::canHandle(const ChannelDataPtr& data) const FL_NO_EXCE
                      "backend maximum";
         return false;
     }
+    // Clear the reason here rather than with the numerics above: show()
+    // records a failed DMA start in mLastError and poll() is what reports it,
+    // so an evaluation must not be able to wipe that. Accepting is the only
+    // outcome that makes a previous decline's reason wrong, and this is the
+    // only path that reaches it.
+    mLastError.clear();
     return true;
 }
 
 void ChannelEngineRpUart::enqueue(ChannelDataPtr channelData) FL_NO_EXCEPT {
-    if (channelData && canHandle(channelData)) {
+    // canHandle() clears the numeric diagnostics and, on a decline, records
+    // the reason -- so nothing is needed here. It is done there rather than
+    // in show() because show() returns at its `mPendingChannels.empty()`
+    // guard before reaching its own reset, and an empty queue is exactly what
+    // a decline produces. Resetting above that guard would be worse: show()
+    // runs every frame, so it would wipe a successful run's diagnostics
+    // before the caller read them. See FastLED#4375.
+    // No `channelData &&` short-circuit: that would skip canHandle() for a
+    // null channel, leaving the diagnostics stale in exactly the case this
+    // change exists to fix -- and making canHandle()'s own null branch
+    // unreachable from here. canHandle() checks `!data` first, so calling it
+    // unconditionally is safe and is what records the reason.
+    if (canHandle(channelData)) {
         mPendingChannels.push_back(fl::move(channelData));
     }
 }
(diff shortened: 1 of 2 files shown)


zackees:
## Corrected: the first version only fixed half the cases, and its "not fixed here" note was wrong

Force-pushed. Two changes, both found by testing the fix on hardware rather than trusting it.

**The clearing was in the wrong place.** I put it in `enqueue()`'s decline path. Hardware showed that only works when `enqueue()` is actually reached — `canHandle()` is *also* called during driver selection with no `enqueue()` following, and that is the path that left `attempted=True baud=2000000` visible on a channel the engine had just refused. The error message updated correctly while the numbers stayed stale, which is what exposed it:

```
tx=0 UART0 WS2811-400KHZ  attempted=True baud=2000000 err=''
tx=0 UART0 WS2812B-V5     attempted=True baud=2000000 err='RP UART: chipset timing needs a baud above...'
                          ^ error right, numbers still inherited
```

Moved into `canHandle()`, which is where a decline is decided regardless of who called it. Now:

```
tx=0 UART0 WS2811-400KHZ  attempted=True  baud=2000000 err=''
tx=0 UART0 WS2812B-V5     attempted=False baud=0       err='RP UART: chipset timing needs a baud above...'
```

**The original PR's "explicitly not fixed here" section was wrong.** I wrote that a per-call `timing` of `WS2812B-V5` reaches the engine as a 2 MHz baud, and flagged it as a separate defect about timing propagation. It is not a separate defect — 2000000 was the stale value this PR exists to remove. A freshly flashed board reports `attempted=False baud=0` on its first call, which is what identified the number as inherited rather than computed. That section is withdrawn; there is no known timing-propagation bug.

I had also reflashed with **master** rather than this branch while checking that, and briefly read the resulting empty error as the fix not working. Worth noting since the two branches now behave differently on exactly this field.

`--rpc-smoke` passes on the deployed branch build; lint clean.

zackees:
Applied — good catch, and it was a real hole.

`enqueue()` had `if (channelData && canHandle(channelData))`. The short-circuit means a **null** channel never reaches `canHandle()`, so the numeric diagnostics are not cleared and `mLastError` keeps the previous reason — the precise failure this PR exists to remove, surviving in the one case I did not think about. It also made `canHandle()`'s own `!data` branch unreachable from this call site, so the "no channel data" message I added could never fire.

Removed the short-circuit. `canHandle()` tests `!data` first, so calling it unconditionally is safe and is what records the reason:

```cpp
if (canHandle(channelData)) {
    mPendingChannels.push_back(fl::move(channelData));
}
```

Re-verified on hardware after redeploying — behaviour on the paths that were already working is unchanged:

```
tx=0 UART0 WS2811-400KHZ  attempted=True  baud=2000000 err=''
tx=0 UART0 WS2812B-V5     attempted=False baud=0       err='RP UART: chipset timing needs a baud above...'
```

`--rpc-smoke` passes, lint clean.

That is the third correction to this change: the clearing started in `enqueue()` (only covered half the cases), then moved to `canHandle()`, and now the call site that could bypass `canHandle()` entirely is closed. Each was found by running it rather than reading it.

zackees:
Applied, taking the prose over the suggestion block — they disagree.

The prose says to clear `mLastError` *before the successful return*, not at evaluation start, so a failed DMA start stays readable until `poll()` reports it. The committable suggestion does the opposite: it appends `mLastError.clear()` to the four numeric resets at the top. The prose is right, and that is what landed:

```cpp
    // Clear the reason here rather than with the numerics above: show()
    // records a failed DMA start in mLastError and poll() is what reports it,
    // so an evaluation must not be able to wipe that. Accepting is the only
    // outcome that makes a previous decline's reason wrong, and this is the
    // only path that reaches it.
    mLastError.clear();
    return true;
```

It is the mirror of the defect this PR is about: a decline was reporting the previous run's numbers, and an accept was leaving the previous decline's reason. Both are now scoped to the evaluation that produced them.

Worth being straight about the reach: this window is not observable through the RPC surface, because `show()` already clears `mLastError` on any path that reaches it, and nothing reads `lastError()` between `canHandle()` returning true and `show()` running. It closes the window for a caller that does, at no cost.

Re-verified on hardware after redeploying, alternating a declining and an accepting geometry on the same engine, 2 cycles:

```
WS2812B-V5     attempted=False started=False baud=0       encoded=0   err='RP UART: chipset timing needs a baud above the UART backend maximum'
WS2811-400KHZ  attempted=True  started=True  baud=2000000 encoded=480 err=''
WS2812B-V5     attempted=False started=False baud=0       encoded=0   err='RP UART: chipset timing needs a baud above the UART backend maximum'
WS2811-400KHZ  attempted=True  started=True  baud=2000000 encoded=480 err=''
```

Stable in both directions. `bash lint` clean, host suite 304/304 + 84/84.

(`bash autoresearch rp2350w --uart` still exits 1: its default is WS2812B-V5, whose 1225 ns needs a baud the PL011 cannot reach, so the engine declines and the harness has no capture to check. That is this PR's diagnostic working — the decline now says so instead of inheriting `attempted=True baud=2000000`. A separate encoder defect at 400 kHz turned up while checking it; filing that on its own.)

Issue #3526, opened 2026-07-01
I2S modern driver: wave8/wave3 general clockless (like PARLIO) + SPI unified — do NOT delegate to Yves


zackees:
## Design directive

The modern classic-ESP32 I2S parallel-out driver (currently scaffolded in `channel_engine_i2s_esp32dev` + `i2s_peripheral_esp32dev_esp`) must be built as a **general clockless driver** in the same shape as PARLIO — using **wave8 / wave3 encoding** driven by a fixed pixel clock — NOT by delegating to the classic Yves `i2s_esp32dev.cpp.hpp` bit-pattern driver.

Origin: user directive during FastLED#3516 Phase 2 drafting session. Was going to wrap the Yves C-style API from the modern peripheral; user redirected because the two are fundamentally different architectures and the modern driver should follow the parallel-IO/wave8 pattern PARLIO established.

## Why not wrap the Yves driver

The Yves classic driver at `src/platforms/esp/32/drivers/i2s/i2s_esp32dev.cpp.hpp` encodes each bit as a variable number of I2S pulses derived from `ChipsetTiming.T1/T2/T3` (see `i2s_define_bit_patterns` and `i2s_transpose_and_encode`). It's a per-chipset bit-pattern encoder that runs the transpose from an ISR streaming refill.

That model:
- Can't be selected per-channel at runtime — the bit patterns are set once at `i2s_init` time
- Doesn't fit the modern channel-manager flow where each `ChannelData` carries its own `ChipsetTimingConfig`
- Isn't the pattern the rest of the parallel-IO drivers use (PARLIO on ESP32-P4/C6/H2/C5 and LCD_CAM on ESP32-S3 both use wave8/wave3 expansion)

## The wave8 / wave3 pattern PARLIO uses

Reference: `src/platforms/esp/32/drivers/parlio/parlio_engine.h` lines 68-69 include `fl/channels/wave3.h` + `fl/channels/wave8.h`. Members at lines 295-309 hold:
```cpp
u32 mTimingT1Ns;
u32 mTimingT2Ns;
u32 mTimingT3Ns;
u32 mResetUs;

fl::Wave8BitExpansionLut  mWave8Lut;     // 8x expansion, always available
fl::Wave8ByteExpansionLut mWave8ByteLut;  // byte-indexed fast path (#2526)
fl::Wave3BitExpansionLut  mWave3Lut;      // 3x expansion, chipset-eligible only
bool                      mUseWave3;
[...]

PR #4413, opened 2026-09-13
fix(esp32): only alias ClocklessController to ClocklessI2S when its header was included

leifclaesson:
Defining `FASTLED_ESP32_I2S` on a classic ESP32 breaks FastLED's own compilation in 3.10.5. The failure is not in the sketch, it's in the library's own unity build sources.

**Reproduce**

Build anything for ESP32 with `-DFASTLED_ESP32_I2S`.

FastLED 3.10.5, ESP32 (classic, Xtensa), arduino-esp32 3.3.10 (ESP-IDF 5.5.4), xtensa-esp-elf-g++ 14.2.0.

**Error**

```
src/platforms/esp/clockless.h:35:31: error: 'ClocklessI2S' does not name a type; did you mean 'ClocklessSPI'?
   35 |   using ClocklessController = ClocklessI2S<DATA_PIN, TIMING, RGB_ORDER, XTRA0, FLIP, WAIT_TIME>;
      |                               ^~~~~~~~~~~~
      |                               ClocklessSPI
src/fl/chipsets/clockless_controller_impl.h:29:37: error: 'ClocklessController' does not name a type; did you mean 'ClocklessBlockController'?
```

followed by about two dozen `expected template-name before '<' token` through `chipsets.h`. It comes out of `src/fl/build/platforms+.cpp`. `src/fl/build/src.cpp` fails the same way behind it, make just stops at the first one.

**Cause**

Two conditions on the same feature disagree.

`src/platforms/esp/32/core/fastled_esp32.h:41` includes the I2S driver only when `FASTLED_INTERNAL` is not defined:

```c
#ifndef FASTLED_INTERNAL
#include "platforms/esp/32/drivers/i2s/clockless_i2s_esp32.h"
#endif
```

`src/platforms/esp/clockless.h:32` aliases the type with no such condition:

```c
#if defined(FASTLED_ESP32_I2S) && !ESP_IDF_VERSION_6_OR_HIGHER
  template <int DATA_PIN, typename TIMING, EOrder RGB_ORDER = RGB, int XTRA0 = 0, bool FLIP = false, int WAIT_TIME = 5>
  using ClocklessController = ClocklessI2S<DATA_PIN, TIMING, RGB_ORDER, XTRA0, FLIP, WAIT_TIME>;
```

The unity sources define `FASTLED_INTERNAL` and then include `FastLED.h` (`platforms/esp/compile_test.hpp:5`, `FastLED.cpp.hpp:1`), so in those the alias names a type whose header was deliberately skipped.

**Why not just drop the guard**

Removing the `#ifndef FASTLED_INTERNAL` around the include trades the error for a hard one. `clockless_i2s_esp32.h:104` refuses to be included under `FASTLED_INTERNAL`, and it should, because it carries file-static driver state (`gControllers`, `gNumControllers`) that belongs in the sketch translation unit only.

**What this changes**

The include site publishes the fact, and the alias tests that instead of restating the condition, so the two can't drift apart again.

```diff
--- a/src/platforms/esp/32/core/fastled_esp32.h
+++ b/src/platforms/esp/32/core/fastled_esp32.h
@@ -40,6 +40,7 @@
 #else
 #ifndef FASTLED_INTERNAL
 #include "platforms/esp/32/drivers/i2s/clockless_i2s_esp32.h"
+#define FASTLED_ESP32_I2S_DRIVER_AVAILABLE 1
 #endif
 #endif // ESP_IDF_VERSION_6_OR_HIGHER
 #endif // FASTLED_ESP32_I2S
```

```diff
--- a/src/platforms/esp/clockless.h
+++ b/src/platforms/esp/clockless.h
@@ -29,7 +29,7 @@ namespace fl {
 // Define platform-default ClocklessController alias for ESP32
 // Multiple driver types are available (ClocklessIdf4/ClocklessIdf5, ClocklessSPI, ClocklessI2S)
 // This alias selects the preferred default for backward compatibility
-#if defined(FASTLED_ESP32_I2S) && !ESP_IDF_VERSION_6_OR_HIGHER
+#if defined(FASTLED_ESP32_I2S_DRIVER_AVAILABLE)
   // I2S driver requested explicitly (not available on ESP-IDF 6.0+)
   template <int DATA_PIN, typename TIMING, EOrder RGB_ORDER = RGB, int XTRA0 = 0, bool FLIP = false, int WAIT_TIME = 5>
   using ClocklessController = ClocklessI2S<DATA_PIN, TIMING, RGB_ORDER, XTRA0, FLIP, WAIT_TIME>;
```

The ESP-IDF 6 case is unchanged, the include site already handles it and the new macro can't be defined there.

**Verification**

The internal units now fall through to the RMT branch, which costs nothing, because they only ever name these templates. The ESP32 `compile_test` is all preprocessor and nothing instantiates a clockless controller in there. I checked all 26 unity objects with `nm` and not one of them contains a clockless controller symbol.

The sketch translation unit still resolves `ClocklessController` to `ClocklessI2S`. I verified that with a `static_assert`, with `nm` on the object, and with a decoy `static_assert` naming the RMT type that correctly fails to compile.

On a 16-output WS281x controller firmware, all 224 translation units now compile. Before the change the build stopped at `platforms+.cpp`, which is number 179 of those 224, so the last fifth of the tree was never reached. A second ESP32 project that does not define `FASTLED_ESP32_I2S` was rebuilt as a regression check and is unaffected, still on RMT.

One note in case it saves someone a search: the `FASTLED_INTERNAL` guard predates the 3.10.4 tree reset in #3345, and that commit touched neither file, so this isn't fallout from it.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ESP32 I2S driver selection so clockless LED controllers consistently use the available driver.
  * Prevented mismatches between included ESP32 I2S support and controller configuration, improving build reliability across supported environments.


zackees:
Thanks for the clear diagnosis. You were right that the include guard and the alias condition had drifted apart.

On current `master` this is superseded. The code both halves of this PR patch has been removed: the `ClocklessI2S` (Yves parallel-out) driver was deleted in #3526 Phase 2e. `platforms/esp/clockless.h` no longer has the `FASTLED_ESP32_I2S` alias branch, and `fastled_esp32.h` no longer includes `clockless_i2s_esp32.h`. That's also why this branch now conflicts.

Checked on master (fbuild 2.5.23, arduino-esp32 / xtensa-esp-elf-gcc 14.2.0):

```
bash compile esp32dev --examples Blink --defines FASTLED_ESP32_I2S=1
FastLED.h:55: warning: "-DFASTLED_ESP32_I2S is removed (FastLED#3526 Phase 2e). Drop the flag; the modern I2S engine is now on Bus::FLEX_IO, 0. ..."
Flash: 323.59KB / 4.00MB
SUCCESS: Blink
```

So defining the flag no longer breaks the unity build. It emits a deprecation warning pointing at the replacement: drop the flag and select the I2S engine at runtime with `FastLED.setExclusiveDriver<fl::Bus::FLEX_IO, 0>()`.

Unless you're still seeing a failure on master (if so, please post the build output), this PR can be closed as superseded.

zackees:
Closing as superseded. The bug was real and the diagnosis was spot on: the `#ifndef FASTLED_INTERNAL` include guard and the `ClocklessController` alias condition had drifted apart. Both halves this PR patches have since been removed from `master`: the `ClocklessI2S` (Yves parallel-out) driver and `clockless_i2s_esp32.h` were deleted in #3526 Phase 2e, so there's nothing left for the alias to point at. On current `master`, `-DFASTLED_ESP32_I2S=1` builds a classic ESP32 cleanly with a deprecation warning pointing to `FastLED.setExclusiveDriver<fl::Bus::FLEX_IO, 0>()`. Thanks for the careful report. If you still see a build failure on `master`, please reopen with the build output.

Issue #4444, opened 2026-09-19
lint: enforce that every FastLED version string agrees, and that a release tag matches them


zackees:
Follow-up to #4442 / #4443.

## Problem

The version is written by hand in four places, and nothing checks that they agree with each other or with the release tag:

| Location | Form |
|---|---|
| `library.properties` | `version=X.Y.Z` |
| `library.json` | `"version": "X.Y.Z"` |
| `src/FastLED.h` | `#define FASTLED_VERSION XYYYZZZ` (1 + 3 + 3 digits) |
| `release_notes.md` | first heading, `FastLED X.Y.Z` (optionally ` (Next Release)`) |

They drifted without anything noticing. Before #4443, master had `library.properties` = 3.10.4, `library.json` = 3.10.3 and `FASTLED_VERSION` = 3010004 while 3.10.5 was already published. Tag `3.10.4` itself shipped with `library.json` still saying 3.10.3. An Arduino UNO Q user who installed master correctly saw "3.10.4" in the IDE, concluded the install had not worked, and the Library Manager offered to "update" them back to a broken release.

## Proposed check

**1. Consistency (runs in `bash lint`, every PR).** Parse the four locations and fail unless they all encode the same X.Y.Z, including the integer encoding of `FASTLED_VERSION`. Report every location and its value on failure. This is a cross-file structural check, so per `agents/docs/linter-architecture.md` it belongs on the Python side, not in the Rust per-file linter.

**2. Not behind the latest release (same check).** The shared version must be strictly greater than the highest `X.Y.Z` git tag that is an ancestor of `HEAD` — master is always the *next* release. Must degrade to a skip with a note, not a failure, when tags are unavailable (shallow CI clones, ZIP downloads); CI jobs that want this enforced need `fetch-depth: 0` or `fetch-tags`.

**3. Tag match (runs on tag push).** A workflow triggered on `X.Y.Z` tags fails if the tag name differs from the in-tree version, and if the top `release_notes.md` heading still says `(Next Release)`. No workflow currently triggers on version tags, so this is a new one.

## Acceptance

- Reverting any single hunk of #4443's version bump makes `bash lint` fail and name the offending file.
- A unit test covers the parser for each of the four formats plus the integer encoding (e.g. `3.10.6` ↔ `3010006`).
- Pushing a tag that disagrees with the tree fails the tag workflow.

## Open question

[...]

Issue #4442, opened 2026-09-19
Uno Q: master reports a version older than the 3.10.5 release, and CI never builds the real ArduinoCore-zephyr core


zackees:
Follow-up to #4408 and #4416 (Reddit report from an Arduino UNO Q user who installed master from the GitHub ZIP and still got compile errors).

## What happened

The reporter installed master correctly on Sep 13. It failed anyway, because #4410 only fixed the first layer of errors (`CoreDebug_Type` / `DWT_Type` / `COMP1`); the next layer — Zephyr's `min3()` macro from `<zephyr/sys/util.h>` clobbering the helper in `src/fl/gfx/rgbw.cpp.hpp` — was only fixed on Sep 17 in #4422. Both bugs reached users for the same reason, and the reporter was additionally confused by a version label. Two defects remain:

### 1. master identifies itself as 3.10.4, below the published 3.10.5

| File | master | tag `3.10.5` |
|---|---|---|
| `library.properties` | `3.10.4` | `3.10.5` |
| `library.json` | `3.10.3` | `3.10.5` |
| `src/FastLED.h` `FASTLED_VERSION` | `3010004` | — |
| `release_notes.md` heading | `FastLED 3.10.4 (Next Release)` | `FastLED 3.10.4` |

Cause: #3345 reset the tree to the 3.10.4 state, 3.10.5 was tagged from that, and #3347 restored the old master tree — including its pre-release version strings.

Effect: a correct master install shows up in the Arduino IDE as "3.10.4" (the reporter asked about exactly this), looks like an old release, and the Library Manager offers to "update" it to 3.10.5 — which is the broken release for this board.

### 2. The `arduino_uno_q` CI job does not build the core UNO Q users have

`ci/boards.py` builds `arduino_uno_q` with the STM32duino STM32U585 toolchain, with a comment saying no ArduinoCore-zephyr UNO Q platform is published. That is no longer true: `arduino:zephyr` 1.0.0 ("Arduino Q Boards", FQBN `arduino:zephyr:unoq`) is in the default Arduino package index, and it is what the IDE installs. Its `Arduino.h` pulls in Zephyr kernel headers and CMSIS 6, which is where both collisions came from; the STM32duino surrogate has neither, so CI stayed green through both bugs.

## Reproduction (Linux, arduino-cli 1.4.1, `arduino:zephyr` 1.0.0, `Arduino_RouterBridge` 0.4.3)

[...]

PR #4447, opened 2026-09-19
docs(agents): allow registry publishing as the one exception to the PlatformIO ban

zackees:
Maintainer decision: `pio pkg publish` is unbanned, so a release can reach the PlatformIO registry. That registry is at 3.10.3 while 3.10.5 is tagged and already in the Arduino index; there is no other way to publish to it.

`CLAUDE.md` now states the exception and its limits:
- only the publish subcommand, only in the release workflow/script
- no build, upload, test or monitor use — those stay on fbuild
- the ban lint and the forbidden-command hook are unchanged and still block it everywhere; the PR that adds the publish step allowlists that one file, with a reason

Rules text only. No lint, hook or workflow change here — nothing uses the exception yet. `ci/lint/banned_build_tools.py` passes on this branch (`CLAUDE.md` is already an allowed path).

Context: release-process plan following #4442 / #4444.

## Summary by CodeRabbit

* **Chores**
  * Updated release workflow permissions to allow PlatformIO package publishing.
  * Continued restricting build, test, device, upload, and monitor commands in release workflows.


zackees:
Closing: the maintainer decided not to publish with a registry token. FastLED will keep relying on the registry's crawler, so the publish subcommand was removed from #4448 and this exception is not needed. The ban stays absolute.

